PR #329, opened 2015-05-12
Add markdown docs for review.

gramster:
This is not meant to be merged, just commented on. I've just spent several hours yesterday and today trying to get the docstrings from the code into a Google Docs document in a form that is not a complete mess and haven't had much luck yet. Docs doesn't support markdown; about the best you can do AFAICT is to generate HTML, view the HTML in the browser, and copy and paste that into docs, but by the end of that process you have a pretty horrible result. 

It's actually easier/clearer to read the plain markdown I have here, and I'm hoping once I generate the pull request it may be possible to comment on the formatted version (although I suspect not). Either way, I suggest we use the PR to discuss the API for now.

There are three files: one for the context object, one for the storage APIs, and one for the BQ APIs.

### DIFF
--- a/sources/sdk/docs/bigquery.md
+++ b/sources/sdk/docs/bigquery.md
@@ -1,0 +1,2177 @@
+Module gcp.bigquery
+-------------------
+
+Google Cloud Platform library - BigQuery Functionality.
+
+Functions
+---------
+- **dataset** (name, context=None)
+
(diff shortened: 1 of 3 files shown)


gramster:
Closing this PR now as the feedback has been incorporated and this is pretty out of date.

PR #588, opened 2015-10-01
Making files strict compliant

### LINKED ISSUE #583: sources/web won't compile with TypeScript 1.6
Starting from typescript 1.6 Module bodies are parsed in strict mode: https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#module-bodies-are-parsed-in-strict-mode

When compiled it gives following errors:
datalab/jupyter.ts(88,29): error TS1121: Octal literals are not allowed in strict mode.
datalab/server.ts(27,8): error TS1214: Identifier expected. 'static' is a reserved word in strict mode. Modules are automatically in strict mode.
failed

vlasenkoalexey:
This should fix #583

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -86,7 +86,7 @@ function createJupyterServer(userId: string): JupyterServer {
 
   var userDir = userManager.getUserDir(userId);
   if (!fs.existsSync(userDir)) {
-    fs.mkdirSync(userDir, 0755);
+    fs.mkdirSync(userDir, parseInt('0755',8));
   }
 
   var server: JupyterServer = {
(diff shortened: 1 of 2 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -86,7 +86,7 @@ function createJupyterServer(userId: string): JupyterServer {
 
   var userDir = userManager.getUserDir(userId);
   if (!fs.existsSync(userDir)) {
-    fs.mkdirSync(userDir, 0755);
+    fs.mkdirSync(userDir, parseInt('0755',8));
```
supriyagarg: Just using the string also works, and is more compact:
fs.mkdirSync(userDir, '0755');

supriyagarg:
+1. This works, and lets us compile sources/web.

sources/web/datalab/jupyter.ts
gramster: I wonder whether you can rely on that working always though - as the node docs say the argument should be an int.

gramster:
This was fixed in e8659d9

Issue #606, opened 2015-10-04
UDF syntax woes


nikhilk:
Can't seem to make the intuitive work.

Cell 1:
````
%%bigquery udf --module word_filter
...
````

# First Attempt
Cell 2:
````
%%sql

DEFINE QUERY words
  SELECT word, word_count AS count, corpus
  FROM [publicdata:samples.shakespeare]
  WHERE length(word) >= 5

SELECT word, sum(count) as count
FROM $word_filter($words)
GROUP BY word 
ORDER BY count DESC
LIMIT 100
````

**Result -- error: Encountered " "FROM" "FROM "" at line 2, column 1.**

# Second Attempt
Cell 2 (added module name)
````
%%sql --module aaa

DEFINE QUERY words
  SELECT word, word_count AS count, corpus
  FROM [publicdata:samples.shakespeare]
  WHERE length(word) >= 5

SELECT word, sum(count) as count
FROM $word_filter($words)
GROUP BY word 
ORDER BY count DESC
LIMIT 100
````
Cell 3:
````
bq.Query(aaa, word_filter=word_filter).sql
````

Output:
````
u'SELECT word, sum(count) as count\nFROM &lt;gcp.bigquery._udf.UDF object at 0x7fcd4a3c6c50&gt;((SELECT word, word_count AS count, corpus\n  FROM [publicdata:samples.shakespeare]\n  WHERE length(word) >= 5))\nGROUP BY word \nORDER BY count DESC\nLIMIT 100'
````

**Note -- odd &lt;gcp.bigquery._udf.UDF object at 0x7fcd4a3c6c50&gt; in SQL

What am I missing? Whats the way to use these in the way other samples show using SQL.

Desired usage:

- None of the $ syntax. It should mimic what users see in the BigQuery docs.

- Perhaps all UDFs should be automatically included (or some heuristic where UDF names appearing in the SQL are automatically included rather than all).

- UDFs shouldn't be handled as parameter values.

I think we need to address this stuff before we put out the UDF feature in beta, otherwise we risk breaking changes.

In the context of breaking changes, we'll want to also change the unit testing of JavaScript, so it isn't run on the client, but rather in BigQuery.

Maybe we need to take a bigger look at UDFs before putting them out.

gramster:
I'd like to understand a bit the motivation behind the comment "we'll want to also change the unit testing of JavaScript, so it isn't run on the client, but rather in BigQuery.". 

[...]

PR #614, opened 2015-10-05
UDF Support: Direct calls and GCS hosted code

gramster:
Add the ability to call UDFs directly in the SQL without having to compose them via Python code into a FunctionCall first, as well as reference code in GCS.

Issue #606

### DIFF
--- a/sources/lib/api/gcp/bigquery/_query.py
+++ b/sources/lib/api/gcp/bigquery/_query.py
@@ -12,10 +12,12 @@
 
 """Implements Query BigQuery API."""
 
+import re
 import gcp._util
 import gcp.data
 import _api
 import _sampling
+import _udf
 import _utils
 
 
@@ -41,7 +43,7 @@ class Query(object):
     """
     return Query(_sampling.Sampling.sampling_query(sql, fields, count, sampling), context=context)
 
-  def __init__(self, sql, scripts=None, context=None, values=None, **kwargs):
+  def __init__(self, sql, context=None, values=None, udfs=None, **kwargs):
     """Initializes an instance of a Query object.
 
     Args:
@@ -52,7 +54,7 @@ class Query(object):
           It is possible to have variable references in a query string too provided the variables
           are passed as keyword arguments to this constructor.
 
-      scripts: array of UDFs referenced in the SQL.
+      udfs: array of UDFs referenced in the SQL.
       context: an optional Context object providing project_id and credentials. If a specific
           project id or credentials are unspecified, the default ones configured at the global
           level are used.
@@ -66,23 +68,70 @@ class Query(object):
     Raises:
       Exception if expansion of any variables failed.
       """
-    if kwargs or values or not isinstance(sql, basestring):
-      if values is None:
-        values = kwargs
-      sql, code = gcp.data.SqlModule.expand(sql, values)
-      if code:
-        if scripts is None:
-          scripts = code
-        else:
-          scripts.extend(code)
-
     if context is None:
       context = gcp.Context.default()
     self._context = context
     self._api = _api.Api(context)
     self._sql = sql
-    self._scripts = scripts
     self._results = None
+    self._code = None
+    self._imports = []
+    if values is None:
+      values = kwargs
+
+    self._sql = gcp.data.SqlModule.expand(sql, values, udfs)
+
+    # We need to take care not to include the same UDF code twice so we use sets.
+    udfs = set(udfs if udfs else [])
+    for value in values.values():
+      if isinstance(value, _udf.UDF):
+        udfs.add(value)
+    included_udfs = set([])
+
+    tokens = gcp.data.tokenize(self._sql)
+    udf_dict = {udf.name: udf for udf in udfs}
+
+    for i, token in enumerate(tokens):
+      prior = i - 1
+      while prior >= 0 and tokens[prior].isspace():
+        prior -= 1
+      if prior < 0 or not tokens[prior].upper() == 'FROM':
+        continue
+      next = i + 1
+      while next < len(tokens) and tokens[next].isspace():
+        next += 1
+      if next >= len(tokens) or not tokens[next] == '(':
+        continue
+
+      # We know now we have a 'FROM token (' sequence.
+
+      if token in udf_dict:
+        udf = udf_dict[token]
+        if token not in included_udfs:
+          included_udfs.add(token)
+          if self._code is None:
+            self._code = []
+          self._code.append(udf.code)
+          if udf.imports:
+            self._imports.extend(udf.imports)
+
+        fields = ', '.join([f[0] for f in udf._outputs])
+        tokens[i] = '(SELECT %s FROM %s' % (fields, token)
+
+        # Find the closing parenthesis and add the additional one now needed.
+        num_paren = 0
+        j = i + 1
+        while j < len(tokens):
+          if tokens[j] == '(':
+            num_paren += 1
+          elif tokens[j] == ')':
+            num_paren -= 1
+            if num_paren == 0:
+              tokens[j] = '))'
+              break
+          j += 1
+
+    self._sql = ''.join(tokens)
 
   def _repr_sql_(self):
     """Creates a SQL representation of this object.
@@ -116,7 +165,7 @@ class Query(object):
   @property
   def scripts(self):
     """ Get the code for any Javascript UDFs used in the query. """
-    return self._scripts
+    return self._code
 
   def results(self, use_cache=True):
     """Retrieves table of results for the query. May block if the query must be executed first.
@@ -264,7 +313,7 @@ class Query(object):
       An exception if the query was malformed.
     """
     try:
-      query_result = self._api.jobs_insert_query(self._sql, self._scripts, dry_run=True)
+      query_result = self._api.jobs_insert_query(self._sql, self._code, self._imports, dry_run=True)
     except Exception as e:
       raise e
     return query_result['statistics']['query']
@@ -298,7 +347,7 @@ class Query(object):
       table_name = _utils.parse_table_name(table_name, self._api.project_id)
 
     try:
-      query_result = self._api.jobs_insert_query(self._sql, self._scripts,
+      query_result = self._api.jobs_insert_query(self._sql, self._code, self._imports,
                                                  table_name=table_name,
                                                  append=append,
                                                  overwrite=overwrite,
--- a/sources/lib/api/gcp/bigquery/_udf.py
+++ b/sources/lib/api/gcp/bigquery/_udf.py
@@ -13,64 +13,47 @@
 """Google Cloud Platform library - BigQuery UDF Functionality."""
 
 import json
+import gcp.storage
 
 
-class FunctionCall(object):
-  """Represents a BigQuery UDF invocation.
+class UDF(object):
+  """Represents a BigQuery UDF declaration.
   """
 
-  def __init__(self, data, inputs, outputs, name, implementation):
-    """Initializes a UDF object from its pieces.
+  @property
+  def name(self):
+    return self._name
+
+  @property
+  def imports(self):
+    return self._imports
+
+  @property
+  def code(self):
+    return self._code
+
+  def __init__(self, inputs, outputs, name, implementation, support_code=None, imports=None):
+    """Initializes a Function object from its pieces.
 
     Args:
-      data: the query or table over which the UDF operates.
       inputs: a list of string field names representing the schema of input.
       outputs: a list of name/type tuples representing the schema of the output.
-      name: the name of the UDF function
+      name: the name of the javascript function
       implementation: a javascript function implementing the logic.
-    """
-    self._sql = FunctionCall._build_sql(name, inputs, data)
-    self._code = FunctionCall._build_js(inputs, outputs, name, implementation)
-
-  @property
-  def sql(self):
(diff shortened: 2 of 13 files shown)


gramster:
It might be good to get this in before you leave if you have time as I think it is orthogonal to other changes suggested (GCS hosted UDFs, not using JS for testing).

gramster:
So I've realized I overlooked something here because of the example we have been using with Shakespeare. When we have the intermediate Python code to create a UDF 'call' we wrap the UDF in a 'SELECT field1, .... FROM fn(...)' fragment. However, we are not doing any substitution if we call the UDF directly in a %%sql module; we just note that we need to include the code when we issue the request. That means we run foul of the inability to do SELECT * from a UDF. We actually need to find the start and end parentheses that wrap the UDF call so that we can wrap the whole thing in a SELECT field1,... etc, which brings us back to parsing SQL. Sigh.

gramster:
The direct-call UDFs are working properly now with 'SELECT *' . I'm going to add some more tests but things are looking good.

sources/lib/api/gcp/bigquery/_udf.py
```diff
@@ -13,34 +13,27 @@
 """Google Cloud Platform library - BigQuery UDF Functionality."""
 
 import json
+import gcp.storage
 
 
 class FunctionCall(object):
```
nikhilk: Do we still need this as well as FunctionEvaluation?
I am assuming we're no longer going to allow you to locally call a UDF directly.

sources/lib/api/gcp/bigquery/_udf.py
```diff
@@ -120,19 +84,77 @@ def __init__(self, inputs, outputs, name, implementation):
...
       """
-    self._inputs = inputs
     self._outputs = outputs
     self._name = name
     self._implementation = implementation
+    self._support_code = support_code
+    self._imports = imports
+    self._code = UDF._build_js(inputs, outputs, name, implementation, support_code)
+
+  def expand_imported_code(self):
```
nikhilk: Not sure we should keep pushing the browser running javascript as a way to test UDFs.

sources/lib/api/gcp/bigquery/_udf.py
```diff
@@ -120,19 +84,77 @@ def __init__(self, inputs, outputs, name, implementation):
...
+
+    # Construct a json representation of the output schema
+    # For example, [{'name':'field1','type':'string'},...]
+    output_fields = [{'name': f[0], 'type': f[1]} for f in outputs]
+    output_fields = json.dumps(output_fields, sort_keys=True)
+
+    # Build the JS from the individual bits with proper escaping of the implementation
+    if support_code is None:
+      support_code = ''
+    return '%s\n%s=%s;\nbigquery.defineFunction(\'%s\', %s, %s, %s);' % \
```
nikhilk: I think we should shift to named args... if not in this change, and sometime soon. These quickly become unreadable.

sources/lib/api/gcp/data/__init__.py
```diff
@@ -27,10 +27,109 @@ def sql(sql_template, **kwargs):
...
+    The formatted SQL statement with placeholders replaced with their values.
   Raises:
     Exception if a placeholder was found in the SQL statement, but did not have
     a corresponding argument value.
   """
   return _util.Sql.format(sql_template, kwargs)
+
+
+def tokenize(sql):
+  """ This is a crude tokenizer for our limited purposes.
```
nikhilk: Was this just for extracting out UDF uses? Maybe you mentioned, and I don't remember, but given a list of udf names, can we just look for "name(...)"?

So rather than finding all function calls, we have a fixed list of function calls to find.

sources/lib/api/gcp/bigquery/_udf.py
gramster: We want to keep FunctionEvaluation for local testing (I don't think we want to get rid of that; even the BQ web console uses local evaluation in the browser for UDF testing). But we don't really need FunctionCall if we expect users to only test locally or call the UDF within SQL.

sources/lib/api/gcp/bigquery/_udf.py
gramster: It's a low cost option. BQ console does this too as it doesn't cost anything. It may be easier to debug as well (getting local console errors rather than errors back from BQ). So I'd argue for keeping it.

sources/lib/api/gcp/bigquery/_udf.py
gramster: I'm not sure what you mean; this is JS code.

sources/lib/api/gcp/data/__init__.py
gramster: We need to find the corresponding closing parenthesis for any particular call, so we can wrap the whole thing in (SELECT f1, ... FROM ...). This tokenizer was created primarily for that, but once we were using it for that it made sense to use it to find the function calls as well; the resulting code is quite a bit simpler than an earlier version that used REs to find the calls and the tokenizer to find the call ends.

sources/lib/api/gcp/bigquery/_udf.py
nikhilk: So would we load everything from GCS as well? How do you know the dependencies of a single javascript function itself?

sources/lib/api/gcp/bigquery/_udf.py
nikhilk: The string formatting is happening in python. Using named placeholders instead of %s would make this readable/understandable.

sources/lib/api/gcp/data/__init__.py
nikhilk: Side note -- this shouldn't be public or perhaps even in __init__.py.

Secondly coroutines would be simpler to read through rather than hand-coded state machines, assuming they are an option. See http://eli.thegreenplace.net/2009/08/29/co-routines-as-an-alternative-to-state-machines/

I guess we're going to invariably end up in sql parsing. We should think about how we do this correctly across the scenarios -- parsing out UDF references, parsing begin/end of queries, comments in %%sql modules etc.

sources/lib/api/gcp/bigquery/_udf.py
nikhilk: I think we only need to support calling UDFs within SQL.

Calling into the javascript to test locally is interesting, but I don't know how realistic, given the javascript you are writing can depend on javascript in GCS.

In fact, if you are trying to write scalar UDFs, the best way to do it right now is with the UDF is GCS as a reusable script, and the specific UDF for the specific query schema as inline. For example, I want to write a UDF for geohashing over lat+lon which I can write and store in GCS to be used by a query-specific UDF that operates over columns a,b,c,lat,lon and roundtrips the a,b,c part.

I am assuming testing with the actual BQ dremel engine is cheap enough over sample data, so why would be not always do that instead of testing in the browser... just seems like a cheap hack ... yes, I wrote it, and it was actually written as a demo over lunch at an all day offsite when someone asked me if notebooks could enable test scenarios :)

gramster: I'll remove it, but once again though I feel I should point out that in-browser testing is what the BQ team themselves promote. So if it is a cheap hack its a cheap hack that the company promotes.

http://storage.googleapis.com/bigquery-udf-test-tool/testtool.html

gramster: Actually, if we're supporting calling UDFs in only SQL then we don't need UDF.\_\_call\_\_, nor do we need FunctionEvaluation or FunctionCall; they can all go.

sources/lib/api/gcp/data/__init__.py
gramster: Right now it is public as it is defined in gcp.data but used in gcp.bigquery. I can move it to gcp.bigquery although we may want it later.

gramster: I have an alternative implementation; do you prefer this:

def _next_token(sql):
      i = 0
    
      # We use some lambda's to make the logic more clear. The start_* functions return
      # true if i is the index of the start of that construct, while the end_* functions
      # return true if i point to the first character beyond that construct or the end of the
      # content.
      #
      # We don't currently need numbers so the tokenizer here just does sequences of
      # digits as a convenience to shrink the total number of tokens. If we needed numbers
      # later we would need a special handler for these much like strings.
    
      start_multi_line_comment = lambda s, i: s[i] == '/' and i < len(s) - 1 and s[i + 1] == '*'
      end_multi_line_comment = lambda s, i: s[i - 2] == '*' and s[i - 1] == '/'
      start_single_line_comment = lambda s, i: s[i] == '-' and i < len(s) - 1 and s[i + 1] == '-'
      end_single_line_comment = lambda s, i: s[i - 1] == '\n'
      start_whitespace = lambda s, i: s[i].isspace()
      end_whitespace = lambda s, i: not s[i].isspace()
      start_number = lambda s, i: s[i].isdigit()
      end_number = lambda s, i: not s[i].isdigit()
      start_identifier = lambda s, i: s[i].isalpha() or s[i] == '_' or s[i] == '$'
      end_identifier = lambda s, i: not(s[i].isalnum() or s[i] == '_')
      start_string = lambda s, i: s[i] =='"' or s[i] == "'"
      always_true = lambda s, i: True
    
      while i < len(sql):
        start = i
        if start_multi_line_comment(sql, i):
          i += 2
          end_checker = end_multi_line_comment
        elif start_single_line_comment(sql, i):
          i += 2
          end_checker = end_single_line_comment
        elif start_whitespace(sql, i):
          i += 1
          end_checker = end_whitespace
        elif start_identifier(sql, i):
          i += 1
          end_checker = end_identifier
        elif start_number(sql, i):
          i += 1
          end_checker = end_number
        elif start_string(sql, i):
          # Special handling here as we need to check for escaped closing quotes.
          quote = sql[i]
          end_checker = always_true
          i += 1
          while i < len(sql) and sql[i] != quote:
            i += 2 if sql[i] == '\\' else 1
          i += 1  # skip quote
        else:
          # We return single characters for everything else
          i += 1
          end_checker = always_true
    
        while i < len(sql) and not end_checker(sql, i):
          i += 1
    
        (yield sql[start:i])

sources/lib/api/gcp/bigquery/_udf.py
gramster: Done

sources/lib/api/gcp/bigquery/_udf.py
gramster: Moot point now as this is gone.

sources/lib/api/gcp/bigquery/_udf.py
gramster: Done

gramster:
I have removed all the \_\_call\_\_ related code, and changed the tokenizer to not be an FSM but a generator. Also in changing to the tokenizer it turned out we lost the ability to report on unresolvable UDF  names so that has been fixed; you'll see what looks like a big diff in Query constructor but its mostly an indentation change; git diff is pretty lame in that respect.

sources/lib/api/gcp/bigquery/_query.py
```diff
@@ -66,23 +68,74 @@ def __init__(self, sql, scripts=None, context=None, values=None, **kwargs):
...
+      while next < len(tokens) and tokens[next].isspace():
+        next += 1
+      if next >= len(tokens) or not tokens[next] == '(':
+        continue
+
+      # We know now we have a 'FROM token (' sequence.
+
+      if token not in udf_dict:
+        utoken = token.upper()
+        if utoken != 'TABLE_DATE_RANGE' and utoken != 'TABLE_DATE_RANGE_STRICT' and \
```
nikhilk: Will it work if we don't check for this? By work, I mean BigQuery will raise an appropriate error message. It would be better if we didn't hardcode some select function names here.

sources/lib/api/gcp/bigquery/_udf.py
```diff
@@ -89,50 +73,13 @@ def _build_js(inputs, outputs, name, implementation):
...
-      return FunctionEvaluation(self._implementation, data)
-    else:
-      return FunctionCall(data, self._inputs, self._outputs, self._name, self._implementation)
+    if support_code is None:
+      support_code = ''
+    return ('{code}\n{name}={implementation};\n' + \
+            'bigquery.defineFunction(\'{name}\', {inputs}, {outputs}, {name});')\
+                .format(code=support_code,
+                        name=name,
+                        implementation=implementation.replace('"', '\\"'),
```
nikhilk: Do we also need to escape the escapes ... so \ -> \\ before escaping quotes of course.

nikhilk: Or why do we need to do this either ... the implementation isn't going in into a string.

nikhilk:
LGTM.

sources/lib/api/gcp/bigquery/_query.py
gramster: Do you mean should we not raise errors ourselves if UDFs are not defined? If so, yes, BigQuery will report it, although it will uppercase the UDF name. E.g.:

    RequestException: Unknown TVF: FILTER

for a call to 'filter'.

sources/lib/api/gcp/bigquery/_udf.py
gramster: Correct, not needed.

PR #648, opened 2015-10-15
Fix a bug in the parsing of records with repeated scalar fields.

gramster:
We were expecting a list of values, but we get a list of dictionaries with each containing a single value keyed on 'v'.


nikhilk:
Do we have a test for this? Can we make sure we add that alongside or along the way?

nikhilk:
The change itself looks good of course.

PR #650, opened 2015-10-20
Don't make notebook copies under datalab path.

gramster:
If the user makes a copy of a notebook, and that notebook is in the
datalab folder, make the copy in the root.

### DIFF
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -267,6 +267,29 @@ function initializeNotebookApplication(ipy, notebook, events, dialog, utils) {
     notebook.prototype.save_notebook = function() {
       return originalSave.apply(this, /* check_lastmodified */ [ false ]);
     }
+
+    // A replacement notebook copy function that makes copies in the root if
+    // the source is under datalab/.
+    notebook.prototype.copy_notebook = function() {
+      var that = this;
+      var base_url = this.base_url;
+      var w = window.open('', IPython._target);
+      var parent = utils.url_path_split(this.notebook_path)[0];
+      if (parent.startsWith('datalab/')) {
+        parent = '';
+      }
+      this.contents.copy(this.notebook_path, parent).then(
+        function (data) {
+          w.location = utils.url_join_encode(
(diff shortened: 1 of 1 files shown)


sources/web/datalab/static/datalab.js
```diff
@@ -267,6 +267,29 @@ function initializeNotebookApplication(ipy, notebook, events, dialog, utils) {
...
+
+    // A replacement notebook copy function that makes copies in the root if
+    // the source is under datalab/.
+    notebook.prototype.copy_notebook = function() {
+      var that = this;
+      var base_url = this.base_url;
+      var w = window.open('', IPython._target);
+      var parent = utils.url_path_split(this.notebook_path)[0];
+      if (parent.startsWith('datalab/')) {
+        parent = '';
```
nikhilk: This would result in /datalab/samples/BigQuery/n.ipynb going to /samples/BigQuery/n.ipynb if I understand it right.

Would completely stripping off the path, so its just n.ipynb make more sense?

At some level, I feel this introduces or rather only worsens the impact of another shortcoming - no ability to move files within the UI. So now this creates files elsewhere where I might not want them to go.

gramster: Not quite. This does the old behavior if the notebook is not under /datalab (so the copy goes in the old directory). If the notebook is under /datalab then the copy is made in the top-level directory.  So /datalab/samples/BigQuery/n.ipynb would go to /n.ipynb

This was based off a request by Dinesh and is the location he suggested. It would be better in general if we had a 'copy to a different (specified) folder' functionality but in the absence of that we have to pick somewhere.

nikhilk: OK, I misread the change. I thought only /datalab was being modified, but the entire path is being removed. So this is fine... but yeah, the inability to specify path when creating copies, and inability to rename across folders leaves something to be desired (and addressed at some point).

For example, maybe this override is where we should prompt the user for a name and optional path.

nikhilk:
Change LGTM. We should make it possible to specify name + path in our override in a separate CL.

PR #654, opened 2015-10-22
Fix a bug in saving notebooks.

gramster:
The original save was replaced by a new version to work around an issue
with workspace sync. However, the new version does not return the wrapped
return value, so places where we chain off that future (e.g. with calls to
then()) got broken. This means, for example, that 'Convert to HTML' will
fail if the notebook has unsaved changes.


qimingj:
LGTM. Thanks!

PR #655, opened 2015-10-22
Jobs should run as my current project, unless specified

fhoffa:
Jobs should run as my current project, unless specified.

My datalab project is the one that should be used for running the queries, even if I have a destination table in a different project.

The destination table project needs to give me permissions to write into their dataset, but the query should be charged to the authenticated project.

Note that project_id is parsed a second time later in this method - there are 2 project ids, and they should be allowed to be independent.

### DIFF
--- a/sources/lib/api/gcp/bigquery/_api.py
+++ b/sources/lib/api/gcp/bigquery/_api.py
@@ -139,10 +139,7 @@ class Api(object):
     Raises:
       Exception if there is an error performing the operation.
     """
-    if table_name:
-      url = Api._ENDPOINT + (Api._JOBS_PATH % (table_name.project_id, ''))
-    else:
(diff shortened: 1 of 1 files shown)


gramster:
LGTM, thanks!

gramster:
Do you want to merge or should I?

nikhilk:
Gram, we should merge contribution pull requests.

Felipe, very much appreciate the contribution!

geekraver:
Sadly I can't do a merge from this PR. It seems like there is a way by adding Felipe's fork as an origin and doing a fetch but my git foo is not good enough to feel confident in doing that without something going wrong so I'll probably have to recreate this.

nikhilk:
I'll help w this on Monday.

gramster:
I figured out the problem.

PR #658, opened 2015-10-29
Support for querying from external tables.

gramster:
We leverage the new external tables for loading data from GCS too.
You can now do ExternalTable.load(BQTable). We may want to deprecate
the old Table.load API or make it take an ExternalTable argument to
have a symmetrical Table.load(ExternalTable).

Arguably the naming should be different - ExternalTable.load should
maybe be called ExternalTable.import?

### DIFF
--- a/sources/lib/api/gcp/bigquery/__init__.py
+++ b/sources/lib/api/gcp/bigquery/__init__.py
@@ -12,7 +12,9 @@
 
 """Google Cloud Platform library - BigQuery Functionality."""
 
+from _csv_options import CSVOptions
 from _dataset import DataSet, DataSets
+from _federated_table import FederatedTable
 from _job import Job
 from _query import Query
 from _query_job import QueryJob
--- a/sources/lib/api/gcp/bigquery/_federated_table.py
+++ b/sources/lib/api/gcp/bigquery/_federated_table.py
@@ -1,0 +1,98 @@
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+
+"""Implemenents External Table functionality."""
+
+import gcp
+import _api
+import _csv_options
+import _job
+
+
+class FederatedTable(object):
+
+  @staticmethod
+  def from_storage(source, source_format='csv', csv_options=None, ignore_unknown_values=False,
+                   max_bad_records=0, compressed=False, schema=None):
+
+    """ Create an external table for a GCS object.
+
+    Args:
+      source: the URL of the source objects(s). Can include a wildcard '*' at the end of the item
+         name. Can be a single source or a list.
+      source_format: the format of the data, 'csv' or 'json'; default 'csv'.
+      csv_options: For CSV files, the options such as quote character and delimiter.
+      ignore_unknown_values: If True, accept rows that contain values that do not match the schema;
+          the unknown values are ignored (default False).
+      max_bad_records The maximum number of bad records that are allowed (and ignored) before
+          returning an 'invalid' error in the Job result (default 0).
+      compressed: whether the data is GZ compressed or not (default False). Note that compressed
+          data can be used as a federated table but cannot be loaded into a BQ Table.
+      schema: the schema of the data. This is required for this table to be used as a federated
+          table or to be loaded using a Table object that itself has no schema (default None).
+
+  """
+    result = FederatedTable()
+    # Do some sanity checking and concert some params from friendly form to form used by BQ.
+    if source_format == 'csv':
+      result._bq_source_format = 'CSV'
+      if csv_options is None:
+        csv_options = _csv_options.CSVOptions()  # use defaults
+    elif source_format == 'json':
+      if csv_options:
+        raise Exception('CSV options are not support for JSON tables')
+      result._bq_source_format = 'NEWLINE_DELIMITED_JSON'
+    else:
+      raise Exception("Invalid source format %s" % source_format)
+
+    result._source = source if isinstance(source, list) else [source]
+    result._source_format = source_format
+    result._csv_options = csv_options
+    result._ignore_unknown_values = ignore_unknown_values
+    result._max_bad_records = max_bad_records
+    result._compressed = compressed
+    result._schema = schema
+    return result
+
+  def __init__(self):
+
+    """ Create an external table reference. Do not call this directly; use factory method(s). """
+    # Do some sanity checking and concert some params from friendly form to form used by BQ.
+    self._bq_source_format = None
+    self._source = None
+    self._source_format = None
+    self._csv_options = None
+    self._ignore_unknown_values = None
+    self._max_bad_records = None
+    self._compressed = None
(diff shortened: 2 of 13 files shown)


sources/lib/api/gcp/bigquery/__init__.py
```diff
@@ -12,7 +12,9 @@
 
 """Google Cloud Platform library - BigQuery Functionality."""
 
+from _csv_options import CSVOptions
```
nikhilk: I can't remember specifically, but is there some guideline in python about CSVOptions vs. CsvOptions?

sources/lib/api/gcp/bigquery/_external_table.py
```diff
@@ -0,0 +1,151 @@
...
+
+"""Implemenents External Table functionality."""
+
+import gcp
+import _api
+import _csv_options
+import _job
+
+
+class ExternalTable(object):
```
nikhilk: We might as well call this FederatedTable so people can connect the dots from the BQ docs. Not really a fan of that name.

I wish the concept that BQ would have introduced is that of a DataSource, so it would generalize to both DataSets and Tables... because I might have external table(s) that are related.

sources/lib/api/gcp/bigquery/_external_table.py
```diff
@@ -0,0 +1,151 @@
...
+
+  def __init__(self, source, source_format='csv', csv_options=None,
```
nikhilk: I think a static factory method might be better. FederatedTable.from_storage or something like that, so that when a different source is supported later, we can add another factory method.

sources/lib/api/gcp/bigquery/_external_table.py
```diff
@@ -0,0 +1,151 @@
...
+      'sourceUris': self._source,
+    }
+    if self._source_format == 'csv' and self._csv_options:
+      json['csvOptions'] = {}
+      json['csvOptions'].update(self._csv_options._to_query_json())
+    if self._schema:
+      json['schema'] = {'fields': self._schema._bq_schema}
+    return json
+
+  def load_async(self, table, mode, context=None):
```
nikhilk: This seemed a bit odd here... is the semantic we want "load an external table"?

I am assuming this is about creating an actual BQ table and loading data into it, in which case, this probably is better on Table.

sources/lib/api/gcp/bigquery/_query.py
```diff
@@ -26,7 +28,8 @@ class Query(object):
   """
 
   @staticmethod
-  def sampling_query(sql, context, fields=None, count=5, sampling=None):
+  def sampling_query(sql, context, fields=None, count=5, sampling=None, scripts=None,
+                     external_tables=None):
```
nikhilk: So the docs seem to use Federated Data Sources (https://cloud.google.com/bigquery/federated-data-sources) here -- so maybe this can be datasources rather than external_tables in the parameter naming.

sources/lib/api/gcp/bigquery/_schema.py
```diff
@@ -277,3 +277,7 @@ def __eq__(self, other):
       if not self._map[name] == other_map[name]:
         return False
     return True
+
+  def __ne__(self, other):
```
nikhilk: Is this a python thing that you implement the operator for inequality rather than equality?

sources/lib/api/gcp/bigquery/_table.py
```diff
@@ -416,7 +425,7 @@ def extract(self, destination, format='csv', csv_delimiter=',', csv_header=True,
     return job
 
   def load_async(self, source, mode='create', source_format='csv',
-                 csv_delimiter=',', csv_skip_header_rows=0, encoding='utf-8',
+                 csv_delimiter=',', csv_skip_leading_rows=0, encoding='utf-8',
```
nikhilk: This is a breaking change from the strict definition, but probably ok for now. Lets be careful about these...

sources/lib/api/gcp/bigquery/_table.py
```diff
@@ -450,40 +459,14 @@ def load_async(self, source, mode='create', source_format='csv',
...
-                                            quote=quote,
-                                            skip_leading_rows=csv_skip_header_rows)
-    except Exception as e:
-      raise e
-    return self._init_job_from_response(response)
+    csv_options = _csv_options.CSVOptions(csv_delimiter, csv_skip_leading_rows, encoding, quote,
+                                          allow_quoted_newlines, allow_jagged_rows)
+    dest = _external_table.ExternalTable(source, source_format, csv_options,
+                                         ignore_unknown_values, max_bad_records)
+    return dest.load_async(self, mode, self._context)
```
nikhilk: To correspond with the comment there, I think this functionality belongs here, since its about actual BigQuery tables.

nikhilk:
Added my comments on this one.

sources/lib/api/gcp/bigquery/__init__.py
gramster: The style guide doesn't say. In Python standard libraries acronyms seem to be capitalized (StringIO, UTF8Recoder, ...)

sources/lib/api/gcp/bigquery/_external_table.py
gramster: Done

sources/lib/api/gcp/bigquery/_external_table.py
gramster: Done

sources/lib/api/gcp/bigquery/_external_table.py
gramster: I think the right thing to do is to have Table.load take a FederatedTable parameter, but that breaks the existing signature of Table.load.

sources/lib/api/gcp/bigquery/_table.py
gramster: Done

sources/lib/api/gcp/bigquery/_schema.py
gramster: Both are implemented. We need to implement both because in each case this is a structural comparison, not referential comparison.

gramster:
PTAL

sources/lib/api/gcp/bigquery/_federated_table.py
```diff
@@ -0,0 +1,106 @@
...
+      compressed: whether the data is GZ compressed or not (default False). Note that compressed
+          data can be used as a federated table but cannot be loaded into a BQ Table.
+      schema: the schema of the data. This is required for this table to be used as a federated
+          table or to be loaded using a Table object that itself has no schema (default None).
+
+  """
+    return FederatedTable(source, source_format, csv_options, ignore_unknown_values,
+                          max_bad_records, compressed, schema)
+
+  def __init__(self, source, source_format, csv_options, ignore_unknown_values, max_bad_records,
```
nikhilk: The thought was to not have a ctor that was specific to storage, so that would work for non-storage scenarios later.

gramster: Well, there are no private constructors, but the idea is that people shouldn't call this, they should call the factory method - and calling this now would be painful as all the defaults have been removed.

nikhilk: We don't need to have a ctor that takes args... it can be a parameterless ctor, and the static factory method can set the various attributes.

My question is this - when there is another source, what will the second factory method do, when the ctor that exists right now doesn't work for it?

gramster: Other than 'source' (which is a string holding URL, so other formats are likely to support some variant of that), it seems that when a new source gets added most if not all of these would quite possibly still apply, but I'll move them out - at the risk of having to duplicate code then when a new source gets added. Until there is a new source you can't really call which is better IMO.

gramster: Done

PR #660, opened 2015-10-30
Notebook test framework.

gramster:
This comprises an autotesting script, Selenium driver and framework for handling
exceptions (apparent failures that are expected).

Some notebooks are changed, to fix text/plain sections for query results and where
tables are now output as lists. A few other notebooks that are out-of-date have
the config file set up to accept them as-is for now.

The test.py script can be invoked with a 'vcr' argument in which case HTTP calls
will be recorded to YAML files or played back from those files if they already
exist, which speeds up test execution by an order of magnitude. To use this
feature a special test version of the container is required; it is recommended that
this be used locally only (so there are build and run scripts for this but no stage
or publish scripts).

To use against a cloud deployment a --profile argumemt must be given so that the
browser opens with the user's authenticated profile.

Tests are specified in a YAML file. Entire notebooks can be ignore, or individual
cells. For even finer control individual MIME-type outputs can be used on a
particular cell.


gramster:
Closing this; I have created a new version with a few tweaks and no modification to current notebooks.

PR #661, opened 2015-10-31
Fix some bugs found through incidental inspection.

gramster:


nikhilk:
Changes LGTM.
Should we have any tests that account for these code paths?

gramster:
Looks like there's still a bug on one; missing the 'return'. Also, we may want to consider removing the gcp.data.sql function; it isn't necessarily that useful as it doesn't handle all the complexity of our BigQuery queries. I.e. I'm not sure we have a use case for it any more; it's there as a historical artifact.

PR #665, opened 2015-11-02
Fix a test.

gramster:
This test was broken (correctly) as we made a change recently to fix an
issue where if we were querying a table with a different project we
used that project to run the query. When the bug was fixed the test
was updated to reflect the correct behavior.


nikhilk:
LGTM

PR #666, opened 2015-11-03
Sphinx Read-the-Docs theme documentation config.

gramster:
To use, install the preprequisites:

    pip install sphinx sphinx_rtd_theme sphinxcontrib-napoleon

then in the docs directory, do 'make html' (or epub, or pdf, etc).

Output will be in the docs/_build directory.

More importantly, we should be able to enable the readthedocs webhook service on GitHub and have the docs rebuilt and hosted on RTD automatically, updated on each commit.


qimingj:
LGTM. Cool!

PR #669, opened 2015-11-07
Fix a minor inefficiency.

gramster:
Query.export calls Table.export. The latter blocks waiting for the Job
to complete and returns the Job. Query.export then calls wait() on the
Job too; this is benign but unneccessary. Simplifying the code.


nikhilk:
LGTM

PR #671, opened 2015-11-11
fix workspace issue that when it fails to sync, it does not reschedul…

qimingj:
Fix workspace sync issue that when it fails to sync, it does not reschedule the sync.


nikhilk:
LGTM

Issue #600, opened 2015-10-03
SQL module parsing needs to ignore comments


nikhilk:
Seems like the parsing doesn't account for comments (it finds variables that need replacing even within comments)

Note tried both "-- comments" and /\* */ comments. Don't know if comment handling itself accounts for these.

gramster:
The new tokenizer will handle SQL-style comments correctly. We still drop lines that start with '#'; that is handled as part of the magic in gcp.datalab (while SQL comments are handled as part of Query SQL expansion in gcp.bigquery). The rationale for the former is that we can have leading Python code, but it happens in the SQL code too. I'll fix that in a separate branch so '#' does not affect the SQL queries. Correcting the variable expansion in SQL comments will have to wait until the new UDF code is in so we can leverage the tokenizer instead of the current regexps.

qimingj:
LGTM. Thanks for the explanation.

PR #674, opened 2015-11-11
Don't strip Python-style comments out of SQL code in %%sql cells.

gramster:
Related to issue #600 600


qimingj:
LGTM. Will that also change the style of the text after "#"? If I do
%%sql
# select ...
It gives "select .." a comment style (green). Should we and can we change this?

gramster:
%%sql cells should currently be styled as SQL; we have no special handling
for the Python variable declarations. I don't think we are going to mess
with this anyway until we start reworking the UI.

PR #675, opened 2015-11-11
Include a static rendering of table data in the table viewer.

gramster:
This is useful for notebooks viewed in non-execution environments like
github. We will render the data in the div initially and the charts code
will overwrite that with a new rendering if that code executes.


nikhilk:
Code seems fine. It will probably help to see what the results are... specifically how does the static table rendering switch to dynamic?

gramster:
The dynamic table overwrites the innerHTML property and thus gets rid of the static table.

nikhilk:
OK, I guess I was expecting you always see a static HTML, even in the interactive notebook case, unless you actually care about seeing more. Something to revisit perhaps. That is motivated by the tables produced from the charting api being just ugly, so if we switch to something better, maybe interactive always is just fine.

PR #677, opened 2015-11-13
Charts now add an image/png display output to notebook.

gramster:


gramster:
Just want to capture this - there was some discussion yesterday about not hiding the PNG in interactive notebooks, and that instead we want to generate the text/html result as a display output, not execution output. However, we we are already attempting to do that (returning the charting code in IPython.core.display.HTML) and it still goes in as an execution output. If we put the png in as an execution output it will not show in nbviewer as the (empty) text/html takes precedence. So we are left with what I have done so far, namely hiding the PNG if the text/html can execute Javascript, or possibly we can try change the execution output type to a display output type in the JS chart code itself (I'll give that a shot).

gramster:
Hmm, well that just gives a notebook validation fails when you try to save the result, so I think I'm back to hiding the png.

gramster:
This needs more work; closing for now.

Issue #685, opened 2015-11-16
to_dataframe() bug when query returns no results


gramster:
See
https://stackoverflow.com/questions/33683992/to-dataframe-bug-when-query-returns-no-results

gramster:
Fixed in 4d92929

PR #686, opened 2015-11-16
Fix the handling of to_dataframe when the table is empty.

gramster:
Issue #685

### DIFF
--- a/sources/lib/api/gcp/bigquery/_table.py
+++ b/sources/lib/api/gcp/bigquery/_table.py
@@ -627,7 +627,7 @@ class Table(object):
 
     # Need to reorder the dataframe to preserve column ordering
     ordered_fields = [field.name for field in self.schema]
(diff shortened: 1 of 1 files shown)


sources/lib/api/gcp/bigquery/_table.py
```diff
@@ -627,7 +627,7 @@ def to_dataframe(self, start_row=0, max_rows=None):
 
     # Need to reorder the dataframe to preserve column ordering
     ordered_fields = [field.name for field in self.schema]
-    return df[ordered_fields]
+    return df[ordered_fields] if df else None
```
nikhilk: Wouldn't an empty DataFrame be a better choice... so the user doesn't have to check for None in notebook code?

gramster: Done.

PR #690, opened 2015-11-20
workspace update --- use new version of wsync

qimingj:
wsync 0.4.0 changes the following:

1. Support -username and -email flags to avoid the "Rand Om Hacker" name show up in Dev console.
2. Require a workspace name appended in "wsync sync" command.
3. "-repo" needs to be replaced with "-workdir".

All the above were confirmed with wsync owner.


nikhilk:
LGTM

PR #692, opened 2015-11-23
Use GCS as auth implementation. Also make version check script config…

qimingj:
…urable.

### DIFF
--- a/containers/datalab/content/run-local.sh
+++ b/containers/datalab/content/run-local.sh
@@ -18,6 +18,7 @@
 export DATALAB_ENV=local
 export DATALAB_INSTANCE_NAME=local
 export METADATA_HOST=localhost
+export VERSION_CHECK_SCRIPT_URL=https://storage.googleapis.com/cloud-datalab/deploy/version_check_stage.js
 
 # Setup environment variables.
 . /datalab/setup-env.sh
--- a/containers/datalab/content/run-local.sh
+++ b/containers/datalab/content/run-local.sh
@@ -18,7 +18,6 @@
 export DATALAB_ENV=local
 export DATALAB_INSTANCE_NAME=local
 export METADATA_HOST=localhost
-export VERSION_CHECK_SCRIPT_URL=https://storage.googleapis.com/cloud-datalab/deploy/version_check_stage.js
 
 # Setup environment variables.
 . /datalab/setup-env.sh
--- a/sources/web/datalab/authentication.ts
+++ b/sources/web/datalab/authentication.ts
@@ -57,37 +57,29 @@ function getAccessTokenFromMetadataServer(cb: common.Callback<string>) {
   }
 }
 
-function lookupUserFromDatastore(userId: string, accessToken: string, cb: common.Callback<boolean>) {
-  var postData: any = {
-    keys: [
-      {
-        path: [
-          {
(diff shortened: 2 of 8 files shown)


containers/datalab/content/run-local.sh
```diff
@@ -18,6 +18,7 @@
 export DATALAB_ENV=local
 export DATALAB_INSTANCE_NAME=local
 export METADATA_HOST=localhost
+export VERSION_CHECK_SCRIPT_URL=https://storage.googleapis.com/cloud-datalab/deploy/version_check_stage.js
```
nikhilk: Could this go in settings.json directly?

sources/web/datalab/authentication.ts
```diff
@@ -57,37 +57,29 @@ function getAccessTokenFromMetadataServer(cb: common.Callback<string>) {
...
   try {
-    httpapi.posts('www.googleapis.com', dataStorePath, null, postData, accessToken, null, function(e: Error, data: any) {
+    httpapi.gets('www.googleapis.com', storagePath, null, accessToken, null, function(e: Error, data: any) {
       if (e) {
-        logging.getLogger().error(e, 'Failed to query Datastore for authentication.');
-        cb && cb(e, null);
+        if (e.message.length >= 3 && e.message.substr(e.message.length - 3) == '404') {
+          // Most of times we cannot tell whether it is a transient error or the object not found,
+          // because they both return 404. But we assume it is caused by the object not exist.
+          cb && cb(null, false);
```
nikhilk: This can be simplified just a bit, as I think cb is always going to be non-null, right?

nikhilk:
Looks good -- just couple of minor suggestions.

sources/web/datalab/common.d.ts
```diff
@@ -43,6 +43,7 @@ declare module common {
     versionId: string;
     instanceId: string;
     analyticsId: string;
+    versionCheckScriptUrl: string;
```
nikhilk: Can you call this "configScript"?

containers/datalab/content/run-local.sh
nikhilk: Can you call this CONFIG_URL

sources/web/datalab/templates/tree.html
```diff
@@ -197,7 +197,7 @@
      });
      window.datalab = {};
   </script>
-  <script src="https://storage.googleapis.com/cloud-datalab/deploy/version_check.js"></script>
+  <script src="<%versionCheckScriptUrl%>"></script>
```
nikhilk: configUrl

sources/web/datalab/authentication.ts
qimingj: true. I want to be safe here since someday we may pass in a non-null cb.

containers/datalab/content/run-local.sh
qimingj: done

sources/web/datalab/common.d.ts
qimingj: done

sources/web/datalab/templates/tree.html
qimingj: done

PR #693, opened 2015-11-23
Fix storage sample notebook, and a few regressions that crept in from

gramster:
merge conflicts when merging federated table support.


nikhilk:
LGTM

PR #694, opened 2015-11-24
include tensorflow

qimingj:
Played with some sample TF code and it works well. :)


nikhilk:
LGTM. Thanks for including!

PR #696, opened 2015-11-24
Notebook test framework.

gramster:
This comprises an autotesting script, Selenium driver and framework for handling
exceptions (apparent failures that are expected).

Some notebooks are changed, to fix text/plain sections for query results and where
tables are now output as lists. A few other notebooks that are out-of-date have
the config file set up to accept them as-is for now.

The test.py script can be invoked with a 'vcr' argument in which case HTTP calls
will be recorded to YAML files or played back from those files if they already
exist, which speeds up test execution by an order of magnitude. To use this
feature a special test version of the container is required; it is recommended that
this be used locally only (so there are build and run scripts for this but no stage
or publish scripts).

To use against a cloud deployment a --profile argumemt must be given so that the
browser opens with the user's authenticated profile.

Tests are specified in a YAML file. Entire notebooks can be ignore, or individual
cells. For even finer control individual MIME-type outputs can be used on a
particular cell.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ test
 *.pyc
 *.idea
 *.iml
+*.ipynb.yaml
 
 # For building the server incrementally from an IDE like Eclipse it can be useful to create
 # the two symlinks below to point to sources/server/src/shared, but we don't want them
--- a/containers/datalab/build-test.sh
+++ b/containers/datalab/build-test.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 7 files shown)


gramster:
Couple of comments on this: the only change to the current container is the inclusion of test.js in the container, so that webdriver can inject it with a src attribute pointing to the local file system. It does not have to be in the container; we could put it elsewhere and change src. Ideally it would just be injected as an inline script by test.py but I couldn't get that to work easily (I tried reading from a file, escaping \\ and then escaping quotes but must have had something wrong). 

The build-test.sh and run-test.sh scripts are for building a different version of the container that supports recording or replaying network traffic so we can run tests much faster. build-test.sh just adds a couple of extra dependencies and run-test.sh only differs in the container label it used on start.

A lot of our checked-in notebooks are out of date, and originally I was rerunning them as part of this change to bring them up-to-date but I have decoupled that for now and just made the necessary adjustments to test.yaml to have the tests pass with the notebooks as-is. We could update them later separately.

ojarjur:
This was merged via #1115, but GitHub doesn't seem to know that.

PR #702, opened 2015-11-25
fix publish.sh tool

qimingj:


content/publish.sh
```diff
@@ -14,9 +14,13 @@
 # limitations under the License.
 
 # Publishes content files to Google Cloud Storage as publicly accessible content.
-
 # Synchronize content from the repository content directory to the
 # cloud within gs://cloud-datalab/content
+
+# First delete all content because rsync may not always override destination objects.
+gsutil -m rm -r -f gs://cloud-datalab/content
```
nikhilk: Is it better to copy over rather than first delete everything in the destination? I don't know... so sharing the question.

gramster: If we copied over first then the delete would be more complex and we would need to know to delete only the things that no longer exist. So deleting first seems simpler.

PR #703, opened 2015-11-27
Updatefix

qimingj:


content/publish.sh
```diff
@@ -14,9 +14,13 @@
 # limitations under the License.
 
 # Publishes content files to Google Cloud Storage as publicly accessible content.
-
 # Synchronize content from the repository content directory to the
 # cloud within gs://cloud-datalab/content
+
+# First delete all content because rsync may not always override destination objects.
+gsutil -m rm -r -f gs://cloud-datalab/content
```
nikhilk: Can we just change the rsync to a cp, rather than trying to empty the bucket first?

nikhilk:
Coment on the publish script, but LGTM on the actual fix.

content/publish.sh
qimingj: Did some experiments and it seems cp does not delete objects that we want to remove. Emptying the bucket seems cleaner?

nikhilk: We really don't want to ever delete (since we can't delete from the users' repos anyway)

qimingj: I'll merge the update fix and create a new one on the publish fix.

PR #705, opened 2015-11-30
Minor updates to genomics sample.

deflaux:
* Add a bit more narrative.
* Add .to_dataframe() where applicable so that results will display on github.
* Fix typo "genomics data" -> "genomic data" in the title but not the notebook name (so that links are not broken).


deflaux:
@nikhilk  There's one more change I'd like to add to this PR but I don't have the syntax right just yet.  Can you let me know the right way to parameterize an IN clause?  Thanks!

Some datasets use a 'chr' prefix in front of the chromosome name and others do not.  
* Python code update:
```
chromosome = ["1", "chr1"]
```
* SQL template update:
```
WHERE reference_name IN ($chromosome)
```
* resulting SQL --> notice the broken WHERE clause
```
u"SELECT
  reference_name,
  reference_bases,
  alternate_bases,
  INTEGER(FLOOR(start / 100000)) AS window,
  CONCAT(reference_bases, CONCAT(STRING('->'), alternate_bases)) AS mutation,
  COUNT(alternate_bases) WITHIN RECORD AS num_alts,
FROM
  [genomics-public-data:platinum_genomes.variants]
WHERE reference_name IN (['1', 'chr1'])
HAVING
  # Limit to bi-allelic SNP variants
  num_alts = 1
  AND reference_bases IN ('A','C','G','T')
  AND alternate_bases IN ('A','C','G','T')"
```

gramster:
There is a PR right now to support tuples/lists for IN clauses but in the current public beta this is not supported.

deflaux:
Okay then, I'll save that change for a later PR @gramster .

I think this is good to merge, but please let me know if you have any feedback that should be addressed before merging it.  Thanks!

gramster:
Thanks for doing this, Nicole!

It's mostly good but Nikhil and Dinesh should probably weigh in on the
addition of .to_dataframe(). I understand your reasoning, but I am not sure
we should make that change for three reasons:

- the other samples have the same issue and they aren't being changed
- we have an open issue to address this so that it shouldn't be necessary -
so this would be a temporary band-aid at best
- it may be confusing to users without them knowing why it is being done

gramster:
Closing this as we did the merge separately in a different PR.

PR #706, opened 2015-11-30
Samplefix

qimingj:


gramster:
You should probably remove the Storage commands notebook from this as it has been fixed separately. Also the Hello/hello change is not necessary I believe.

qimingj:
I'll abandon this pull request. The sample fix was contributed by Dinesh I believe. :)

gramster:
Okay. I think I'll wait for Dinesh to come back. I had a reason for using datalab-udf-samples instead of datalab-samples (which we use in a different notebook) - if we run both notebooks in parallel we can end up deleting the bucket in one notebook when it is still being used from the other notebook.

Issue #707, opened 2015-12-04
Socket closing during request?


geekraver:
See http://stackoverflow.com/questions/34093994/google-cloud-datalab-reload-set-from-bigquery-leads-to-respondnotready

Some clues: http://stackoverflow.com/questions/3231543/python-httplib-responsenotready

gramster:
Dug into this some. httplib2 has the ability to reuse a connection but AFAICT it should not do so if the Http object is not being reused (which it isn't). However, it may be prudent to add a "connection: close" header to get the server to close the connection itself after sending a response.

The other possibility might be if it takes a long time to get a response. So we should also add a handler for the ResponseNotReady exception so we can bubble it up as a regular Exception like we do with other errors that are not embedded in a response.

gramster:
Closing this in the hope that it is fixed by 7d3c38688435e5cd337d859670d48fecf27df41a (can't verify as we don't have a reliable repro). We can reopen if we see it in future.

PR #711, opened 2015-12-08
Attempt to avoid/handle ResponseNotReady exceptions better.

gramster:
We catch them and rethrow as regular exceptions like we do for some other
exceptions (better handling) and we have the server close the connection after
sending a response (attempt to avoid) - httplib2 has had issues in the past with
reusing sockets and having them close when keepalive timeouts happen, so by
having the server close the connection we may avoid reuse if it is happening;
at least it doesn't do any harm to add this header.

Issue #707.


qimingj:
Like you said it is probably an httplib2 bug, since in our code we don't reuse http objects, which is another source of problems.

According to https://github.com/jcgregorio/httplib2/pull/313, they fixed a very similar issue so perhaps we can upgrade httplib version to 0.9.2? We are using 0.9.1 I believe.

gramster:
That's a good idea. I never saw that but saw a couple of other cases in the past with this happening due to bugs in httplib2, with some people recommending against using it altogether. But we should update at least. I'll change this PR to be that change.

gramster:
Revised to just update httplib2 version.

PR #712, opened 2015-12-08
Fix the storage command sample notebook and fix a bug in list expansi…

gramster:
…on in

storage magics.

This notebook was not in the automated notebook tests as originally it was so
far out of date the automated test failed on (almost?) every cell as all the
new outputs were different. Unfortunately I didn't remember that it had been
skipped so when testing the notebook updates recently it was overlooked.

### DIFF
--- a/content/datalab/tutorials/Storage/Storage Commands.ipynb
+++ b/content/datalab/tutorials/Storage/Storage Commands.ipynb
@@ -116,7 +116,7 @@
     {
      "data": {
       "text/html": [
-       "<table><tr><th>Name</th><th>Type</th><th>Size</th><th>Updated</th></tr><tr><td>Hello.txt</td><td>text/plain</td><td>14</td><td>2015-10-04 16:51:31.855000+00:00</td></tr></table>"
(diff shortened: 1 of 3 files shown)


content/datalab/tutorials/Storage/Storage Commands.ipynb
```diff
@@ -128,7 +128,7 @@
     }
    ],
    "source": [
-    "%%storage list gs://cloud-datalab-samples"
+    "%%storage list --bucket gs://cloud-datalab-samples"
```
qimingj: -b also works. Should we encourage the shorter form? Same for -s, -d, -v, -o.

qimingj:
LGTM. If I remember correctly "UDFs using Code in Cloud Storage" may need to be updated too.

content/datalab/tutorials/Storage/Storage Commands.ipynb
gramster: No, in our samples we have stuck to longer form as being clearer.

gramster:
I checked the UDF notebook; it looks fine (it uses Python, not magics).

Issue #688, opened 2015-11-19
If notebook execution fails, the 'Running' text doesn't disappear from later cells


gramster:
This may be a dup. If you do a 'run all' and then execution doesn't complete as a cell fails then the remaining cells keep showing 'Running' and the spinner.

gramster:
Fixed in 77009b8109917d9b73369821ecae358715228523

PR #717, opened 2015-12-09
Remove the 'Running' text from uunexecuted cells upon an error.

gramster:
Issue #688

### DIFF
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -322,7 +322,15 @@ function initializeNotebookApplication(ipy, notebook, events, dialog, utils) {
         }
         else {
           this.element.addClass('completed');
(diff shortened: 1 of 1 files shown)


sources/web/datalab/static/datalab.js
```diff
@@ -324,6 +324,8 @@ function initializeNotebookApplication(ipy, notebook, events, dialog, utils) {
           this.element.addClass('completed');
           this.element.find('div.status').remove();
         }
+      } else {
+        this.element.find('div.status').remove();
```
qimingj: Safer to check this.element.find('div.status') for null?

gramster: I'll add the check although I suspect it isn't needed.

gramster:
Added null check.

Issue #681, opened 2015-11-16
MathJax not working in markdown cells


gramster:
Constructs like $pi$ should work.

gramster:
Fixed in 7630a8f127b6d3f5bec3baaf101bce7cb9007740

PR #718, opened 2015-12-10
Add (back) MathJax support.

gramster:
We have had two customers request this so far via Google-Cloud-Datalab-Feedback@.

I looked at the size of the scripts that get downloaded due to this. While the
MathJax source is about 17MB, the minified scripts are not bad at all, coming in
at about 400kB.

I went for the CDN-hosted scripts as these may already be cached and may be faster
to retrieve than through hosting in the container. That said, I went for the the
2.5 release; we could instead just use 'latest' in the URL to get the latest
release which could improve likelihood of cached versions.

As far as the impact supporting MathJax could have on a new UX is concerned, it
can be estimated by looking at the impact on Jupyter, which is mostly in:

  https://github.com/jupyter/notebook/blob/master/notebook/static/notebook/js/mathjaxutils.js

textcell.js and outputarea.js each have a few lines referencing this; most other
MathJax-related code in Jupyter (including some of mathjaxutils.js) is related just
to MathJax-related config (enabled or not, URL value, and handling load failure).

Issue #681

### DIFF
--- a/containers/datalab/config/ipython.py
+++ b/containers/datalab/config/ipython.py
@@ -29,3 +29,5 @@ c.InteractiveShellApp.exec_lines = []
 # Enable matplotlib renderings to show up inline in the notebook.
 c.InteractiveShellApp.matplotlib = 'inline'
 
+c.NotebookApp.mathjax_url = 'https://cdn.mathjax.org/mathjax/latest/MathJax.js';
+
--- a/sources/web/datalab/config/settings.cloud.json
+++ b/sources/web/datalab/config/settings.cloud.json
@@ -15,7 +15,6 @@
     "notebook",
     "-y",
     "--no-browser",
-    "--no-mathjax",
     "--log-level=DEBUG",
     "--debug",
     "--NotebookApp.allow_origin=\"*\"",
--- a/sources/web/datalab/config/settings.debug.json
+++ b/sources/web/datalab/config/settings.debug.json
@@ -15,7 +15,6 @@
     "notebook",
     "-y",
     "--no-browser",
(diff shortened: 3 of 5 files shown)


gramster:
Jupyter calls it. There is not meant to be any user intervention required to enable it if it is not explicitly disabled. You can just go ahead and enter math inside $$ in markdown cells.

nikhilk:
That seems a bit too much to always load IMO. 17MB just to render some equations... crazy! We should have some way for users who actually want to render math pay the price of loading it.

gramster:
400kb minified

nikhilk:
Also super large, and regardless, still that much script being loaded at startup.

gramster:
I'll experiment a bit and see if I can make a %enable_mathjax magic to load it on demand.

nikhilk:
We could probably generalize to some %%extension command that loads a
python library, runs a method on it and generate some output. There are
similar needs for other libraries - to augment requirejs, and optionally
run some script.

Example:

%%extension mathjax

I'll experiment a bit and see if I can make a %enable_mathjax magic to load
it on demand.

—
Reply to this email directly or view it on GitHub
<https://github.com/GoogleCloudPlatform/datalab/pull/718#issuecomment-166499945>
.

gramster:
I have a new version that I will open as a new PR. It adds a %extension magic that will load mathjax on demand. For now I am not making the underlying implementation generic (loading another Python script in turn) - that is an implementation detail that should not affect the user experience and until we actually have a need for additional extensions it may not be clear exactly how best to generalize this. We can refactor later if we add additional possible arguments to %extension.

Issue #604, opened 2015-10-04
%%bigquery execute should take inline sql


nikhilk:
Just like #560 which was about %%bigquery sample.

Similarly execute could take inline sql. This is low pri for now, but would become interesting when %%sql cannot assume bigquery, so marking as future.

gramster:
Fixed in bbf4ae10a6c6f249b723af77c558d9605239b8ed

PR #722, opened 2015-12-10
Add the ability to specify SQL inline in execute/sample/etc magics.

gramster:
Issue #604

### DIFF
--- a/sources/lib/datalab/gcp/datalab/_bigquery.py
+++ b/sources/lib/datalab/gcp/datalab/_bigquery.py
@@ -61,7 +61,7 @@ def _create_dry_run_subparser(parser):
   dry_run_parser = parser.subcommand('dryrun',
       'Execute a dry run of a BigQuery query and display approximate usage statistics')
   dry_run_parser.add_argument('-q', '--query',
-                             help='The name of the query to be dry run')
+                              help='The name of the query to be dry run')
   dry_run_parser.add_argument('-v', '--verbose',
                               help='Show the expanded SQL that is being executed',
                               action='store_true')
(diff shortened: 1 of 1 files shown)


sources/lib/datalab/gcp/datalab/_bigquery.py
```diff
@@ -184,32 +183,37 @@ def _create_load_subparser(parser):
...
-    config: the cell contents which can be variable value overrides.
+    cell: the cell contents which can be variable value overrides (if args has a 'query'
+        value) or inline SQL otherwise.
     env: a dictionary that is used for looking up variable values.
   Returns:
     A Query object.
   """
-  sql_arg = args['query']
+  sql_arg = args.get('query', None)
+  if sql_arg is None:
```
qimingj: if sql_arg is None, check and make sure cell is string. Otherwise, make sure cell is dict. This way we can print better error messages (instead of an exception thrown from env.update() or gcp.bigquery.Query() call)?

sources/lib/datalab/gcp/datalab/_utils.py
```diff
@@ -207,6 +207,8 @@ def parse_config(config, env):
   """ Parse a config from a magic cell body. This could be JSON or YAML. We turn it into
```
qimingj: nit: This could be JSON or YAML -> This could be JSON, YAML, or inline sql query.

sources/lib/datalab/gcp/datalab/_bigquery.py
```diff
@@ -184,32 +183,37 @@ def _create_load_subparser(parser):
   return load_parser
 
 
-def _get_query_argument(args, config, env):
+def _get_query_argument(args, cell, env):
```
qimingj: You might also want to update some arg names and comments. For example:

def _execute_cell(args, config):
  """Implements the BigQuery cell magic used to execute BQ queries.
   The supported syntax is:
   %%bigquery execute -q|--sql <query identifier> <other args>
   <config>
  Args:
    args: the arguments following '%bigquery execute'.
    config: optional contents of the cell interpreted as YAML or JSON.
  Returns:
    The QueryResultsTable
  """

Here "config" arg could be renamed to "cell_content". Same for dryrun.

qimingj:
LGTM with a couple of minor observations. Thanks!

gramster:
I noticed that call to _get_sql_arg always had a call to parse_config just before, so I moved the parse_config call into get_sql_arg which cleaned things up a bit and made the other changes you suggest unnecessary.  PTAL.

qimingj:
Looks great! LGTM.

PR #726, opened 2015-12-16
A simple alternative to widgets.

gramster:
One of the complaints we have had from users is the lack of support for Jupyter widgets.
This change addresses some of that limitation by allowing the variables in %%sql modules
that are displayed with %%chart to include simple dynamic controls which will re-execute
the query with the new values when changed (with client-side caching of past results).

This is controlled via the YAML chart options, in a 'variables' section. There are five
types of controls supported, 'text' (textbox), 'set' (checkbox group), 'picker' (dropdown
allowing a single value), 'slider' (combined slider/textbox for numeric value with
min/max/step parameters), and 'boolean' (a checkbox for a Boolean value).

 Each can have an optional 'label' property that defaults to the variable name, and a
'value' property to set the initial value, which, if not present, defaults to the default
value in the %%sql cell. The control type is set by a 'type' property which will default to
'text' for strings and 'set' for lists but must otherwise be specified.

For sliders, if s 'step' property is not specified and 'min' and 'max' are ints a 'step' of
1 will be used; else a step of (max - min) / 10.0 will be used.

For pickers, if the variable is not a list, a list of values can be provided with the
'choices' property.

Some examples:

    %%sql -m mother_ages

    state=['WA', 'TX', 'AZ']

    SELECT mother_age, COUNT(*) AS count FROM [publicdata:samples.natality]
    WHERE state IN $state
    GROUP BY mother_age
    ORDER BY mother_age

    %chart paged_table -d mother_ages
    variables:
      state:
        label: 'State'

The above will create a set of three checkboxes which can be checked/unchecked in different
combinations to affect the resulting table via altering the IN clause of the query.

    %%sql -m total_for_age_by_state

    age=20

    SELECT state, COUNT(*) AS count FROM [publicdata:samples.natality]
    WHERE mother_age==$age
    GROUP BY state
    ORDER BY state

    %chart table -d total_for_age_by_state
    variables:
      age:
        type: slider
        min: 10
        max: 60

The above will create a slider to allow an age in the range 10 through 60 to be selected.

    %%sql -m total_by_state
    state = 'WA'
    married = True

    SELECT mother_age, COUNT(*) AS count FROM [publicdata:samples.natality]
    WHERE state==$state AND mother_married==$married
    GROUP BY mother_age
    ORDER BY mother_age

    %chart columns -d natality
    title: Births by Age and Marital Status of Mother
    hAxis:
      title: Age
    variables:
      state:
        type: picker
        choices: [AZ, TX, WA]
      married:
        type: boolean

The above will create both a picker for the three choices of state and a checkbox to toggle
the value of martial status.

### DIFF
--- a/sources/lib/datalab/gcp/datalab/_bigquery.py
+++ b/sources/lib/datalab/gcp/datalab/_bigquery.py
@@ -231,7 +231,7 @@ def _sample_cell(args, cell_body):
     was specified. None otherwise.
   """
 
-  env = _notebook_environment()
+  env = _utils.notebook_environment()
   query = None
   table = None
   view = None
@@ -241,7 +241,7 @@ def _sample_cell(args, cell_body):
   elif args['table']:
     table = _get_table(args['table'])
   elif args['view']:
-    view = _get_notebook_item(args['view'])
+    view = _utils.get_notebook_item(args['view'])
     if not isinstance(view, gcp.bigquery.View):
       raise Exception('%s is not a view' % args['view'])
   else:
@@ -289,7 +289,7 @@ def _dryrun_cell(args, cell_body):
   Returns:
     The response wrapped in a DryRunStats object
   """
-  query = _get_query_argument(args, cell_body, _notebook_environment())
+  query = _get_query_argument(args, cell_body, _utils.notebook_environment())
 
   if args['verbose']:
     print query.sql
@@ -363,7 +363,7 @@ def _udf_cell(args, js):
 
   # Finally build the UDF object
   udf = gcp.bigquery.UDF(inputs, outputs, variable_name, js, support_code, imports)
-  _notebook_environment()[variable_name] = udf
+  _utils.notebook_environment()[variable_name] = udf
 
 
 def _execute_cell(args, cell_body):
@@ -379,7 +379,7 @@ def _execute_cell(args, cell_body):
   Returns:
     The QueryResultsTable
   """
-  query = _get_query_argument(args, cell_body, _notebook_environment())
+  query = _get_query_argument(args, cell_body, _utils.notebook_environment())
   if args['verbose']:
     print query.sql
   return query.execute(args['target'], table_mode=args['mode'], use_cache=not args['nocache'],
@@ -403,7 +403,7 @@ def _pipeline_cell(args, cell_body):
     raise Exception('Deploying a pipeline is not yet supported')
 
   env = {}
-  for key, value in _notebook_environment().iteritems():
+  for key, value in _utils.notebook_environment().iteritems():
     if isinstance(value, gcp.bigquery._udf.FunctionCall):
       env[key] = value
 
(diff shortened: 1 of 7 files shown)


nikhilk:
It seems this might be even better if it were generalized a bit... example:

    %%form
    config to create pickers etc. and associated python variables in the notebook

%%sql and %%chart would use those as any other variable.

gramster:
Regarding generalizing these, I have mixed feelings. It seems to me that there is a degree of coupling between a chart and the type of controls that is tighter than between the query and the control.

OTOH I see some benefit to making our variable definitions in %%sql cells richer, which could allow us to in most cases simplify what we need to specify in %chart. E.g. like we have source/datestring functions for some variables we could have functions for each of these controls in %%sql cell (not specifying the type, necessarily, but - say - specifying min/max/step values for variables that fall in a range). Then %chart would not need to specify these.

We could also infer these if the user doesn't specify them explicitly. E.g. a var that is just assigned an int could be inferred to be a textbox typed to int values - and we could automatically add controls for all non-overridden variables in %chart. So going back to the first example above we wouldn't need to specify anything extra in %chart to get the control. Similarly for the second example if we had declared in the %%sql something like:

  age = range_var(10, 60, default=20)

In the third example, without specifying anything we could get a text box for 'state' and a checkbox for 'married'.

That seems to allow us factor out much of the behavior that may be coupled to the %sql (the choice of values, ranges, etc), while leaving the choice of control type (if not a sensible default) to %chart.

Thoughts?

nikhilk:
I am not suggesting to put this into a SQL cell. Just making sure. I am
suggesting creating a new %%form cell, and having widgets to be its output
rather than in the chart. With better notebook ux these would be elevated
to a notebook level sidebar or something else rather than lost in the
vertical scroll.

I think there is some overlap with args in a SQL cell. I think it would be
a good time to think how args, form etc generalize to notebook level
arguments... esp. as we build out deployment from a notebook.

gramster:
Bradley, regarding YAML help, I think we would want to do this via samples. The YAML can include all the options from Google Charts so trying to add inline help for YAML is problematic.

gramster:
Nikhil, one of my concerns with moving them out and then treating them as *implicit* variables is that we are breaking the current model where all variables are declared locally (unless they refer to other 'modules' like queries and UDFs). We're starting to have a mixed model - which begs the question of whether we should just have a %%parameters section with a notebook-scoped set of variables (a discussion we had before and decided against).

The other issue with making the controls separate - either in a cell (which raises all sorts of usability issues if the control cell and the affected cell are not both in scroll view, as you note) or later elevating them to a sidebar -  is that notebook execution gets more complex when a control is changed. Changing a control can affect multiple cells and we need to find and update all of these, but some may not have been executed and/or execution may be in progress. Perhaps that isn't too bad but it seems it could get complicated. Currently with the controls being local to a cell they only appear when the cell gets executed and the event handlers get bound at the same time; if you clear the cell or start re-executing it they are gone. This keeps things pretty simple.

My suggestion is to keep this as is, as it has value, is simple, and does not break existing things and to deal with the broader issue of separate controls (and dashboards) separately, in conjunction with an updated notebook UX (and perhaps then the answer will be a set of global variables that can populate a sidebar and/or a dashboard).

qimingj:
Samples are fine if they cover all the options. Maybe create a notebook to make sure all options are described and some are in demo? Maybe we can create such notebook and check it in, and then provide a github link in the helper message of %chart?

aaront:
What's keeping back the support for Jupyter widgets? Is it an issue in their code?

gramster:
Jupyter widgets rely on a communication channel that doesn't exist in
Datalab.

selwyth:
2 years later, is it supported? I see #1689 suggest that it is supported now, but I had the same issue as here when trying to do it, so am curious if it is and I'm doing it wrong.

qimingj:
@yebrahim? Was a new release made after the fix?

yebrahim:
Yes, the widgets should be working in the latest version of Datalab.
@selwyth can you open a new issue and post the error you're getting, as well as the version of Datalab you're using?

selwyth:
Posted to #1980

Issue #713, opened 2015-12-08
Better handling of insufficient browser width


gramster:
If the browser window is narrow a lot of our menubar items disappear. We should address this somehow (quick n dirty might be to detect that and throw up a message like we do with the update message).

gramster:
Fixed in 6edaf9fca5bd7dfb8ce5f388b721423f1de6cf29

gramster:
The 'fix' wasn't so reopening.

ojarjur:
There are two ways to address this issue:

1. Make the layout of the page gracefully scale down to match small window sizes.
2. Set a minimum size for the layout of the page and make the user's window scroll over that page if it is too small to fit the minimum size.

The first option is preferable, but it requires much more sophisticated UI logic than we currently have, so I've instead implemented the second option.

With that, this should now be fixed.

PR #727, opened 2015-12-16
Handle limited width browser experience better.

gramster:
Change the CSS so that if the user has a browser of insufficient width we will
expand the toolbar to multiple lines as necessary instead of the toolbar items
disappearing.

Issue #713

### DIFF
--- a/sources/web/datalab/static/datalab.css
+++ b/sources/web/datalab/static/datalab.css
@@ -115,13 +115,14 @@ body {
 #mainToolbar, #sidebarToolbar {
   line-height: 48px;
   vertical-align: middle;
-  height: 48px;
+  min-height: 48px;
   color: #444;
   background-color: #f3f3f3;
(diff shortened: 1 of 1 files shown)


nikhilk:
I had tried this or something very similar, and found it had other worse side-effects. Assuming you've played around with the UX enough to make sure it doesn't come with those, then this is LGTM.

gramster:
It seemed fine to me in my tests (Chrome and Safari). Not necessarily pretty when the toolbar wraps but it works. If you make the browser really narrow then the feedback/info icons at the top right start misbehaving but I wasn't trying to address that (and we're talking really narrow here with the toolbar wrapped into 5 or more lines).

gramster:
Perhaps you took 'position:absolute' out of sidebarContent too. I did that along the way to this change and that did make a terrible mess of things. The joys of CSS....

PR #728, opened 2015-12-17
Revert the CSS toolbar overflow changes.

gramster:
I didn't notice when testing this that this results in a toolbar that can be scrolled
out of view if the notebook is longer than will fit in the browser window. That is
pretty undesirable behavior, so reverting for now and will investigate a different
fix.


nikhilk:
That sounds familiar... the reason why I didn't just take care of these sort of things.

nikhilk:
LGTM.

PR #734, opened 2015-12-29
Lazily load mathjax on demand with %extension magic

gramster:


sources/lib/datalab/gcp/datalab/_extension.py
```diff
@@ -0,0 +1,53 @@
...
+  return _utils.handle_magic_line(line, None, parser)
+
+
+def _extension(args, cell):
+  ext = args['ext']
+  if ext == 'mathjax':
+    url = "https://cdn.mathjax.org/mathjax/latest/MathJax.js"
+    config = 'TeX-AMS-MML_HTMLorMML'
+    return IPython.core.display.HTML("""
+      <script>
```
qimingj: If a user runs %extension mathjax in multiple cells, will the <script> section run multiple times?

gramster: Added a guard for that.

qimingj:
LGTM!

Issue #733, opened 2015-12-26
Datalab chart map is not displayed correctly


VelizarVESSELINOV:
Please find an example from the default installed notebook: ```notebooks/datalab/tutorials/Data/Interactive Charts with Google Charting APIs.ipynb```

The map is single color and the button's icons are not readable:
<img width="1023" alt="datalab map" src="https://cloud.githubusercontent.com/assets/8209940/12005769/592cb8be-ab6b-11e5-987b-c71ad8b3f4a0.png">
The screenshot is on my Chrome browser, on the Safari the single color is replaced by the map but the button icons are still unreadable.

gramster:
The buttons being unreadable is something we are aware of. The single color map means the map didn't load - the push pins can still be drawn independently. These are both Google Charts issues, not specific to Datalab.

gramster:
fixed in 16fb344839582fa15fb2ea891b5b3026e87c01f4

VelizarVESSELINOV:
On my DataLab project is still not working map, what it is needed to deploy your fix on existing DataLab project?

VelizarVESSELINOV:
Works now :)

Cecitux:
@VelizarVESSELINOV  
@gramster 
Could you explain me how to result this issue, I have the following error for see a map.

![image](https://cloud.githubusercontent.com/assets/16903090/16935127/5672b228-4d22-11e6-948c-a019c737f011.png)


Thanks

scottma:
I have the same error here, wonder if anyone has found a solution to this.

VelizarVESSELINOV:
At the end, I stopped using `%%chart map` because of stability and limitation issues, JavaScript Google map call is more reliable, scalable, portable.

qimingj:
Can anyone share a bigquery table with me(qimingj@google.com) and the query to reproduce the issue? I cannot reproduce the error using our sample notebooks.

magyarb:
I can not use the %%chart map, because i get the API error: MissingKeyMapError. Where can i specify my API key?

chmeyers:
MissingKeyMapError is this issue: https://github.com/googledatalab/notebooks/issues/74

yebrahim:
Updated that issue with one possible work around to embed your key in the notebook.

PR #736, opened 2015-12-30
Fix the conflict between Bootstrap and map charts.

gramster:
Issue #733


qimingj:
Although I don't quite understand how this addresses the issue, but I am fine with it if it works fine. LGTM!

gramster:
More details here: https://github.com/twbs/bootstrap/issues/1552

Issue #748, opened 2016-01-17
Error when loading a JSON file into a BigQuery table 


hubgit:
Trying to load data from a newline-delimited JSON file into a BigQuery table fails with an error:

## Input

```python
import gcp.bigquery as bq

bq.DataSet('example').create()

table = bq.Table('example.foo').create(schema=schema, overwrite=True)

table.load('gs://{bucket}/docs.ndjson', mode='append', source_format='json')
```

## Output

```
RequestException: Only CSV imports may specify a field delimiter.
```

gramster:
Fixed in #750

PR #750, opened 2016-01-20
CSV load fix

gramster:
BQ is now rejecting Table loads of non-CSV files if CSV options are
specified. We would supply these regardless of the source format, so
this change only adds these arguments to the configuration if the
sourece format is CSV.

Issue #748


gramster:
Unfortunately this is a historical artifact with BigQuery that is hard for them to fix now. The CSV options we have in Query are also applicable only to CSV in load, but they were used at the top-level originally in load and now can't be changed. 

So yes, it applies only to CSV.

gramster:
So this was a good catch, actually. Encoding (but not the other fields) should not be specific to CSV. This is an oversight in BQ federated table access, where they will need to add encoding to the options for querying JSON data. But I will update this change to include factoring out encoding from the CSV options.

gramster:
PTAL

qimingj:
LGTM

PR #751, opened 2016-01-20
Changes to get controls and static charts to play together nicely.

gramster:
Specifically:

- controls are shown but disabled when converting to HTML
- PNG static charts are dynamically removed in converted HTML when the chart is redrawn

The code to do the latter previously relied on the IPython JS APIs being present; it
now falls back to a pure DOM approach if necessary.

Fix the static chart removal so it works when not running under IPython.


qimingj:
LGTM!

VelizarVESSELINOV:
Do you have example of controls?

PR #758, opened 2016-02-02
Fix a couple of bugs found from rerunning samples.

gramster:
I'm separating this from the pull request which has the updated samples
in the hope of having it reviewed sooner.


qimingj:
LGTM.

Issue #763, opened 2016-02-05
Project is null in local datalab setup


parthea:
Hello,

I'm running datalab in a local development environment . I'm unable to run any queries using gcp.bigquery. I ran `gcloud init` and it appears that the project id is correct in the output of `gcloud init` , however I receive the following exception in the local datalabs web interface.

```
/usr/local/lib/python2.7/dist-packages/gcp/bigquery/_query.pyc in to_dataframe(self, start_row, max_rows, use_cache)
    280       A Pandas dataframe containing the table data.
    281     """
--> 282     return self.results(use_cache=use_cache) \
    283         .to_dataframe(start_row=start_row, max_rows=max_rows)
    284 

/usr/local/lib/python2.7/dist-packages/gcp/bigquery/_query.pyc in results(self, use_cache)
    213     """
    214     if not use_cache or (self._results is None):
--> 215       self.execute(use_cache=use_cache)
    216     return self._results.results
    217 

/usr/local/lib/python2.7/dist-packages/gcp/bigquery/_query.pyc in execute(self, table_name, table_mode, use_cache, priority, allow_large_results)
    426     """
    427     job = self.execute_async(table_name=table_name, table_mode=table_mode, use_cache=use_cache,
--> 428                              priority=priority, allow_large_results=allow_large_results)
    429     self._results = job.wait()
    430     return self._results

/usr/local/lib/python2.7/dist-packages/gcp/bigquery/_query.pyc in execute_async(self, table_name, table_mode, use_cache, priority, allow_large_results)
    390                                                  table_definitions=self._external_tables)
    391     except Exception as e:
--> 392       raise e
    393     if 'jobReference' not in query_result:
    394       raise Exception('Unexpected response from server')

RequestException: Not found: Project null
```

gramster:
How are you running it, and on what platform? The container/datalab/run.sh script which runs locally maps your ~/.config directory to a directory in the container so Datalab can get hold of your access token, project ID etc. Its possible that isn't right for your setup.

parthea:
I am running the container/datalab/run.sh script on linux . 

I added the 'shell'  argument to enter into the container. I can see my project number after entering the following command.
`echo $DATALAB_PROJECT_NUM`

I followed these instructions to [initialize](https://github.com/GoogleCloudPlatform/datalab/wiki/Development-Environment) and [build/run](https://github.com/GoogleCloudPlatform/datalab/wiki/Build-and-Run) datalabs. 

[...]

PR #767, opened 2016-02-06
Use latest version of Google Cloud SDK instead of fixed version (0.9.57)

parthea:
PR to resolve issue #763. If accepted, I will also update the development wiki to `gcloud init` instead of `gcloud config set project 'myProject'`


parthea:
I receive the following error during the datalab build process if I use the latest version of Google Cloud SDK

```
Traceback (most recent call last):
  File "/tools/google-cloud-sdk/bin/bootstrapping/gsutil.py", line 12, in <module>
    import bootstrapping
  File "/tools/google-cloud-sdk/bin/bootstrapping/bootstrapping.py", line 9, in <module>
    import setup
  File "/tools/google-cloud-sdk/bin/bootstrapping/setup.py", line 41, in <module>
    reload(google)
ImportError: No module named google
```

Closing this PR, as the changes only appear to solve the issue in #763 after datalab is built at least once using the fixed sdk version (0.9.57).

PR #788, opened 2016-03-18
Fix an issue that when "datalab" branch is deleted but "datalab_main" is

qimingj:
not, a deployment does not recreate "datalab" branch.


gramster:
LGTM

PR #789, opened 2016-04-01
Look for error.message in error responses.

rimey:
The code was looking for error.errors.0.message, but some APIs
(e.g., the Google Monitoring API) don't return error details under
error.errors, while error.message is generally supposed to be present.


supriyagarg:
The changes in this pull request have been made part of a new datalab branch: 

https://github.com/GoogleCloudPlatform/datalab/tree/feature/stackdriver_monitoring

PR #801, opened 2016-04-20
Added back support for mapping in gcloud config. This is not used by …

gramster:
…Datalab

but will allow users of Datalab to use gcloud in bash cells.

Cleaned up some dead code around setting project ID.

Some other small cleanups.


qimingj:
LGTM

PR #812, opened 2016-05-02
Minor cleanups and fixes for local run

gramster:


containers/datalab/content/setup-env.sh
```diff
@@ -15,12 +15,14 @@
 
 # Sets up various environment variables within the docker container.
 
-export DATALAB_USER=`gcloud -q config list --format yaml | grep account | awk -F" " '{print $2}'`
-export DATALAB_PROJECT_ID=`gcloud -q config list --format yaml | grep project | awk -F" " '{print $2}'`
-if [ -z $DATALAB_PROJECT_NUM ]; then
-  export DATALAB_PROJECT_NUM=`curl --silent -H "Metadata-Flavor=Google" http://metadata.google.internal/computeMetadata/v1beta1/project/numeric-project-id`
+if [ -z "$GOOGLE_APPLICATION_CREDENTIALS" ] && [ -z "$GCLOUDSDK_CONFIG" ]
+then
+  export CLOUDSDK_CONFIG=/content/datalab
```
qimingj: I wonder if we can hard code the path in nodejs to be "/content/datalab" if DATALAB_ENV="local", else "/root/.config/gcloud" for GCE run? It seems the location doesn't need to be customized by users?

containers/datalab/content/setup-env.sh
```diff
@@ -15,12 +15,14 @@
...
 fi
-if [ -z $DATALAB_INSTANCE_NAME ]; then
-  export DATALAB_INSTANCE_NAME=$GAE_MODULE_VERSION
+if [ -z "$DATALAB_ENV" ]
+then
+  export DATALAB_ENV="local"
 fi
 
+export DATALAB_USER=`gcloud -q config list --format yaml | grep account | awk -F" " '{print $2}'`
```
qimingj: Is this still needed? The only usage of the variable seems to be as home dir name for jupyter. Since local run is no longer for multiple users, it might be Okay to default it to "anonymous" as in https://github.com/GoogleCloudPlatform/datalab/blob/4a86c50c0a1f366ad2239e3d156780659824abaf/sources/web/datalab/userManager.ts#L51.

sources/web/datalab/auth.ts
```diff
@@ -136,7 +138,10 @@ export function handleAuthFlow(request: http.ServerRequest, response: http.Serve
           logging.getLogger().info('Got tokens');
           oauth2Client.setCredentials(tokens);
           // Push them to Jupyter and handle request.
-          persistCredentials(tokens);
+          var email = persistCredentials(tokens);
+          if (email) {
+            settings.instanceUser = email;
```
qimingj: I think it is probably fine to not set instanceUser so it is consistently "anonymous" in all cases.

qimingj:
A few observations.

containers/datalab/content/setup-env.sh
gramster: Removed the DATALAB_USER and settings.instanceUser code.

containers/datalab/content/setup-env.sh
gramster: Not really. If the auth is done by the node server and we write the auth token ourselves, we want to write it in a place that doesn't clobber the users credential file outside the container. So putting it in datalab folder is fine. But the user may want to use their external credential file to do auth and stop the node server from doing so, so they may want to point it at /content/.config.

sources/web/datalab/auth.ts
gramster: Done

gramster:
PTAL

containers/datalab/content/setup-env.sh
qimingj: I see. Makes sense if we want to support all these scenarios.

qimingj:
LGTM

qimingj:
LGTM.

Issue #819, opened 2016-05-09
Dockerfile still points to run-local.sh which does not exist anymore.


supriyagarg:

gramster:
Fixed in 63b88014

PR #821, opened 2016-05-09
Make the default entrypoint in the Dockerfile run.sh

supriyagarg:
This fixes the bug #819.


gramster:
LGTM

supriyagarg:
The PR #823 includes this change.

Issue #820, opened 2016-05-09
insert() used with one argument in containers/datalab/config/nbconvert.py


supriyagarg:

gramster:
Fixed in 79761571930dc07915a8552e05c3c8ab18906c57

PR #822, opened 2016-05-09
Updated nbconvert.py to use insert with 2 arguments.

supriyagarg:
Now '/datalab/nbconvert' is inserted into the first position.
This fixes the bug #820.


gramster:
I think we should use insert(0,...) instead; we want to take priority over any default templates.

supriyagarg:
Done.

gramster:
LGTM, thanks!

supriyagarg:
The PR #823 addresses the same bug using prepend.

PR #827, opened 2016-05-12
Make signing in to Google optional.

gramster:
Added a sign in button to the tree view instead of forcing auth on first request. Users can now use Datalab without signing in to Google (but signing in is needed before any GCP stuff can be used).

We probably want to change the button/icon to make it more obvious but at least the functionality
is there.

Fixed a couple of regressions that came in the big refactoring change before.

Fixed a bunch of imports in tests.

Folded in PR https://github.com/GoogleCloudPlatform/datalab/pull/815 as it was easier to do this one liner here than try to merge it with all the other changes.

### DIFF
--- a/sources/lib/datalab/datalab/context/_utils.py
+++ b/sources/lib/datalab/datalab/context/_utils.py
@@ -17,6 +17,11 @@
 import oauth2client.client
 import json
 import os
+import path
+
+
+def _in_datalab_docker():
+  return path.exists('/datalab') and os.getenv('DATALAB_ENV')
 
 
 def get_config_dir():
@@ -56,5 +61,10 @@ def get_credentials():
       for entry in creds['data']:
         if entry['key']['type'] == 'google-cloud-sdk':
           return oauth2client.client.OAuth2Credentials.from_json(json.dumps(entry['credential']))
+
+    if type(e) == oauth2client.client.ApplicationDefaultCredentialsError:
+      # If we are in Datalab container, change the message to be about signing in.
+      if _in_datalab_docker():
+        raise Exception('No application credentials found. Perhaps you should sign in.')
 
     raise e
--- a/sources/web/datalab/auth.ts
+++ b/sources/web/datalab/auth.ts
@@ -35,7 +35,6 @@ let appCredFile = gcloudDir + '/application_default_credentials.json';
 
 let scopes = [
   'https://www.googleapis.com/auth/userinfo.email',
-  //'https://www.googleapis.com/auth/userinfo.profile',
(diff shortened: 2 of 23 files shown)


sources/lib/datalab/datalab/context/_utils.py
```diff
@@ -19,6 +19,17 @@
 import os
 
 
+def _in_docker():
+  try:
+    for line in open('/proc/1/cgroup'):
+      if line.find(':/docker/') > 0:
+        return True
+      break
```
qimingj: If you have a "break" here, does it mean it only checks the first line?

gramster: Yes; they all have the same prefix so we only need to check the first

sources/web/datalab/auth.ts
```diff
@@ -145,24 +169,22 @@ export function handleAuthFlow(request: http.ServerRequest, response: http.Serve
         }
       });
     }
-  } else if (!oauth2Client && process.env.DATALAB_ENV == 'local' && !fs.existsSync(appCredFile)) {
+  } else if (!oauth2Client && !isSignedIn()) {
```
qimingj: Add "path.indexof('/signin') == 0" here, in case the function is called with another path?

sources/web/datalab/jupyter.ts
```diff
@@ -297,6 +298,10 @@ function responseHandler(proxyResponse: http.ClientResponse,
     if (path.indexOf('/tree') == 0) {
       // stripping off the /tree/ from the path
       templateData['notebookPath'] = path.substr(6);
+      if (process.env.DATALAB_ENV == 'local') {
+        templateData['isSignedIn'] = auth.isSignedIn().toString();
+        logging.getLogger().info('isSignedIn = ' + templateData['isSignedIn']);
```
qimingj: Too much logging?

gramster: Removed

sources/web/datalab/auth.ts
gramster: Glad you suggested that as I found an issue as a result. We need to handle the race condition of sign in being clicked twice, so if we are already signed in need to just redirect back to referer.

sources/lib/datalab/datalab/context/_utils.py
gramster: Based on our earlier discussion, will change this to check for DATALAB_ENV envvar and /datalab directory.

gramster:
PTAL

qimingj:
LGTM!

Issue #826, opened 2016-05-11
Feature: enable custom library additions 


Di-Ku:
This is a frequent ask: 
"How can I add a set of Python (or other) packages?"
Our current answer is to add a cell to a notebook and do pip install or whatever else you do with a shell command. That is a good workaround but doesn't really provide a good story across notebooks and restarts.

(Capturing discussion with Graham and Bradley)
A better way would be for us to execute a set of commands upon startup from some file external to the container. Interestingly, down the road, the same startup script could apply to someone running a remote kernel in GCE  (i.e. split between kernel & web-server)

gramster:
Also from discussion: use an optional environment variable that can be set
to point to a GCS bucket or file under /content and will run that as part
of startup.

parthea:
I would be interested to work on this feature request if it is not already assigned to another team member.

gramster:
Sure, if you want to give it a go. Some pointers:

- should handle a local file path or a GCS object URL
- you probably want to do it from containers/datalab/content/run.sh

We're about to change auth in the container to be optional (with a sign in
button in tree view, for if there are no cached credentials). So if the
file is in GCS, we should be robust in the face of auth failure.

nikhilk:
I'd propose that we prioritize just a local file (specifically a file in the datalab directory).

This file can exist in GCE as well (by virtue of the user placing it explicitly in the right path on the VM, just like they would for local scenarios) rather than having a local and GCS story in parallel.

nikhilk:
Additionally to the extent this looks like something we didn't invent that would be good... start with something like a requirements.txt file that can be used to do pip install on startup.

Separately we should document how to build derived docker containers, as repeated installs on each instantiation of the container gets frustrating after a few times.

gramster:
Sounds good; that makes it a lot simpler. We don't need an environment
variable then unless we want to use that as a way of supporting multiple
configs.

gramster:
Related to the building of derived containers, we could have the build
script do something similar - see if a file with some name is present and
if it is concatenate it to the base Dockerfile before building the
container.

gramster:
Done in b78afa7

PR #828, opened 2016-05-12
Add ability to create and run a customized image derived from the datalab image

parthea:
A first implementation to enable installation of custom python packages as part of the datalab image build script as described in #826 . 

If the file `'custompackages.txt'` exists in `'$REPO_DIR/containers/datalab'`, then install the python packages listed in the file.

Here is an example of the contents of custompackages.txt:

```
xlrd
```

We could easily modify this PR to also append a command that executes a more flexible install script (*.sh) to the Dockerfile, if the *.sh script exists.

Feedback is welcome.

### DIFF
--- a/containers/datalab/run-extended.sh
+++ b/containers/datalab/run-extended.sh
@@ -1,0 +1,78 @@
+#!/bin/sh
+# Copyright 2015 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Runs the docker container locally. Auth will happen in the browser when
+# the user first opens Datalab after accepting the EULA.
+
+# Passing in 'shell' flag causes the docker container to break into a
+# command prompt, which is useful for tinkering within the container before
+# manually starting the server.
+
+CONTENT=$HOME
+DOCKERIMAGE=datalab
+ENTRYPOINT="/datalab/run.sh"
+if [ "$1" != "" ]; then
+  if [ "$1" != "shell" ]; then
+    CONTENT=$1
+    shift
+  fi
+  if [ "$1" == "shell" ]; then
+    ENTRYPOINT="/bin/bash"
+  fi
+fi
+
+# Custom packages can be installed in datalab prior to run time. There are 2 options
+# available. A simple option is to add the packages to a pip requirements file. Create a file
+# called 'custom-packages.txt' and place it in the $REPO_DIR/containers/datalab folder. Use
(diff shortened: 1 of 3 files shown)


nikhilk:
It might be interesting to think of this in context of creating derived containers rather than variants of the datalab container itself.

Basically a Dockerfile for derived containers with and a run-extended.sh script to issue a docker build command before docker run.

The Dockerfile would look like:

    FROM gcr.io/cloud-datalab/datalab:latest
    RUN pip install ...

The goal is someone using this shouldn't need to clone the datalab repo, i.e. it works for datalab end-users. The user would modify this Dockerfile.

gramster:
I would think we need both. Users who are more familiar with Docker and
comfortable with that approach can build custom containers. But other users
may just want to do docker run with some URL, and not get into doing
container builds.

parthea:
Thanks for the feedback @gramster and @nikhilk . Would you mind taking another look at my most recent change which supports either user provided script with python packages or user provided docker file derived from datalab image? I would be happy to make further changes as required.

parthea:
In the custom installs from script approach I'm creating a docker file for the user.  Since I'm creating a Docker file, I also thought about the committing following refactored code where I use the same Dockerfile name and docker image name for both the custom install from script and custom docker file approach but I was concerned that I may accidentally overwrite an existing custom Dockerfile. 

```
if [ -f custom-packages.txt ] || [ -f Dockerfile-extended.in ];
then
    DOCKERFILE=Dockerfile-extended.in
    DOCKERIMAGE=datalab-extended

    if [ -f custom-packages.txt ];
    then
        # First create a new Docker file called Dockerfile-custom-packages. Start with the standard image
        echo 'FROM gcr.io/cloud-datalab/datalab:latest' > $DOCKERFILE

        # Add the script with a list of custom packages to the Dockerfile
        echo 'ADD custom-packages.txt /datalab/custom-packages.txt' >> $DOCKERFILE
        echo 'RUN pip install -r /datalab/custom-packages.txt' >> $DOCKERFILE
    fi

    # Build the customized docker image derived from the standard datalab image
    docker build -t $DOCKERIMAGE -f $DOCKERFILE .

    # TODO (Tony): Remove this line when gcr.io/cloud-datalab/datalab:latest is updated to a newer version that has /datalab/run.sh
    ENTRYPOINT="/datalab/run-local.sh"
fi
```

containers/datalab/run-extended.sh
```diff
@@ -0,0 +1,78 @@
...
+# limitations under the License.
+
+# Runs the docker container locally. Auth will happen in the browser when
+# the user first opens Datalab after accepting the EULA.
+
+# Passing in 'shell' flag causes the docker container to break into a
+# command prompt, which is useful for tinkering within the container before
+# manually starting the server.
+
+CONTENT=$HOME
```
parthea: In run.sh, we also set `CONTENT=$HOME` as the default directory.  I was thinking it would be better to set the CONTENT variable to `$REPO_DIR/content/datalab/` by default since that is where the samples, tutorials and intro folders reside.

gramster: Yes, $HOME is the folder *outside* the container that gets mapped in to /content/datalab *inside* the container. This should be something generic that the user controls and that does not rely on (for example) the user having a cloned Datalab repo; hence we use $HOME.

gramster:
LGTM, thanks! I'll merge it shortly.

PR #837, opened 2016-05-14
Bring back checkpoints and improve HTTP error messages.

gramster:
Bring back checkpoints. Currently there is just a single checkpoint at the time of last manual save.

Improved generation of error messages from HTTP requests.

(Issue https://github.com/GoogleCloudPlatform/datalab/issues/797)


qimingj:
LGTM

Issue #786, opened 2016-03-12
Potential security concern when running in local mode (for development purposes)


parthea:
Developers should be aware that by default Docker will expose the container port on any ip on the host machine. This is a greater concern if a developer has not setup a firewall (eg. iptables) and does not fully trust the network they are on (eg. public wifi ). Since authentication is disabled in local mode, anyone on the local network can connect using the datalab web interfance and run queries against the user's google account (since the .gcloud folder which contains authentication information is mounted from the host machine in the local option). 

I propose the following:

1. Add a security disclaimer in the wiki and in `$REPO_DIR/containers/datalab/run.sh`
and reference the solutions to restrict access mentioned in [Docker Advanced Networking](https://docs.docker.com/v1.5/articles/networking/)

2.  Modify the file `$REPO_DIR/containers/datalab/run.sh` to run docker with  `-p 127.0.0.1:host_port:container_port` instead of `host_port:container_port` 

From  [Docker Advanced Networking](https://docs.docker.com/v1.5/articles/networking/),

> You can see that Docker has exposed these container ports on 0.0.0.0, the wildcard IP address that will match any possible incoming port on the host machine. If you want to be more restrictive and only allow container services to be contacted through a specific external interface on the host machine, you have two choices. When you invoke docker run you can use either -p IP:host_port:container_port or -p IP::port to specify the external interface for one particular binding.

> Or if you always want Docker port forwards to bind to one specific IP address, you can edit your system-wide Docker server settings (on Ubuntu, by editing DOCKER_OPTS in /etc/default/docker) and add the option --ip=IP_ADDRESS. Remember to restart your Docker server after editing this setting.

gramster:
Local mode is changing to not use mounted gcloud info. This will still be an issue though. Eventually a running kernel will have associated credentials and if someone opens an existing running session on a different machine they could have access. Locking down the ports may be the only good solution.

gramster:
I've been trying -p 127.0.0.1 but haven't managed to make it work yet. Not sure if that is due to other corp restrictions we have; need to test it at home.

parthea:
I just tested it again and `-p 127.0.0.1:8081:8080` worked but I was not up to date with master. Here is a diff of run.sh.

```
tony@TonyLinuxDesktop:~/datalab-parthea/containers/datalab$ git diff run.sh
diff --git a/containers/datalab/run.sh b/containers/datalab/run.sh
index 50651db..cadb560 100755
[...]

PR #848, opened 2016-05-16
Bind to localhost only so docker does not expose the port.

qimingj:


parthea:
LGTM! I have been using this configuration for a month now.

We could also refer users to the [Docker Advanced Networking](https://docs.docker.com/v1.5/articles/networking/) section as mentioned in #786 . The reference could be made available in the wiki, or even as a comment above the docker run command (in case users want to bind to a different interface, for example remote server access).

gramster:
LGTM although this did not work for me on corp network. Did you try it Bradley?

parthea:
Will this configuration work when datalab is installed on a GCE instance? (I haven't tried it)

gramster:
Tested again; does not work on corp on Macs at least.

Probably not an issue on GCE as the GCE instance is firewalled anyway and you have to use an SSH tunnel. That is, we probably don't need it for GCE.

nikhilk:
You do want it in GCE as well, in case someone inadvertently opens the port on the GCE firewall. This is a requirement from security folks.

gramster:
This looks relevant: https://forums.docker.com/t/using-localhost-for-to-access-running-container/3148

qimingj:
gram, thanks. Will do more investigation. Basically, on Linux we need the docker lever ip binding  (the change helps). On Mac, we need vitualbox level ip binding.

PR #850, opened 2016-05-18
Bind to 127.0.0.1 for linux for security reasons.

qimingj:


parthea:
@qimingj 

In #828, I added an additional run script (`run-extended.sh`) which allows users to create their own container which is derived from the datalab image. Based on #828, we would also need to update `run-extended.sh` to include the changes in this PR.

Thinking about the changes in #828 again, I'm wondering if I should remove `run-extended.sh` , and add the code for building the custom container directly into run.sh. The primary reason for this change is to avoid having to make the same code changes in 2 places (`run-extended.sh` and `run.sh`). Should I create a PR to merge the features in `run-extended.sh` into `run.sh` ?

gramster:
Tony, if you do, there is a change to make. Change the line where you do the FROM to:

        # First create a new Docker file called Dockerfile-custom-packages. Start with the standard image
        # TODO: at some point the local Datalab container will be tagged 'latest' rather than 'local'
        # and the line below should change.
        echo 'FROM gcr.io/cloud-datalab/datalab:local' > Dockerfile-custom-packages

And get rid of your code to do use run-local.sh. The datalab:latest container is for managed VMs and so we shouldn't be basing off that. Later we may change the tag to latest again but for now we are using :local

qimingj:
Good point. Perhaps let's wait for #854.

Added the same logic for run-extended.sh. Note that this is only for users who clone repo and run their own image. For users who run "docker run ..." to start Datalab, we'll also update instructions to not expose Datalab on network (on mac/Windows, virtualbox command needs to be updated).

parthea:
> Or you can wait for #854 and I will include this change myself.

Sounds good! I will wait for your changes in #854 to avoid duplicating work.

Issue #31, opened 2014-09-17
Support for configuring log levels at runtime


drewbryant:
Add support for a query string debug flag that can increase the logging verbosity of a running instance without rebuilding.

For example:
<host>/notebooks/123?debug=true

Which would override the default logging levels to enable full debug output.  This is useful for diagnosing a deployed instance without requiring a code change to bump (and then another change later to reduce) log levels.

PR #852, opened 2016-05-19
Get rid of the datalab package; it lives in its own repo now.

gramster:
Also clean up some detritus in /content. Moved the stray notebook to the notebook repo and we don't need publish.sh any more.


nikhilk:
Do we really need to delete anything from this repo? Lets get the other one working, and then we can decide how to EoL this one at the right time.

gramster:
The other one is working.

ojarjur:
I can't speak to the readiness of the new repository, but the changes in this pull request look good to me.

ojarjur:
LGTM

I've verified that (with https://github.com/googledatalab/datalab/pull/31), this works for all of the bundled sample notebooks (except the programming language correlations one, which tries to query a no-longer extant table).

qimingj:
LGTM.

PR #853, opened 2016-05-19
Fix a build failure that was triggered by `cp` skipping a directory.

ojarjur:
The command that does a wildcard copy of files into the nbconvert
directory needs to include the contents of the `codemirror`
subdirectory. Failing to include the "-R" flag was causing that
copy to fail with an error message that the subdirectory was being
ignored.

This change fixes that by adding the required "-R" flag.


gramster:
LGTM! Thanks!

PR #855, opened 2016-05-20
Add support for a customization script in mapped-in volume.

gramster:
Change the Docker extensibility script to use datalab:local instead of
datalab:latest for now.

### DIFF
--- a/containers/datalab/content/run.sh
+++ b/containers/datalab/content/run.sh
@@ -29,5 +29,14 @@ fi
 # Setup environment variables.
 . /datalab/setup-env.sh
 
+# Run the user's custom extension script if it exists. To avoid platform issues with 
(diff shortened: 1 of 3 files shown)


containers/datalab/content/run.sh
```diff
@@ -29,5 +29,14 @@ fi
 # Setup environment variables.
 . /datalab/setup-env.sh
 
+# Run the user's custom extension script if it exists. To avoid platform issues with 
+# execution permissions, line endings, etc, we create a local sanitized copy.
+if [ -f /content/datalab/startup.sh ]
```
qimingj: startup.sh may show up in the home notebook list? Can we make it hidden?

Perhaps it is time to move the home dir to /content/datalab/home or /content/datalab/content?

gramster: Moved to .config

PR #858, opened 2016-05-23
Fix a deployment issue that when project name contains "project"

qimingj:
deployment fails.
Update Tensorflow and gcloud.
Modify web server to make it work on type script compiler with strict mode.


gramster:
LGTM

qimingj:
merged into datalab_managed

PR #864, opened 2016-05-27
Fix two issues that prevented logging in from a local instance.

ojarjur:
The first issue was around the creation of the .config directory.
We had code in multiple places that tried to create it, and in one
of those cases it tried to create that directory repeatedly, resulting
in failure messages in the logs after the first attempt.

This change fixes that by making sure that only the EULA accepted
handler creates that directory, and only in the case that it does
not already exist.

The other issue was the OAuth callback handler, which did not make
sure the OAuth client instance existed before attempting to use it.

This change fixes that by adding that check.

### DIFF
--- a/sources/web/datalab/auth.ts
+++ b/sources/web/datalab/auth.ts
@@ -22,6 +22,7 @@ import fs = require('fs');
 import google = require('googleapis');
 import http = require('http');
 import logging = require('./logging');
+import url = require('url');
 
 
 var oauth2Client: any = undefined;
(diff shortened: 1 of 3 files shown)


sources/web/datalab/auth.ts
```diff
@@ -128,6 +128,17 @@ export function isSignedIn(): boolean {
...
 }
 
+function setOauth2Client(parsed_url: any): void {
+  if (!oauth2Client) {
+    var OAuth2:any = google.auth.OAuth2;
+    // TODO(ojarjur): Ideally, we would get the host and port from the parsed_url
+    // rather than hard-coding them. However, the client ID and secret we are using
+    // are limited to localhost only. We should consider making this configurable, or
+    // using the OAuth flow for non-web applications.
+    oauth2Client = new OAuth2(clientId, clientSecret, 'http://localhost:8081/oauthcallback');
```
qimingj: Can we initialize it in the definition https://github.com/GoogleCloudPlatform/datalab/blob/master/sources/web/datalab/auth.ts#L27?

qimingj:
Thanks for the fix! One observation.

sources/web/datalab/auth.ts
ojarjur: That would be doable, but would prevent us from being able to use the parsed_url to build the callback in the future (as opposed to hardcoding 'localhost:8081').

I'm a little torn about whether or not to keep this flexibility that we aren't using yet, but I suspect we'll want to be able to change at least the port number (without recompiling) in the near future.

qimingj: Good point. I think we are restricted on the the callback url. The port might be changed but it seems it won't be changed at run time. If someday we want to pass in a port number, we can have a init() function which takes the port as parameter. Thoughts?

ojarjur: I don't think the "init-method" approach will work. At least, I tried it and hit a roadblock.

The issue is that the port number we need to use is *not* the port number of the running server, but rather the port number that Docker is listening to.

So, for instance, we currently run the server at port 8080 *inside* the Docker container, however, we have Docker map port 8081 of the host machine to port 8080 of the Docker container.

That means the port we need to include in the OAuth callback is the 8081, not 8080. That, in turn, means that we can't set up the OAuth client at server setup. Instead, we have to wait for an incoming request so that we can parse the port number from the referer string.

qimingj: I see. We do plan to use the same port for docker and server.

But I am not sure whether we can get the port from the client request (in this case, 8081) from our web server. Can we?

ojarjur: Actually, I've added a commit that uses the request to get the port number. We have to dig into the 'referer' query parameter, but we can get it.

I've tested this by changing the port mapping from 8081 to 4242, and I was able to login with the redirect working correctly.

qimingj: Sounds good. Just want to point out that referer may not always work (such as user visiting /signin directly). Perhaps explain the issue in more details in the comment.

qimingj:
LGTM. Thanks!

sources/web/datalab/auth.ts
ojarjur: Added a TODO to explain the hole in the current logic.

PR #869, opened 2016-06-01
Improve the logic for determining the port number of the server.

ojarjur:
Previously, we relied on the "referer" query parameter being passed
in (which is the case when going through the login flow in the UI,
but not when going directly to the "/signin" page).

This changes the logic to instead rely on the request's "host"
header field, which should always be set regardless of whether or
not the "referer" query parameter is set.


gramster:
LGTM

PR #874, opened 2016-06-08
Fix a bug in the 'Convert to Python' functionality.

ojarjur:
The issue was that when you tried to convert a notebook to python,
the server returned a cryptic error message with the name of the
default HTML template.

The root cause was that when we set the 'template_file' config
variable for exporting to HTML, we did so for *every* exporter,
and not just the HTMLExporter.

This change fixes that by only setting that config variable on
the HTMLExporter.


qimingj:
LGTM!

PR #876, opened 2016-06-14
Improve the visibility around setting a default project.

ojarjur:
A number of the datalab libraries require either:
1. Explicitly passing in a Context object with a project_id set
or
2. Setting the project_id in the default Context instance.

When neither of those were done, the subsequent API calls
could fail with very confusing error messages (such as "Not found").

This changes the libraries so that an error is surfaced earlier,
with an explicit explanation of what has gone wrong, and with
instructions on how to fix the issue.

Specifically, attempts to read the 'project_id' property of
a Context object will raise an error if that property has not
been set. The error message will explain that the 'project_id'
has not been set, and will suggest using the '%projects set'
cell magic to set a default project.

This change also fixes the '%projects set' cell magic so that
it actually sets the project ID to be the passed in argument,
rather than a singleton list of the literal string 'id'.

### DIFF
--- a/sources/lib/datalab/datalab/context/_context.py
+++ b/sources/lib/datalab/datalab/context/_context.py
@@ -54,6 +54,8 @@ class Context(object):
     Returns:
       The current project id to associate with API requests.
     """
+    if not self._project_id:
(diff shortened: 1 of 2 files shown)


qimingj:
LGTM

PR #877, opened 2016-06-14
Add the sign-in/sign-out buttons to the notebooks UI.

ojarjur:
There are some sample notebooks which require credentials to run
(because they call into external APIs). If you try to run those
notebooks while signed out, then they will produce an error
message helpfully suggesting that you sign.

However, the sign-in button was only being shown on the tree-view
page and not on the notebook-view page.

This meant that you had to navigate back to the tree-view page,
sign in, and then go back to the notebook to retry running the
code.

This change fixes that by adding the sign-in/sign-out buttons
to the notebook-view page in the exact same place they appear
for the tree-view page.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -293,14 +293,14 @@ function responseHandler(proxyResponse: http.ClientResponse,
       configUrl: appSettings.configUrl,
       baseUrl: '/'
     };
+    if (process.env.DATALAB_ENV == 'local') {
+      templateData['isSignedIn'] = auth.isSignedIn().toString();
+    }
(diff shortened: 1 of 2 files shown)


qimingj:
LGTM! Thanks.

gramster:
The reason this was not done before was that it could result in lost cell
content. You probably need to save the notebook if it has unsaved changes.

ojarjur:
> The reason this was not done before was that it could result in lost cell content. You probably need to save the notebook if it has unsaved changes.

Thanks for the tip, Graham. I'll make sure the sign-in/sign-out buttons save the notebook if it has changes.

PR #879, opened 2016-06-17
Make the signin/signout buttons save notebook changes.

ojarjur:
This change makes the sign in/out buttons save any changes to the
current notebook (if we are on a notebook page) before redirecting
the user.

This prevents users accidentally losing unsaved work if they try
to sign-in from the notebook view page.

This fixes a potential issue reported by @gramster on #877


gramster:
LGTM

qimingj:
LGTM

PR #880, opened 2016-06-22
Add a new template for the file editing page.

ojarjur:
We previously were serving the file editing page using the
built-in Jupyter UI. That meant that it looked wildly different
from the rest of Datalab, and was missing the Datalab-specific
UI elements like the sign-in and feedback buttons.

This version of the editor page is functional, but not yet
feature complete. The remaining missing features will be added
in later commits.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -55,6 +55,7 @@ var callbackManager: callbacks.CallbackManager = new callbacks.CallbackManager()
  */
 var templates: common.Map<string> = {
   'tree': fs.readFileSync(path.join(__dirname, 'templates', 'tree.html'), { encoding: 'utf8' }),
+  'edit': fs.readFileSync(path.join(__dirname, 'templates', 'edit.html'), { encoding: 'utf8' }),
   'nb': fs.readFileSync(path.join(__dirname, 'templates', 'nb.html'), { encoding: 'utf8' })
 };
 
@@ -285,7 +286,7 @@ function responseHandler(proxyResponse: http.ClientResponse,
   // Set a cookie to provide information about the project and authenticated user to the client.
   // Ensure this happens only for page requests, rather than for API requests.
   var path = url.parse(request.url).pathname;
-  if ((path.indexOf('/tree') == 0) || (path.indexOf('/notebooks') == 0)) {
+  if ((path.indexOf('/tree') == 0) || (path.indexOf('/notebooks') == 0) || (path.indexOf('/edit') == 0)) {
     var templateData: common.Map<string> = {
       feedbackId: appSettings.feedbackId,
       versionId: appSettings.versionId,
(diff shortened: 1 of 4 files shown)


sources/web/datalab/static/datalab.js
```diff
@@ -714,6 +714,36 @@ function initializeNotebookApplication(ipy, notebook, events, dialog, utils) {
 }
 
 
+function initializeEditApplication(ipy, editor) {
+  function navigateAlternate(alt) {
+    var url = document.location.href.replace('/edit', alt);
+    url = url + '?download=true';
+
+    if (!editor.clean) {
+      var w = window.open('');
```
qimingj: Is this needed?

sources/web/datalab/jupyter.ts
```diff
@@ -304,8 +305,14 @@ function responseHandler(proxyResponse: http.ClientResponse,
...
       page = 'tree';
-    }
-    else {
+    } else if (path.indexOf('/edit') == 0) {
+      // stripping off the /edit/ from the path
+      templateData['filePath'] = path.substr(6);
+      templateData['fileName'] = path.substr(path.lastIndexOf('/') + 1);
+
+      sendTemplate('edit', templateData, response);
+      page = 'edit';
```
qimingj: This is not needed? Same for '/tree' and '/notebooks' blocks.

sources/web/datalab/static/datalab.js
ojarjur: Yes, that's (presumably) what it was for.

I was following the pattern in the `initializeNotebookApplication` method's `navigateAlternate`, but screwed that up and failed to use the `w` variable.

Not using the blank window does not seem to have impaired usability, so I'm just going to remove it entirely.

sources/web/datalab/jupyter.ts
ojarjur: You're right; it seems to have been a half-completed cleanup to remove the duplicated calls to `sendTemplate`.

Rather than delete the variable, I'm going to complete the cleanup and remove the duplicated calls to `sendTemplate`.

sources/web/datalab/static/datalab.js
```diff
@@ -714,6 +714,36 @@ function initializeNotebookApplication(ipy, notebook, events, dialog, utils) {
 }
 
 
+function initializeEditApplication(ipy, editor) {
+  function navigateAlternate(alt) {
+    var url = document.location.href.replace('/edit', alt);
+    url = url + '?download=true';
```
qimingj: Would it be possible that url contains some query strings already? Maybe we can get the file name after first '/edit/' until '?', just in case?

qimingj:
A few observations. Thanks!

sources/web/datalab/static/datalab.js
ojarjur: Done.

qimingj:
LGTM!

PR #882, opened 2016-06-23
Fix a bug in the template serving.

ojarjur:
PR #880 introduced a bug in the template rendering for the notebook
UI where the name of the template was wrong (`notebook` instead of
`nb`).

This change fixes that.


qimingj:
LGTM

PR #883, opened 2016-06-23
Add support for connecting to a kernel gateway.

ojarjur:
This change allows Datalab to use a [kernel gateway](https://github.com/jupyter/kernel_gateway) for running the notebook kernels.

The setting for whether or not to use a gateway is based on the `KG_URL` environment variable, which defines the URL of the gateway (if any) to use.

If that environment variable is set, then *all* kernels are run using that gateway. Otherwise, all kernels are run locally within the Datalab container.

For example:

    KG_URL=http://localhost:8080 ./containers/datalab/run.sh

will run Datalab with all kernels being run by a kernel gateway running on localhost:8080.

... whereas:

    ./containers/datalab/run.sh

will run Datalab will all kernels running inside of the Datalab container.

Note: This uses the [nb2kg server extension](https://github.com/jupyter/kernel_gateway_demos/tree/master/nb2kg), which is demo code. As such, we require that the user explicitly opt-in to using it.

### DIFF
--- a/containers/datalab/content/setup-env.sh
+++ b/containers/datalab/content/setup-env.sh
@@ -25,3 +25,7 @@ then
   export CLOUDSDK_CONFIG=/content/datalab/.config
 fi
 
+if [ -n "${EXPERIMENTAL_KERNEL_GATEWAY_URL}" ]
+then
+  export KG_URL="${EXPERIMENTAL_KERNEL_GATEWAY_URL}"
+fi
--- a/containers/datalab/run-extended.sh
+++ b/containers/datalab/run-extended.sh
@@ -84,4 +84,5 @@ docker run -it --entrypoint=$ENTRYPOINT \
   -v "$CONTENT:/content" \
   -e "PROJECT_ID=$PROJECT_ID" \
   -e "DATALAB_ENV=local" \
+  -e "EXPERIMENTAL_KERNEL_GATEWAY_URL=${EXPERIMENTAL_KERNEL_GATEWAY_URL}" \
   $DOCKERIMAGE
--- a/containers/datalab/run.sh
+++ b/containers/datalab/run.sh
@@ -49,4 +49,5 @@ docker run -it --entrypoint=$ENTRYPOINT \
   -v "$CONTENT:/content" \
   -e "PROJECT_ID=$PROJECT_ID" \
   -e "DATALAB_ENV=local" \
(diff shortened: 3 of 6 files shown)


containers/datalab/run.sh
```diff
@@ -49,4 +49,5 @@ docker run -it --entrypoint=$ENTRYPOINT \
   -v "$CONTENT:/content" \
   -e "PROJECT_ID=$PROJECT_ID" \
   -e "DATALAB_ENV=local" \
+  -e "KG_URL=${KG_URL}" \
```
parthea: Should this environment variable be added to `containers/datalab/run-extended.sh` also ? https://github.com/GoogleCloudPlatform/datalab/blob/master/containers/datalab/run-extended.sh

ojarjur: Yes, I think it should.

Thanks for catching that!

containers/datalab/run-extended.sh
```diff
@@ -84,4 +84,5 @@ docker run -it --entrypoint=$ENTRYPOINT \
   -v "$CONTENT:/content" \
   -e "PROJECT_ID=$PROJECT_ID" \
   -e "DATALAB_ENV=local" \
+  -e "EXPERIMENTAL_KERNEL_GATEWAY_URL=${EXPERIMENTAL_KG_URL}" \
```
qimingj: Maybe use the same name "EXPERIMENTAL_KERNEL_GATEWAY_UR" for both env var?

containers/datalab/content/setup-env.sh
```diff
@@ -25,3 +25,7 @@ then
   export CLOUDSDK_CONFIG=/content/datalab/.config
 fi
 
+if [ -n "${EXPERIMENTAL_KERNEL_GATEWAY_URL}" ]
+then
+  export KG_URL="${EXPERIMENTAL_KERNEL_GATEWAY_URL}"
```
qimingj: Do we have to use two env var (EXPERIMENTAL_KERNEL_GATEWAY_URL and KG_URL) here? Can we just use EXPERIMENTAL_KERNEL_GATEWAY_URL?

ojarjur: Unfortunately, no.

The server extension specifically requires the environment variable be named "KG_URL".

qimingj: I see. Thanks.

containers/datalab/run-extended.sh
ojarjur: Done.

qimingj:
LGTM. Cool!

PR #885, opened 2016-06-28
Fix a bug in the port selection for the nested Jupyter servers.

### LINKED ISSUE #884: Datalab is loading on different (incorrect) user workspace
Datalab is randomly loading under different users workspace within the serving VM for the datalab app engine (datalab_main).

I am attaching an image of datalab for which I list the workspace folder and you can see that it is on the workspace of my colleague.

We suspect this has been happening for a while since we have randomly disappearing files from workspaces. We also realize that these VMs have some background linux process to synchronize source code with git so perhaps that is the cause also of the missing files.

In any case, this is an issue and it is affecting our work now.
screen shot 2016-06-27 at 11 18 08 am

ojarjur:
Previously, Datalab was picking the next port that it wanted Jupyter
to run on, passing that to the launched Jupyter process using the
'--port' flag, and then assuming that the server did in fact start
listening on that port.

However, if the port selected was already in use, then Jupyter would
automatically try the next port in sequence. That could cause an issue
where the one Jupyter server (for one user) was listening on a port
that we thought was being used by the Jupyter server for a different
user.

This change fixes that bug by doing the following:

1. Verifying that the next port is free before we try to start a
   Jupyter server listening on it.
2. Telling the Jupyter server to not try any additional ports if
   the one specified was not available (so it will instead die in
   that situation).

The combination of these two steps effectively moves the retry
logic for picking a port out of the Jupyter process and in to the
wrapping Node server.

There are still two failure cases that can manifest even with this
change.

The first is that if the attempt to start a Jupyter server
fails more than 10 times, then the retry logic will give up and
an internal error will be reported to the user. This failure mode
always existed, but now we will be properly tracking it.

The second is a race condition where another process grabs the
requested port between the time when we verify that it is free
and the time that the launched Jupyter server starts up. If that
happens, then the Jupyter server will kill itself, and it will
be removed from the users->Jupyter servers map. However, the
request that caused us to spin up a Jupyter server to begin
with will be forwarded to the process that took hold of the
assigned port.

This failure case should be difficult enough to trigger that it
will not occur in practice.

We believe this will fix the issue reported in #884

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -36,6 +36,8 @@ interface JupyterServer {
   notebooks: string;
   childProcess?: childProcess.ChildProcess;
   proxy?: httpProxy.ProxyServer;
+
+  portResolved: ()=>void;
 }
 
 /**
@@ -66,6 +68,7 @@ var appSettings: common.Settings;
 
 function pipeOutput(stream: NodeJS.ReadableStream, port: number, error: boolean) {
   stream.setEncoding('utf8');
+
   stream.on('data', (data: string) => {
     // Jupyter generates a polling kernel message once every 3 seconds
     // per kernel! This adds too much noise into the log, so avoid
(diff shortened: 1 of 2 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -81,67 +88,78 @@ function pipeOutput(stream: NodeJS.ReadableStream, port: number, error: boolean)
  * Starts the Jupyter server, and then creates a proxy object enabling
  * routing HTTP and WebSocket requests to Jupyter.
  */
-function createJupyterServer(userId: string): JupyterServer {
+function createJupyterServer(userId: string, resolved: (server: JupyterServer)=>void, rejected:()=>void) {
```
qimingj: Can we move the "resolved" and "rejected" logic into this function? We don't have different "resolved" or "rejected" functions anyway.

qimingj: The rejected logic can be a loop inside this function, which seems easier to understand compared to recursive calls.

sources/web/datalab/jupyter.ts
```diff
@@ -64,14 +66,19 @@ var templates: common.Map<string> = {
  */
 var appSettings: common.Settings;
 
-function pipeOutput(stream: NodeJS.ReadableStream, port: number, error: boolean) {
+function pipeOutput(server: JupyterServer, stream: NodeJS.ReadableStream, port: number, error: boolean) {
   stream.setEncoding('utf8');
+  var listeningMessage = `The Jupyter Notebook is running at: http://127.0.0.1:${port}`;
```
qimingj: Can we wait for the port to be used to declare server is started successfully? It's not perfect but relying on an stdout message seems a bit more shaky.

qimingj:
Good findings! Two observations on the fix.

sources/web/datalab/jupyter.ts

sources/web/datalab/jupyter.ts
ojarjur: Moving the code for the 'resolved' and 'rejected' methods into this one has been done, but putting the logic into a for loop has not.

Specifically, there's an issue with trying to use a for loop, as this method is using some other (externally defined) methods that are already defined to be asynchronous, so we can't use them in a blocking fashion.

sources/web/datalab/jupyter.ts
```diff
@@ -64,14 +66,19 @@ var templates: common.Map<string> = {
...
+
   stream.on('data', (data: string) => {
     // Jupyter generates a polling kernel message once every 3 seconds
     // per kernel! This adds too much noise into the log, so avoid
     // logging it.
 
     if (data.indexOf('Polling kernel') < 0) {
+      if (data.indexOf(listeningMessage) >= 0) {
```
qimingj: This should be removed, right?

sources/web/datalab/jupyter.ts
```diff
@@ -81,67 +88,93 @@ function pipeOutput(stream: NodeJS.ReadableStream, port: number, error: boolean)
...
+        // Create the proxy.
+        var proxyOptions: httpProxy.ProxyServerOptions = {
+          target: 'http://127.0.0.1:' + port
+        };
+
+        server.proxy = httpProxy.createProxyServer(proxyOptions);
+        server.proxy.on('proxyRes', responseHandler);
+        server.proxy.on('error', errorHandler);
+
+        tcp.waitUntilUsed(server.port).then(
```
qimingj: Ah. Sorry for the miscommunication. What I meant is that we can rely on tcp.waitUntilUsed only, thus no need to wait for the stdout message 'The Jupyter Notebook is running at:...'. "tcp.waitUntilUsed(server.port).then" can be moved right after the jupyter process is started. Also no need for "server.portResolved" and the code can be simplified by a little bit. Thoughts?

ojarjur: > we can rely on tcp.waitUntilUsed only

That by itself is not sufficient. In fact, I didn't add the tcp.waitUntilUsed code in this change... it was already there. All I did was move it inline so the logic is a bit easier to follow.

The reason that is not sufficient is that it will only tell us that some process is listening on that port, not whether or not that process is the one we started.

That's essentially what we are seeing in the bug; the port is in fact in use, but it is in use by a different process than the one we think it is in use by.

sources/web/datalab/jupyter.ts
ojarjur: No, I don't think this can be removed. This is arguably the most important safeguard that is being added.

Specifically, this check ensures that we do not send any traffic to the proxy until after we confirm that the proxy is pointing at the Jupyter server we think it is.

sources/web/datalab/jupyter.ts
qimingj: The fact that the port was free before and after we spawned the jupyter process it is used should be safe enough. If you want to be even safer we can possibly call some libraries to make sure the port is used by the jupyter process we started.

The stdout message may not be reliable in the sense that a new jupyter version may choose to modify the message or removing the message and thus break us. It is not supposed to be used as part of the interface.

ojarjur: I agree that relying on the exact log message is brittle (although we do pin the image to a specific version of Jupyter).

However, I don't think that the port being free before we launch the process is enough of a safety check.

I'll look for a better way to confirm that the Jupyter server is listening on the port we think it is, but I prefer a safety check with false rejects over one with false accepts.

qimingj: Sounds good.

If we do check the stdout, we may also need to watch out for the case that multiple "The Jupyter Notebook is running at..." are produced. We should only check the first one to avoid calling "server.portResolved()" multiple times.

sources/web/datalab/jupyter.ts
```diff
@@ -81,67 +88,96 @@ function pipeOutput(stream: NodeJS.ReadableStream, port: number, error: boolean)
...
+        server.proxy.on('error', errorHandler);
+
+        tcp.waitUntilUsed(server.port, 100, 3000).then(
+          function() {
+            jupyterServers[userId] = server;
+            logging.getLogger().info('Jupyter server started for %s.', userId);
+            callbackManager.invokeAllCallbacks(userId, null);
+          },
+          function(e) {
+            logging.getLogger().error(e, 'Failed to start Jupyter server for user %s.', userId);
```
qimingj: Double check...we don't retry on failures here?

ojarjur: > we don't retry on failures here?

No, not once we get to this point. We're only retrying failures in picking an unused port; once we've picked an unused port, any failures after that are just cascaded upward.

sources/web/datalab/jupyter.ts
qimingj: Now we have two checks --- make sure stdout prints the expected message, and make sure the tcp port is used. Having both is perhaps too much and making the code too complicated.

I still think wait_for_port_free --> start_jyputer_process -> wait_for_port_used is better than relying on stdout message (is it possible the message appears before we start listening to it?). But will leave it to you.

qimingj:
LGTM. Thanks!

sources/web/datalab/jupyter.ts
ojarjur: > is it possible the message appears before we start listening to it?

Ah, there is a risk here... the message could be output to the stderr before we start reading the (streamed) stderr.

OK; this won't work reliably.

I'll remove this check for now and try to come up with a more reliable one later.

qimingj: Yes. Another reason I don't like it is that the logic is not in the same function. If we do wait_for_port_free --> start_jyputer_process -> wait_for_port_used (which may not be the safest either, I agree), we could put all logic inside the same function, and is easier to understand.

ojarjur: I've removed this check.

PTAL

qimingj:
LGTM!

PR #886, opened 2016-06-28
Fix a bug in the port selection for the nested Jupyter servers.

### LINKED ISSUE #884: Datalab is loading on different (incorrect) user workspace

ojarjur:
Previously, Datalab was picking the next port that it wanted Jupyter
to run on, passing that to the launched Jupyter process using the
'--port' flag, and then assuming that the server did in fact start
listening on that port.

However, if the port selected was already in use, then Jupyter would
automatically try the next port in sequence. That could cause an issue
where the one Jupyter server (for one user) was listening on a port
that we thought was being used by the Jupyter server for a different
user.

This change fixes that bug by doing the following:

1. Verifying that the next port is free before we try to start a
   Jupyter server listening on it.
2. Telling the Jupyter server to not try any additional ports if
   the one specified was not available (so it will instead die in
   that situation).

The combination of these two steps effectively moves the retry
logic for picking a port out of the Jupyter process and in to the
wrapping Node server.

There are still two failure cases that can manifest even with this
change.

The first is that if the attempt to start a Jupyter server
fails more than 10 times, then the retry logic will give up and
an internal error will be reported to the user. This failure mode
always existed, but now we will be properly tracking it.

The second is a race condition where another process grabs the
requested port between the time when we verify that it is free
and the time that the launched Jupyter server starts up. If that
happens, then the Jupyter server will kill itself, and it will
be removed from the users->Jupyter servers map. However, the
request that caused us to spin up a Jupyter server to begin
with will be forwarded to the process that took hold of the
assigned port.

This failure case should be difficult enough to trigger that it
will not occur in practice.

We believe this will fix the issue reported in #884

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -36,6 +36,8 @@ interface JupyterServer {
   notebooks: string;
   childProcess?: childProcess.ChildProcess;
   proxy?: httpProxy.ProxyServer;
+
+  portResolved: ()=>void;
 }
 
 /**
@@ -65,6 +67,7 @@ var appSettings: common.Settings;
 
 function pipeOutput(stream: NodeJS.ReadableStream, port: number, error: boolean) {
   stream.setEncoding('utf8');
+
   stream.on('data', (data: string) => {
     // Jupyter generates a polling kernel message once every 3 seconds
     // per kernel! This adds too much noise into the log, so avoid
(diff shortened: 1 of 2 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -137,7 +137,16 @@ function createJupyterServer(userId: string, resolved: (server: JupyterServer)=>
         server.proxy.on('proxyRes', responseHandler);
         server.proxy.on('error', errorHandler);
 
-        resolved(server);
+        tcp.waitUntilUsed(server.port).then(
```
qimingj: Maybe set a larger timeout value? Default is only 300ms and not sure Jupyter is quick enough to start up especially in a potentially heavy loaded VM.

qimingj:
LGTM.

PR #894, opened 2016-07-01
Change the logic for picking the next Jupyter port more intelligent.

### LINKED ISSUE #891: The `nextJupyterPort` variable should always be initialized to a free port
Currently, at startup of the NodeJS server, the `nextJupyterPort` variable is initialized to 9000. When we try to launch a Jupyter server, we will verify that port is free and retry (with the next port) if it isn't, but that will count against the retry count for that creation attempt.

It would be better if the logic to get the next Jupyter port just always returned a port that was (at that time) unused.

ojarjur:
Previously, the logic was just to increment the next port number,
and then try to launch a Jupyter server on that port if it was
free. If that failed at any step, then we would retry, but with
a counter for the max number of retries decremented.

This meant that the next port already being in use counted against
the retry count, and that should not be the case.

With this change, the search for a free port will not count against
the the retry count.

This fixes #891

### DIFF
--- a/externs/ts/node/tcp-port-used.d.ts
+++ b/externs/ts/node/tcp-port-used.d.ts
@@ -18,7 +18,11 @@ declare module 'tcp-port-used' {
     then(resolved: () => void, rejected: (error: Error) => void): void;
   }
 
-  export function waitUntilFree(port: number): SimplePromise;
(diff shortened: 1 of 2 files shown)


qimingj:
LGTM

PR #895, opened 2016-07-01
Change the logic for picking the next Jupyter port more intelligent.

ojarjur:
Previously, the logic was just to increment the next port number,
and then try to launch a Jupyter server on that port if it was
free. If that failed at any step, then we would retry, but with
a counter for the max number of retries decremented.

This meant that the next port already being in use counted against
the retry count, and that should not be the case.

With this change, the search for a free port will not count against
the the retry count.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -45,6 +45,42 @@ interface JupyterServer {
  */
 var jupyterServers: common.Map<JupyterServer> = {};
 var nextJupyterPort = 9000;
+var portRetryAttempts = 500;
+
+/**
+ * Get the next available port and pass it to the given `resolved` callback.
+ */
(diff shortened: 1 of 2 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -47,6 +47,25 @@ var jupyterServers: common.Map<JupyterServer> = {};
 var nextJupyterPort = 9000;
 
 /**
+ * Get the next available port and pass it to the given `resolved` callback.
+ */
+function getNextJupyterPort(resolved: (port: number)=>void, failed: (error: Error)=>void) {
+   var port = nextJupyterPort;
+   nextJupyterPort++;
+
+   tcp.check(port, '127.0.0.1').then(
```
qimingj: Can we put an upper limit on the number of ports we tried? For example from 9000~9200? There might be cases that the tcp.check call fails on any ports given 127.0.0.1, and we will end up in a dead loop.

qimingj:
LGTM!

PR #898, opened 2016-07-02
Make the Datalab server kill abandoned Jupyter servers.

### LINKED ISSUE #892: At startup, the NodeJS server should kill off any running 'jupyter-notebook' processes
If the NodeJS server gets killed, it will automatically be restarted. The new server instance will not know about any of the 'jupyter-notebook' processes that were created by the previous instance, and so will create new ones to use in their place.

However, those previously started processes (which can never again be used) will not be killed. The result is potentially a lot of processes that are taking CPU time without adding any value.

We should fix this by killing off any 'jupyter-notebook' processes at the time the NodeJS server starts.

This should be safe, as the NodeJS server only ever runs inside of a Docker container, so the only killed processes will be those inside of that container.

ojarjur:
This change makes the Datalab NodeJS server, at startup, list
all processes for the 'jupyter-notebook' binary, and kill them.

This fixes #892, where those processes could become abandoned
if the NodeJS server gets restarted.

This should be a safe thing to do at startup, as the NodeJS
server only runs inside of a Docker container, and so it has
an isolated process space.

I.E. the only processes that should be killed are those that
were started by an earlier incarnation of the NodeJS server.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -36,8 +36,6 @@ interface JupyterServer {
   notebooks: string;
   childProcess?: childProcess.ChildProcess;
   proxy?: httpProxy.ProxyServer;
-
-  portResolved: ()=>void;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


qimingj:
LGTM

PR #900, opened 2016-07-07
Fix erroneous duplicated quotes in the HTML templates.

ojarjur:
Both the notebook and tree-view HTML templates have
some attribute values specified with too many closing
quotes (e.g. `class="btn""`).

This does not seem to have affected the rendering of those
pages, but it was confusing the syntax-highlighting features
of some editors when displaying the HTML source.

This change fixes that by removing the duplicated closing quotes.


qimingj:
LGTM. Thanks!

PR #901, opened 2016-07-07
Split the `sessions` UI out into its own page.

ojarjur:
Previously, the `tree` UI included two unrelated views; a directory
view of the file system, and a list view of running sessions.

This was confusing as it was not intuitive that you should open
the file listing in order to be able to switch to the session
list, nor was it obvious that the UI controls on that page
toggled the type of view that was presented.

This change tries to clean all of that up by splitting the
session listing into its own page, and putting a link to that
page in the button toolbar that is in the top, right-hand
corner of the UI.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -91,6 +91,7 @@ var callbackManager: callbacks.CallbackManager = new callbacks.CallbackManager()
  */
 var templates: common.Map<string> = {
   'tree': fs.readFileSync(path.join(__dirname, 'templates', 'tree.html'), { encoding: 'utf8' }),
+  'sessions': fs.readFileSync(path.join(__dirname, 'templates', 'sessions.html'), { encoding: 'utf8' }),
   'edit': fs.readFileSync(path.join(__dirname, 'templates', 'edit.html'), { encoding: 'utf8' }),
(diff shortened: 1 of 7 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -342,9 +343,29 @@ export function handleRequest(request: http.ServerRequest, response: http.Server
     response.end();
     return;
   }
+
+  var path = url.parse(request.url).pathname;
+  if (path.indexOf('/sessions') == 0) {
```
qimingj: Can we move '/tree' and '/edit' here too?

qimingj:
LGTM! A minor comment that can be addressed separately.

sources/web/datalab/jupyter.ts
ojarjur: > Can we move '/tree' and '/edit' here too?

Not without consequences.

Specifically, we rely on the Jupyter server validating the path passed to each of the '/tree', '/edit', and '/nb' handlers. As such, if we changed those handlers to be statically served by the Node JS server, then instead of a 404 message, the client would get the regular HTML page with API calls that will fail.

qimingj: ah, I see. Thanks for the explanation.

PR #904, opened 2016-07-08
Improve error checking and logging.

ojarjur:
This change should improve reliability and debugability in multiple ways:

1. Adding some important logging, including logging requests at
   the time they are received rather than just at the time we
   finish processing them. This enables us to distinguish
   between a request that never made it to the server vs. one
   that the server received but never finished processing.
2. Adding error handling (via logging) for exceptions thrown
   by the main request handler.
3. Adding error handling (logging and invoking per-user callbacks)
   for exceptions thrown during the setup process for the per-user
   Jupyter servers.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -113,71 +113,83 @@ function pipeOutput(stream: NodeJS.ReadableStream, port: number, error: boolean)
   })
 }
 
+function createJupyterServerAtPort(port: number, userId: string, userDir: string) {
+  var server: JupyterServer = {
+    userId: userId,
+    port: port,
+    notebooks: userDir,
+  };
+
+  function exitHandler(code: number, signal: string): void {
+    logging.getLogger().error('Jupyter process %d for user %s exited due to signal: %s',
+                              server.childProcess.pid, userId, signal);
+    delete jupyterServers[server.userId];
+  }
+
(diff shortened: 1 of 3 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -113,71 +113,83 @@ function pipeOutput(stream: NodeJS.ReadableStream, port: number, error: boolean)
...
+}
+
 /**
  * Starts the Jupyter server, and then creates a proxy object enabling
  * routing HTTP and WebSocket requests to Jupyter.
  */
 function createJupyterServer(userId: string, remainingAttempts: number) {
+  logging.getLogger().info('Checking user dir for %s exists', userId);
+  var userDir = userManager.getUserDir(userId);
+  if (!fs.existsSync(userDir)) {
```
qimingj: Can you add the logic to create "/content" if not exist, perhaps in a separate CL? After looking at a few logs, I think this is probably what happened after VM restarted.

qimingj:
LGTM

sources/web/datalab/jupyter.ts
ojarjur: I'll do that in a separate pull request.

It seems that the version in master already does that inside of the Docker container before starting the NodeJS server.

I'll send out a PR to make the datalab-managed branch do the same.

PR #905, opened 2016-07-08
Improve error checking and logging.

ojarjur:
This change should improve reliability and debugability in multiple ways:

1. Adding some important logging, including logging requests at
   the time they are received rather than just at the time we
   finish processing them. This enables us to distinguish
   between a request that never made it to the server vs. one
   that the server received but never finished processing.
2. Adding error handling (via logging) for exceptions thrown
   by the main request handler.
3. Adding error handling (logging and invoking per-user callbacks)
   for exceptions thrown during the setup process for the per-user
   Jupyter servers.


qimingj:
LGTM

PR #908, opened 2016-07-12
Change the '/_info' handler to be usable as a JSON API.

ojarjur:
Eventually, this will enable us to add more system-status information to the sessions page in the UI, by having it call into the '/_info' handler to get that data.

This currently includes the list of running Jupyter servers, the total memory available to the system, and the amount of memory currently in use.

In later changes, we will add information about the amount of memory in use by each Jupyter server, and their CPU usage.

### DIFF
--- a/sources/web/datalab/info.ts
+++ b/sources/web/datalab/info.ts
@@ -24,45 +24,20 @@ import url = require('url');
  */
 var appSettings: common.Settings;
 
-function stringifyMap(map: {[index: string]: string}): string {
-  var textBuilder: string[] = [];
-
-  var names: string[] = [];
-  for (var n in map) {
-    names.push(n);
-  }
-  names = names.sort();
-
-  for (var i = 0; i < names.length; i++) {
-    textBuilder.push(names[i] + ': ' + map[names[i]]);
-  }
-
-  return textBuilder.join('\n');
-}
-
 /**
  * Implements information request handling.
  * @param request the incoming health request.
  * @param response the outgoing health response.
  */
 function requestHandler(request: http.ServerRequest, response: http.ServerResponse): void {
+  var info: any = {};
+  info['env'] = process.env;
+  info['requestHeaders'] = request.headers;
+  info['settings'] = appSettings;
+  info['servers'] = jupyter.getInfo();
+
   response.writeHead(200, { 'Content-Type': 'text/plain' });
-
-  response.write('Environment Variables:\n');
-  response.write(stringifyMap(process.env));
-  response.write('\n\n');
-
-  response.write('Application Settings:\n');
-  response.write(JSON.stringify(appSettings, null, 2));
-  response.write('\n\n');
-
-  response.write('Request Headers:\n');
-  response.write(stringifyMap(request.headers));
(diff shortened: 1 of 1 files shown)


nikhilk:
This was meant as human-readable page -- would it make sense to serve ux-specific data on a different endpoint?

Also, note that the user should be already logged in to get to this, so not sure which sensitive information is being removed, but potentially some of it is actually useful for the human-readable version.

ojarjur:
> This was meant as human-readable page -- would it make sense to serve ux-specific data on a different endpoint?

I actually find the JSON formatted version much more readable than the older free-form version (it's pretty-printed JSON), but I can see the point of having one endpoint that scrubs info and another one that doesn't.

ojarjur:
> Also, note that the user should be already logged in to get to this, so not sure which sensitive information is being removed, but potentially some of it is actually useful for the human-readable version.

I expanded the commit message for that change to make it more clear why the information is being scrubbed: this is so that users don't accidentally include their auth credentials in bug reports.

qimingj:
Is this for MVM or local Datalab? For "local" run, the kernel may be on a different VM, so perhaps it is more valuable to do it in Datalab lib?

ojarjur:
> For "local" run, the kernel may be on a different VM, so perhaps it is more valuable to do it in Datalab lib?

I don't think this is a one-or-the-other situation, but rather a case where we need to expose both bits of information.

In particular, if a user is seeing slowness and wants to debug it, then they need to be able to understand both how heavily loaded the local machine is *and* how heavily loaded any kernel gateways are.

qimingj:
The system information on notebook server is less useful than that of kernels, although they may run on same VM.

Assuming notebook server runs locally, there should be only one jupyter instance running in container. The /_info entry point is useful in managed VM Datalab, since it is for multiple users and all kernels run on the same VM. But it may not be so useful for local-run Datalab.

Maybe
- /_info should always return whether kernels run on the same VM or remotely. If remotely, where they are. 
- Create magics (%sys?) which shows diagnostics on the kernel system, as part of Datalab lib.

Thoughts?

ojarjur:
I agree that the information about each Jupyter server should include where it is running, but I want to delay that until a later PR.

As far as cell magics for system info, I'm not sure what that would buy us over just using `%bash` to run things like `ps` and `top`.

qimingj:
Perhaps same question can be asked regarding the /_info entry point. What does it do more than !ps and !top in a notebook?

Assuming it is for local run, if notebook server and kernels run on same VM, then the system information can be retrieved in a notebook by %%bash, !top, !ps, or !cat /proc/*, as you mentioned. If they don't run on the same VM, the system information on the notebook server may not be that interesting anyway compared to kernel system information.

There should always be only one Jupyter notebook server instance and one user in local-run Datalab. IMO, the most valuable information of /_info entry point would be jupyter config, request info, how many kernels are running and where they run. These are information that is only known by the web server. 

Again the question is whether it is for managed VM Datalab or local-run Datalab? It seems to me this will be useful for MVM Datalab which serves multi-users and hosts multiple jupyter instances, but not much for local-run Datalab.

PR #909, opened 2016-07-12
Update the reference to the datalab python libraries.

ojarjur:
The repository 'googledatalab/datalab' is being renamed as
'googledatalab/pydatalab' so that this repository can be
migrated to the 'googledatalab' organization.

This change is necessary for the building of the Docker image
to pick up the library from its new location.


qimingj:
LGTM

PR #910, opened 2016-07-14
Initial image for the kernel-gateway

ojarjur:
This change takes the existing, main Dockerfile for Datalab, and
splits it into three separate images:

1. A base image used by the other images. This includes the
   pydatalab library and it's assorted run-time dependencies.
2. The main image for running a stand-alone Datalab instance.
3. A new image for running a kernel gateway for Datalab.

Each of the three images also was given a corresponding `README.md`
file describing them.

Although the main image remains usable by itself, with this change
it will be possible to run the Datalab UI on a separate machine
from the one used for running the kernels.

Additional instructions for running Datalab in such a split
configuration will be coming in a later change.


containers/gateway/build.sh
```diff
@@ -0,0 +1,43 @@
...
+cd ../base
+docker build -t datalab-base .
+popd
+
+# Build the docker image
+docker build -t datalab-gateway .
+
+# Finally cleanup
+rm -rf build
+rm -rf notebooks
```
qimingj: Remove it as well.

qimingj:
Since we have a kernel gateway container which listens on 8082, can we also have a containers/datalab/run-with-kernel-gateway.sh, which includes something like:

export EXPERIMENTAL_KERNEL_GATEWAY_URL=https://127.0.0.1:8082
. run.sh

to make it easier for user?

qimingj:
LGTM! A few minor comments.

containers/datalab/build.sh
```diff
@@ -26,7 +26,12 @@ cat Dockerfile.in | sed $VERSION_SUBSTITUTION | sed $COMMIT_SUBSTITUTION > Docke
 
 # Copy build outputs as a dependency of the Dockerfile
 rsync -avp ../../build/ build
-rsync -avp ../../content/datalab/ notebooks
```
ojarjur: Done.

containers/gateway/build.sh
ojarjur: Done.

ojarjur:
> Since we have a kernel gateway container which listens on 8082, can we also have a containers/datalab/run-with-kernel-gateway.sh, which includes something like:
> 
> export EXPERIMENTAL_KERNEL_GATEWAY_URL=https://127.0.0.1:8082
> . run.sh
> 
> to make it easier for user?

That's coming in a later pull request. The actual setup is a bit more complicated than that (for instance, you have to connect to the non-routable IP address exposed by Docker, rather than 127.0.0.1), so I think it deserves a separate code review.

qimingj:
LGTM. Thanks!

PR #912, opened 2016-07-15
Add scripts for running Datalab in a split setup.

ojarjur:
This change adds to new scripts that make it easy to run the Datalab UI
locally while connecting it to a kernel gateway running inside a GCE VM.

Specifically, the two new scripts are:

1. containers/gateway/deploy.sh - for setting up the kernel gateway.
2. containers/gateway/run-with-gce.sh - for running the Datalab UI.

The main entry point for a user wanting to run their kernels in the
cloud will be the 'run-with-gce.sh' script, which will run the
'deploy.sh' script iff there isn't already a VM running a kernel gateway.

### DIFF
--- a/containers/gateway/deploy.sh
+++ b/containers/gateway/deploy.sh
@@ -1,0 +1,118 @@
+#!/bin/bash -e
+
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+USAGE="USAGE: ${0} <PROJECT> <ZONE>
+
+Alternatively, you can set the default project and zone using the gcloud tool:
+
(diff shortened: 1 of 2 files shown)


containers/gateway/deploy.sh
```diff
@@ -0,0 +1,115 @@
...
+fi
+
+echo "${DOCS}"
+
+echo "Will deploy a GCE VM to project ${PROJECT} in zone ${ZONE}"
+read -p "Proceed? [y/N] " PROCEED
+
+if [[ "${PROCEED}" != "y" ]]; then
+  echo "Deploy cancelled"
+  exit
```
qimingj: Should it return non-zero code so run-with-gce.sh can catch it?

containers/gateway/deploy.sh
```diff
@@ -0,0 +1,115 @@
...
+fi
+
+./build.sh || exit ${ERR_DOCKER_BUILD}
+IMAGE="gcr.io/${PROJECT}/datalab-gateway"
+docker tag -f datalab-gateway "${IMAGE}"
+gcloud docker push "${IMAGE}" || exit ${ERR_DOCKER_PUSH}
```
qimingj: It seems this script is used in development environment, where it relies on the building system. Maybe we should have another "consumer script" which relies on gcloud only? That should be a separate change though.

qimingj:
Very cool!  LGTM with a couple minor observations.

containers/gateway/deploy.sh
ojarjur: Done.

containers/gateway/deploy.sh
ojarjur: Agreed. I added a TODO for that.

PR #913, opened 2016-07-18
Fix a bug when specifying a project and zone via command line arguments.

ojarjur:
There was a bug in the initial checkin for the 'run-with-gce.sh' script
where, if specified on the command line, the selected project and zone
would not be passed to the nested call to the 'deploy.sh' script.

This was because the call to 'deploy.sh' always used the default values
configured with gcloud, so if those differed from the explicitly-passed
values, then the deploy script would try to use the wrong project
and/or zone.

This change fixes that bug by having the project and zone always be
explicitly passed to the call to the 'deploy.sh' script.


qimingj:
LGTM

PR #914, opened 2016-07-18
Usability improvements for the run-with-gce.sh script.

ojarjur:
This change fixes two notable usability issues with running Datalab
against a kernel gateway in a GCE VM. Specifically, with this change we will:

1. Create a separate VM for each user.
2. Enforce that the user's SSH keys are copied to that VM before trying
   to connect to it.

The first change fixes an issue where multiple users could have been
making changes to the same kernel if they were using the same kernel
gateway. This was due to the fact that the only information used to
identify a kernel in the kernel gateway was the path to and name of
the notebook. As such, if two users were running the same notebook,
they would have wound up touching the same kernel (and sharing
mutable state).

The second change fixes a first-time-use issue where the script would
not cause a user's SSH keys to be transferred to the gateway VM before
trying to connect to that VM using SSH. This meant that a user who had
connected to other VMs (using gcloud) would have been fine, but any
other user would have hit permission denied issues.

### DIFF
--- a/containers/datalab/run-with-gce.sh
+++ b/containers/datalab/run-with-gce.sh
@@ -32,7 +32,8 @@ to look it up using the 'ifconfig' command.
 "
 
 ERR_USAGE=1
-ERR_DEPLOY=2
+ERR_LOGIN=2
+ERR_DEPLOY=3
 
 DOCS="This script runs Datalab connected to a kernel gateway running in a GCE VM.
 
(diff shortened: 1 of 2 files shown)


containers/datalab/run-with-gce.sh
```diff
@@ -60,14 +61,39 @@ fi
...
+USER_EMAIL=`gcloud info --format="value(config.account)"`
+if [[ -z "${USER_EMAIL}" ]]; then
+  echo "You must log in via the gcloud tool"
+  echo "    gcloud auth login"
+  exit ${ERR_LOGIN}
+fi
+
+echo "Looking up gateway VM for ${USER_EMAIL}..."
+
+USER_HASH=`echo "${USER_EMAIL}" | openssl sha1 -r | cut -d ' ' -f 1`
```
qimingj: It seems openssl is not built-in in Windows? If so, can we use pure string replacement to sanitize the email to remove the dependency? Or maybe you plan to create a different script for end users (not for Datalab developers)?

ojarjur: > can we use pure string replacement to sanitize the email to remove the dependency?

Not in any sort of reliable way. Email addresses are far too complicated to be able to sanitize them in a simple way without breaking for valid addresses (for example, they can include quoted strings).

That being said, we can check if the email address can be simply sanitized, and only hash it if that fails. I've changed the code to do exactly that.

PTAL

> Or maybe you plan to create a different script for end users (not for Datalab developers)?

The script for end users will be based on this one, but it will be for Linux/Mac only. We'll have to create a separate batch script for windows users.

containers/datalab/run-with-gce.sh
```diff
@@ -83,8 +80,33 @@ fi
...
+# '[-a-zA-Z0-9]+'. Since this mapping is 1:1, we ensure that the uniqueness
+# of email addresses is maintained. There is additionally a length limit
+# of 63 characters for instance names, so we do not use the escaped email
+# address if it is longer than 32 characters.
+EMAIL_WITH_HYPHENS_ESCAPED=`echo "${USER_EMAIL}" | sed -e 's/-/-h-/g'`
+EMAIL_WITH_AT_ESCAPED=`echo "${EMAIL_WITH_HYPHENS_ESCAPED}" | sed -e 's/@/-at-/g'`
+EMAIL_WITH_DOT_ESCAPED=`echo "${EMAIL_WITH_AT_ESCAPED}" | sed -e 's/\./-dot-/g'`
+ESCAPED_USER_EMAIL=`echo "${EMAIL_WITH_DOT_ESCAPED}" | sed -e 's/[^-a-zA-Z0-9]//g'`
+TRUNCATED_USER_EMAIL="${ESCAPED_USER_EMAIL:0:32}"
+if [[ -z "${ESCAPED_USER_EMAIL}" || "${ESCAPED_USER_EMAIL}" != "${TRUNCATED_USER_EMAIL}" ]]; then
```
qimingj: Since we only use TRUNCATED_USER_EMAIL when it is equal to ESCAPED_USER_EMAIL, maybe we can just have: 
if [[ -z "${ESCAPED_USER_EMAIL}" || {#ESCAPED_USER_EMAIL}  > 32 ]]; then
and no need to truncate the email?

containers/datalab/run-with-gce.sh
qimingj: Sounds good.

qimingj:
LGTM. One minor observation.

containers/datalab/run-with-gce.sh
ojarjur: > Since we only use TRUNCATED_USER_EMAIL when it is equal to ESCAPED_USER_EMAIL, maybe we can just have: 
> if [[ -z "${ESCAPED_USER_EMAIL}" || {#ESCAPED_USER_EMAIL} > 32 ]]; then
and no need to truncate the email?

Nice! I didn't know you could do that in bash, so thanks for the suggestion.

PR #917, opened 2016-07-20
Add support for writing logs to GCP logging

ojarjur:
This change makes the 'containers/gateway/deploy.sh' script set up the
gateway VM in such a way that the Docker logs for the kernel gateway
container get imported into the GCP logging service.


qimingj:
LGTM. Great!

PR #919, opened 2016-07-22
Fixed README link in Help panel, and quotes in build.sh.

0emily0:


sources/build.sh
```diff
@@ -16,7 +16,7 @@
 # Builds all components.
 
 if [ -z "$REPO_DIR" ];
-  then echo "REPO_DIR is not set. Please run `source tools/initenv.sh` first";
+  then echo "REPO_DIR is not set. Please run source ../tools/initenv.sh first";
```
qimingj: Can we add the quotes, perhaps escaping them by \"?

sources/web/datalab/templates/nb.html
```diff
@@ -222,7 +222,7 @@
                 <br />
                 <p><b>Docs and Samples</b></p>
                 <p>
-                  The <a href="/notebooks/datalab/Readme.ipynb" target"_blank">Cloud Datalab Guide</a> featuring
+                  The <a href="notebooks/datalab/docs/notebooks/Readme.ipynb" target"_blank">Cloud Datalab Guide</a> featuring
```
qimingj: Thanks for fixing it! Can you also fix target="_blank"? It is missing a "=".

qimingj: Also, I think the leading slash "/" should be kept?

sources/build.sh
0emily0: DONE

sources/web/datalab/templates/nb.html
0emily0: DONE

0emily0:
PTAL

qimingj:
LGTM. Thanks!

PR #921, opened 2016-07-23
Add a new Docker image for running with a backend on GCP.

ojarjur:
This encapsulates all of the logic for spinning up a GCE
VM running a Datalab kernel gateway, setting up an SSH
tunnel connected to that VM, and then running a local
Datalab instance pointed at that SSH tunnel.

This should make it much simpler to run Datalab with kernels in the cloud.

### DIFF
--- a/containers/gcp/build.sh
+++ b/containers/gcp/build.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+# Copyright 2016 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Builds the docker image for running Datalab on the Google Cloud Platform.
+
+USAGE="USAGE: ${0} [-p] [-h]
+
+Where '-p' indicates that the 'datalab-gateway' Docker image should be
+built locally and pushed to the Google Container Registry for the
+project defined in the environment variable 'PROJECT_ID' (or, if that
+variable is not provided, the default project set in the gcloud config).
+"
+
+FORCE_PUSH=""
+while getopts ":ph" OPT; do
+  case $OPT in
+    p)
+      FORCE_PUSH="true"
+      ;;
+    h)
+      echo "${USAGE}"
+      exit 0
+      ;;
+    \?)
+      echo "Unknown args: -${OPTARG}"
+      echo "${USAGE}"
+      exit 1
+      ;;
+  esac
+done
+
+if [ "${FORCE_PUSH}" == "true" ]; then
+  PROJECT_ID=${PROJECT_ID:-`gcloud config list 2> /dev/null | grep 'project = ' | cut -d ' ' -f 3`}
+  if [ -z "${PROJECT_ID}" ]; then
+    echo "You must specify a target project to which to push the gateway image"
+    exit 1
+  fi
+
+  GATEWAY_IMAGE="gcr.io/${PROJECT_ID}/datalab-gateway"
+  echo "Forcing a build and push of the image ${GATEWAY_IMAGE}"
+  cd ../gateway
+  ./build.sh
+  cd ../gcp
+
+  docker tag -f datalab-gateway "${GATEWAY_IMAGE}"
+  gcloud --project="${PROJECT_ID}" docker push "${GATEWAY_IMAGE}"
+fi
+
+# Build the base docker image
+cd ../datalab
+./build.sh
+cd ../gcp
+
+# Build the docker image
+docker build -t datalab-gcp .
--- a/containers/gcp/content/run-with-gce.sh
+++ b/containers/gcp/content/run-with-gce.sh
@@ -14,21 +14,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-USAGE="USAGE: ${0} [<PROJECT> <ZONE> [<DOCKER_BRIDGE_IP>]]
+USAGE="USAGE: ${0}
 
-Where <PROJECT> is the ID of the Google Cloud Platform project that will host
-the kernel gateway, <ZONE> is the Google Compute Engine zone where the kernel
-gateway will run, and <DOCKER_BRIDGE_IP> is the IP address of the 'docker0'
-network bridge.
+You must also specify a project ID and zone. The project ID is the ID of
+the Google Cloud Platform project that will host the kernel gateway and 
+the zone is the Google Compute Engine zone where the kernel gateway will run.
 
-The <PROJECT> and <ZONE> arguments can be omitted if you set the default
-project and zone using the gcloud tool:
+These may be specified by either setting the PROJECT_ID and ZONE environment
+variables, or by setting the default project and zone using the gcloud tool:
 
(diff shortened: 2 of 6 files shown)


containers/gcp/content/run-with-gce.sh
```diff
@@ -107,45 +106,21 @@ else
   INSTANCE_PREFIX="datalab-${ESCAPED_USER_EMAIL}"
 fi
 
-INSTANCE=`gcloud compute instances list --project "${PROJECT}" --zone "${ZONE}" --regex "${INSTANCE_PREFIX}-[0-9]*" --limit 1 --format "value(name)"`
+INSTANCE=`gcloud compute instances list --project "${PROJECT_ID}" --zone "${ZONE}" --regex "${INSTANCE_PREFIX}-[0-9]*" --limit 1 --format "value(name)"`
```
qimingj: I tried to find out how auth works here but I couldn't...On container startup it is not authenticated, right? How does "gcloud compute instances" work? Maybe you can add some comments here to explain it?

ojarjur: The new image has to be run in a container that has a volume mapping the host's gcloud config directory to the container's gcloud config directory (so it uses the hosts auth configuration).

You're absolutely right that I need to document that. I also forgot to add a README to this directory. I'll add that file and put the details in there.

ojarjur: I've added a README file with docs outlining how the auth works.

PTAL

qimingj: That explains it. :) Thanks.

containers/gcp/run.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh
+# Copyright 2015 Google Inc. All rights reserved.
```
qimingj: 2016

containers/gcp/run.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# Runs the docker container locally. Auth will happen in the browser when
+# the user first opens Datalab after accepting the EULA.
+
+# Passing in 'shell' flag causes the docker container to break into a
+# command prompt, which is useful for tinkering within the container before
+# manually starting the server.
+
+CONTENT=$HOME
+GCLOUD_CONFIG=`gcloud info --quiet --format 'value(config.paths.global_config_dir)'`
```
qimingj: Perhaps we need a sanity check on gcloud existence and minimum version?

containers/gcp/run.sh
```diff
@@ -0,0 +1,76 @@
...
+ENTRYPOINT="/datalab/run-with-gce.sh"
+if [ "$1" != "" ]; then
+  if [ "$1" != "shell" ]; then
+    CONTENT=$1
+    shift
+  fi
```
qimingj: can it be an "else"?

containers/gcp/run.sh
ojarjur: This isn't new; it's a copy of the ../datalab/run.sh file.

containers/gcp/run.sh
```diff
@@ -0,0 +1,76 @@
...
+    ENTRYPOINT="/bin/bash"
+  fi
+fi
+
+# Ensure that the target project has a copy of the gateway image.
+# We do this rather than pointing the VM directly at a public
+# image so that the VM will not accidentally pick up a new version
+# when rebooted.
+PROJECT_ID=${PROJECT_ID:-`gcloud config list 2> /dev/null | grep 'project = ' | cut -d ' ' -f 3`}
+GATEWAY_IMAGE="gcr.io/${PROJECT_ID}/datalab-gateway"
```
qimingj: Also add ${USER} to the image so we don't stomp each other's image if we happen to use same project?

qimingj: It seems we cannot since the deploy.sh does not know $(USER). I guess it is fine.

containers/gcp/run.sh
ojarjur: It's not an else, both branches can be run.

So, for instance, you can run `./run.sh ~/my-workspace shell` to both set the content directory to `~/my-workspace` and override the entrypoint to be a simple shell.

I haven't yet needed this functionality for debugging images, but I think we should keep the debugging instructions for `containers/datalab` and `containers/gcp` in sync, so we should either support the current options in both, or remove them in both.

containers/gcp/run.sh
ojarjur: That shouldn't happen, as this script will only auto-deploy the image if it doesn't exist yet.

I'll add a comment to that effect.

containers/gcp/run.sh
ojarjur: Added a bunch of sanity checks.

I don't think we want to specify a minimum version, as that sort of check is too liable to falsely reject setups that will actually work. Instead, I've suggested that the user upgrade to the latest
version of gcloud prior to running (as part of the usage string).

containers/gcp/run.sh
qimingj: I see. I didn't realize the shift command is for that. Thanks for explaining it.

containers/gcp/run.sh
```diff
@@ -0,0 +1,128 @@
...
+# and
+# 2. The project can switch to a version of the image with custom extensions.
+GATEWAY_IMAGE="gcr.io/${PROJECT_ID}/datalab-gateway"
+
+# Only update the gateway image if it has not been pushed to the project yet.
+# We want to enforce that the image is there, but do not want to
+# override a previously pushed image in case the project has a
+# customized gateway image.
+gcloud docker pull "${GATEWAY_IMAGE}" || PUSH_IMAGE="true"
+if [ "${PUSH_IMAGE}" == "true" ]; then
```
qimingj: If this script is for development environment, I would expect it always pushes the locally built gateway image. Otherwise, after modifying anything in gateway image, it won't be picked up by running this script without pushing the gateway image explicitly.

Maybe always push a local image by default, and add a flag to optionally use existing image?

containers/gcp/run.sh
qimingj: Sounds good.

containers/gcp/run.sh
ojarjur: I understand the concern about someone doing `./build.sh && ./run/.sh` and not seeing their changes.

However, I'm not comfortable with the run script potentially stomping on the gateway image if a developer isn't trying to modify that.

In particular, one of the common scenarios should be someone developing a custom extension to the gateway image, and testing that. In that scenario, force pushing a new image built from the base gateway Dockerfile will get in the user's way instead of helping them.

As an alternative, I'll look into maybe making the `build.sh` script include an option to build and push the gateway image.

qimingj: We can add a release.sh (or stage.sh, which append $(USER) to the image name) to release the gateway image.

The run.sh here can take an optional param indicating the gateway image name to use. If not, then pull the default one.

qimingj:
LGTM. We can modify the script to handle the gateway image pushing in a separate change.

containers/gcp/run.sh
ojarjur: I've added a '-p' flag to the build.sh script.

containers/gcp/build.sh
```diff
@@ -0,0 +1,66 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Builds the docker image for running Datalab on the Google Cloud Platform.
+
+USAGE="USAGE: ${0} [-p] [-h]
+
+Where '-p' indicates that the 'datalab-gateway' Docker image should be
+built locally and pushed to the Google Container Registry for the
+project defined in the environment variable 'PROJECT_ID'
```
qimingj: ... or the gcloud active project?

qimingj:
LGTM!

containers/gcp/build.sh
ojarjur: Done.

PR #923, opened 2016-07-26
Simplify the run step for Datalab with a kernel gateway in GCP.

ojarjur:
This change is a response to feedback about the previous steps
being too complicated and/or brittle. The aim of this is to
make the end-user instructions for Datalab-on-GCP as simple as
possible.

Specifically, the following improvements are included:

1. Remove all dependencies on 'gcloud' from the setup and run.
   This has the consequence of the user potentially having to
   sign-in to gcloud a second time if they have already signed
   in to it from the host, but that sign-in is persisted, so
   they will only have to do it one more time.
2. Make the PROJECT_ID and ZONE variables optional rather than
   required. If they are not provided, then the user will be
   prompted to enter them at the command line. Either way, the
   values selected will be persisted so they can be reused by
   later runs of the tool.
3. Add the option of explicitly specifying the name of the VM
   instance when running the tool. This gives more flexibility
   to users who want to manually create the VM themselves, while
   still having Datalab managed the SSH tunnel for them.

With these changes, the new instruction to run Datalab (after
installing Docker) will simply be:

    docker run -it -p "8081:8080" \
      -v "${HOME}:/content" \
      gcr.io/cloud-datalab/<NAME_OF_THE_DOCKER_IMAGE>

### DIFF
--- a/containers/gcp/content/deploy.sh
+++ b/containers/gcp/content/deploy.sh
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ERR_DOCKER_PUSH=1
+ERR_CANCELLED=1
 ERR_NETWORK_CREATE=2
 ERR_FIREWALL_RULE=3
 ERR_INSTANCE_CREATE=4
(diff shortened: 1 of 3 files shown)


qimingj:
LGTM. That's much better than mapping in gcloud config!

PR #930, opened 2016-08-02
Remove the setting of the HOME environment variable during setup.

ojarjur:
This was causing issues when running Datalab on Windows, as strange
issues seem to occur if you have your HOME variable set to an
externally mapped volume on Windows.


qimingj:
LGTM. Nice investigation!

PR #932, opened 2016-08-03
Fix a bug in the project lookup.

ojarjur:
This fixes a bug where the code did not properly handle the scenario
of the project being missing.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -387,7 +387,7 @@ function getBaseTemplateData(request: http.ServerRequest): common.Map<string> {
   if (process.env.DATALAB_ENV == 'local') {
     templateData['isSignedIn'] = auth.isSignedIn().toString();
   }
-  if (auth.isSignedIn() && process.env.PROJECT_NUMBER != '') {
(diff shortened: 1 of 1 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -387,7 +387,7 @@ function getBaseTemplateData(request: http.ServerRequest): common.Map<string> {
   if (process.env.DATALAB_ENV == 'local') {
     templateData['isSignedIn'] = auth.isSignedIn().toString();
   }
-  if (auth.isSignedIn() && process.env.PROJECT_NUMBER != '') {
+  if (auth.isSignedIn() && ('PROJECT_NUMBER' in process.env) && process.env.PROJECT_NUMBER != '') {
```
qimingj: Can we just do if (process.env.PROJECT_NUMBER)?

parthea:
Thanks, Omar! 

It worked with this change. I didn't set the project id during my testing so that explains why I hit a different code path!

sources/web/datalab/jupyter.ts

qimingj:
LGTM

PR #935, opened 2016-08-07
Update links to point to correct repository

parthea:
Update repository links in source code to point to https://github.com/googledatalab instead of https://github.com/GoogleCloudPlatform


ojarjur:
LGTM

PR #944, opened 2016-08-12
Simplify the entrypoint(s) for the Datalab container

ojarjur:
This change combines the 'run-with-gce.sh' and 'run-debug.sh' into the
single 'run.sh' script, which is now the default entrypoint to the
container.


qimingj:
LGTM

PR #946, opened 2016-08-15
Remove the need to login from the command line

ojarjur:
This change allows the user to perform the OAuth flow to log in
to Datalab at startup via the web browser.

That, in turn, streamlines the user experience for running Datalab
with a backend in GCP. This is because the user may not be logged
in initially, but we need them to be logged in before we can create
an SSH tunnel connected to their VM.

Previously, we satisfied that requirement by using the 'gcloud auth
login' flow from the command line, but that is cumbersome. With this
change that flow is now performed using the browser, which makes
it less cumbersome.

### DIFF
--- a/sources/web/datalab/login.ts
+++ b/sources/web/datalab/login.ts
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2016 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/**
+ * This file defines a special-purpose server used solely for performing the OAuth login
+ * flow prior to starting the main Datalab server.
+ *
+ * This is used for scenarios when the user must already be logged in before we can start
+ * Datalab (e.g. because we need to create an SSH tunnel to a GCE VM running a kernel gateway).
+ */
+
+/// <reference path="common.d.ts" />
+
+import auth = require('./auth');
+import http = require('http');
+import logging = require('./logging');
+import settings = require('./settings');
+import url = require('url');
+
+var appSettings: common.Settings;
+var server: http.Server;
+
+var successPage: string = `<html>
+  <head>
+    <meta http-equiv="refresh" content="30;url=/" />
+  </head>
+  <body>
+    <h1>Login successful.</h1>
+    <p>Startup should complete soon. Redirecting in 30 seconds...</p>
+  </body>
+</html>`;
+
+/**
+ * Handle incoming HTTP requests.
+ *
+ * The bulk of the work is done by the imported auth package. The only exceptions
+ * are redirecting to the start of the auth flow at the beginning, and serving the
+ * success page at the end.
+ */
+function requestHandler(request: http.ServerRequest, response: http.ServerResponse) {
+  var parsed_url = url.parse(request.url, true);
+  var path = parsed_url.pathname;
+  if (path.indexOf('/signin') == 0 || path.indexOf('/oauthcallback') == 0) {
+    auth.handleAuthFlow(request, response, parsed_url, appSettings);
+    return;
+  } else if (path.indexOf('/exit') == 0 && auth.isSignedIn()) {
+    response.statusCode = 200;
+    response.write(successPage);
+    response.end();
+    process.exit();
+    return;
+  } else {
+    var referer = '/signin?referer=%2Fexit';
+    response.statusCode = 302;
+    response.setHeader('Location', referer);
+    response.end();
+    return;
+  }
+}
+
+/**
+ * Load the configuration settings, and then start the server, which
+ * runs until the user signs in.
+ */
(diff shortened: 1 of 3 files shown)


parthea:
I ran into a reproducible issue where a user has to run the script `./run.sh` in `$REPO_DIR/containers/datalab/` twice. 

Steps to reproduce:

1. Run `sudo rm -rf datalab/` in the home folder to make sure that credentials are not cached.
2. Set `GATEWAY_VM='project/zone/gateway-vm-name'` in `$REPO_DIR/containers/datalab/run.sh`
3. Run script `./run.sh` in `$REPO_DIR/containers/datalab/`
4. I see `'Please visit http://localhost:8081 to log in'` in the console. So I click on it, and login
5. I see the following in the console output and script terminates. Note the message: `'Project <my_project_id> not found'`
```
./run.sh: 41: [: unexpected operator
Please visit http://localhost:8081 to log in
Project XXXXXX not found
USAGE: One of...

To run locally:

    docker run -it -p "8081:8080" -v "${HOME}:/content" gcr.io/cloud-datalab/datalab

Or, to connect to a kernel gateway in a GCE VM;

    docker run -it -p "8081:8080" -v "${HOME}:/content" \
      -e "GATEWAY_VM=${PROJECT_ID}/${ZONE}/${INSTANCE}"
      gcr.io/cloud-datalab/datalab

```
After I logged in, the http://localhost:8081/exit# page said `'Login successful. Startup should complete soon. Redirecting in 30 seconds...'` , and it redirects to http://localhost:8081/ which displays `'This site can’t be reached`' (since the script terminated)

If I run script `./run.sh` in `$REPO_DIR/containers/datalab/` again, everything works as expected.

sources/web/datalab/login.ts
```diff
@@ -0,0 +1,110 @@
...
+ * Handle incoming HTTP requests.
+ *
+ * The bulk of the work is done by the imported auth package. The only exceptions
+ * are redirecting to the start of the auth flow at the beginning, and serving the
+ * success page at the end.
+ */
+function requestHandler(request: http.ServerRequest, response: http.ServerResponse) {
+  var parsed_url = url.parse(request.url, true);
+  var path = parsed_url.pathname;
+  if (path.indexOf('/signin') == 0 || path.indexOf('/oauthcallback') == 0) {
```
qimingj: Can we simplify this a little bit such as:
if (auth.isSignedIn()) {
  response.statusCode = 200;
  response.write(successPage);
  response.end();
  process.exit();
} 
else if (path.indexOf('/signin') == 0 || path.indexOf('/oauthcallback') == 0) {
  ...(same)
}
else {
  ...(same)
}
It seems we can remove the /exit block?

sources/web/datalab/login.ts
```diff
@@ -0,0 +1,110 @@
...
+if (appSettings != null) {
+  appSettings.consoleLogging = false;
+  logging.initializeLoggers(appSettings);
+
+  server = http.createServer(requestHandler);
+  server.listen(appSettings.serverPort);
+
+  console.log('Please visit http://localhost:8081 to log in');
+}
+
```
qimingj: Add a console log here loadSettings failed? We should never be here though but in case it fails we'll know how to debug immediately.

qimingj:
LGTM with 2 minor observations. Also let's investigate the issue parthea found.

qimingj:
LGTM

sources/web/datalab/login.ts
ojarjur: We can't remove the '/exit' block (since the oauthcallback code needs something to redirect to), but the redirect to it here was redundant (since the oauthcallback handler already invokes it).

sources/web/datalab/login.ts
ojarjur: Done.

sources/web/datalab/login.ts
qimingj: I think it is fine if we have the following as the first condition check:

if (auth.isSignedIn()) {
response.statusCode = 200;
response.write(successPage);
response.end();
process.exit();
} 

It will visit /exit with state signed-in and will run the code in this block.

ojarjur: > I think it is fine if we have the following as the first condition check:

That was the code I originally wrote, but then decided was not sufficiently reliable.

Specifically, if any requests come to the server that are not from the user's browser (e.g. from an errant process running on their machine) then the response to the first request after successfully signing in might not actually get sent back to the user.

Checking for the expected path gives us a limited form of protection against that, with no additional code blocks (i.e. just a longer test expression).

qimingj: I see. Makes sense. Thanks for explaining this!

ojarjur:
> I ran into a reproducible issue where a user has to run the script ./run.sh in $REPO_DIR/containers/datalab/ twice.

I wasn't able to reproduce the error, but looking through the code, I can only find one potential source of this.

Since the failure is transient, that tells me that the project does exist, so the most likely culprit is some sort of race condition where the login server terminates before the user is fully authenticated.

Most of the logic for writing the user credentials  (which is [here](https://github.com/googledatalab/datalab/blob/master/sources/web/datalab/auth.ts#L127)) is synchronous and has to successfully complete before the redirect to the success page will be served. However, there is one exception to that.

Specifically, the `saveUserCredFile` method calls the `setGcloudAccount` method, which in turn executes `gcloud config set account`. However, that exec is asynchronous, so it might not have completed by the time the server shuts down.

If that happens, then it would not be surprising for subsequent gcloud commands to fail with a 'not found' error.

I'll change that call from `exec` to `execSync` to make sure that it always finishes before the caller believes that the login process is completed.

qimingj:
LGTM

parthea:
I can confirm that the issue (`Project XXXX not found`) is resolved now with the latest master which uses `execSync` instead of `exec`.

PR #948, opened 2016-08-17
Enable the notebook notary feature.

ojarjur:
This change turns on Jupyter's notebook notary feature using
a secret read from a file on the user's machine (which we
automatically generate if it is missing).

This has the consequence that certain types of charts saved in
cell outputs will not be displayed until the user (re)runs the
corresponding cells.


qimingj:
LGTM

PR #953, opened 2016-08-23
Fix ordering issues in startup environment settings

ojarjur:
The `run.sh` script, at startup, calls the `setup-env.sh` script
to setup basic environment variables.

However, the logic in those two scripts had evolved over time and
some environment variables were being reference before they were
set, causing the kernel-gateway feature to silently fail (all
operations would run on the local machine).

This change fixes that by changing the `setup-env.sh` script to
only set environment variables that can be determined at the
beginning of the container's existence without any other setup
having been performed.

### DIFF
--- a/containers/datalab/content/run.sh
+++ b/containers/datalab/content/run.sh
@@ -18,13 +18,13 @@ USAGE='USAGE: One of...
 
 To run locally:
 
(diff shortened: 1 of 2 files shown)


qimingj:
LGTM

PR #954, opened 2016-08-26
Initial support for custom themes

ojarjur:
This adds a new user setting called 'theme' that can be set to
one of 'light', 'dark', or 'custom'.

Both 'light' and 'dark' are canned themes with 'light' matching the
existing Datalab style, and 'dark' modifying that theme to have
dark backgrounds rather than light ones.

The 'custom' setting is for using a user-specified CSS file stored
in the 'datalab/.config' directory named 'custom.css'.

This change does not incorporate a UI for modifying the theme
setting, but the user can change it by directly modifying the
'datalab/.config/settings.json' file.

The canned dark theme still needs some work.

### DIFF
--- a/sources/web/datalab/static.ts
+++ b/sources/web/datalab/static.ts
@@ -20,7 +20,9 @@ import fs = require('fs');
 import http = require('http');
 import logging = require('./logging');
 import path = require('path');
+import settings = require('./settings');
 import url = require('url');
+import userManager = require('./userManager');
 
 var JUPYTER_DIR = '/usr/local/lib/python2.7/dist-packages/notebook';
 var CONTENT_TYPES: common.Map<string> = {
@@ -31,10 +33,13 @@ var CONTENT_TYPES: common.Map<string> = {
   '.txt': 'text/plain',
   '.html': 'text/html'
 };
+var CUSTOM_THEME_FILE = 'custom.css';
+var DEFAULT_THEME_FILE = 'light.css';
 
 var contentCache: common.Map<Buffer> = {};
+var watchedDynamicContent: common.Map<boolean> = {};
 
-function getContent(filePath: string, cb: common.Callback<Buffer>): void {
+function getContent(filePath: string, cb: common.Callback<Buffer>, isDynamic: boolean = false): void {
   var content = contentCache[filePath];
   if (content != null) {
     process.nextTick(function() {
@@ -47,6 +52,16 @@ function getContent(filePath: string, cb: common.Callback<Buffer>): void {
         cb(error, null);
       }
       else {
+        if (isDynamic && !watchedDynamicContent[filePath]) {
+          fs.watch(filePath, function(eventType, filename) {
+            logging.getLogger().info('Clearing cache for updated file: %s', filePath);
+            contentCache[filePath] = null;
+            if (eventType == 'rename') {
+              watchedDynamicContent[filePath] = false;
+            }
+          });
+          watchedDynamicContent[filePath] = true;
+        }
         contentCache[filePath] = content;
         cb(null, content);
       }
@@ -58,8 +73,11 @@ function getContent(filePath: string, cb: common.Callback<Buffer>): void {
  * Sends a static file as the response.
  * @param filePath the full path of the static file to send.
  * @param response the out-going response associated with the current HTTP request.
+ * @param alternatePath the path to a static Datalab file to send if the given file is missing.
+ * @param isDynamic indication of whether or not the file contents might change.
  */
-function sendFile(filePath: string, response: http.ServerResponse) {
+function sendFile(filePath: string, response: http.ServerResponse,
+                  alternatePath: string = "", isDynamic: boolean = false) {
   var extension = path.extname(filePath);
   var contentType = CONTENT_TYPES[extension.toLowerCase()] || 'application/octet-stream';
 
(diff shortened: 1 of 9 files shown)


sources/web/datalab/static/dark.css
```diff
@@ -0,0 +1,141 @@
+/*
+ * Copyright 2015 Google Inc. All rights reserved.
```
ojarjur: Note: Both this and the 'light.css' file are derived from the 'datalab.css' file, so they are expected to maintain the same copyright header.

qimingj:
Do we plan to make the UI change so users can select a theme there?

ojarjur:
> Do we plan to make the UI change so users can select a theme there?

I'd rather not do that as part of this change. I think a settings UI is potentially a large enough change that it should be done as its own separate review.

qimingj:
I see. It is quite some work. Good to know it will come too.

sources/web/datalab/static.ts
```diff
@@ -112,6 +116,23 @@ function datalabFileExists(filePath: string) {
...
 
 /**
+ * Sends a static 'custom.css' file located within the user's config directory.
+ *
+ * @param userId the ID of the current user.
+ * @param response the out-going response associated with the current HTTP request.
+ */
+function sendUserCustomTheme(userId: string, response: http.ServerResponse): void {
+    var customThemePath = path.join(settings.getUserConfigDir(userId), CUSTOM_THEME_FILE);
+    fs.stat(customThemePath, function(e, stats) {
```
qimingj: We can probably cache the state (whether custom theme file available) so we don't need to read from disk for each request. In the future if they change theme from UI, we can trigger the state refreshing (assume the UI change will come soon).

ojarjur: Actually, I'm now wondering if we should have an entirely separate cache for this type of dynamic file.

Right now, the only way to force a cache flush is to restart Datalab, and if a user is trying to tweak their custom theme, then that results in a very slow and cumbersome iterative development process.

What we could do instead is store both the file contents and it's metadata in the cache, then always run 'fs.stat' before looking in the cache. If we see the same timestamp then return the cached contents otherwise re-read the file, cache the updated contents, and then return them.

That would mean we always run fs.stat for reading these dynamic files, but would give the user an easier feedback loop where after saving their changes they just have to reload the webpage.

What do you think?

qimingj: In this particular case, once we have the UI ready we can do a cache refreshing on an UI event (versus now, we cannot track when user modifies the config file). So we get both userbility and performance benefits.

For general handling of those types of files, it depends on how dynamic they are and if there is a way to track user actions so we know when to refresh. A disk read seems costly at per request level, especially on disks without SSD, so I would try to avoid that whenever possible.

ojarjur: > In this particular case, once we have the UI ready we can do a cache refreshing on an UI event 

The UI will only be for the setting, not for editing the theme file. As such, refreshing the cache based on UI events will not be sufficient.

qimingj:
Considering the fact that there is only a single user for each Datalab instance, we are probably fine from performance perspective.

So LGTM.

ojarjur:
I've updated the file serving to perform cache invalidation when the cached file changes.

PTAL.

sources/web/datalab/static.ts
```diff
@@ -59,22 +71,26 @@ function getContent(filePath: string, cb: common.Callback<Buffer>): void {
  * @param filePath the full path of the static file to send.
  * @param response the out-going response associated with the current HTTP request.
  */
-function sendFile(filePath: string, response: http.ServerResponse) {
+function sendFile(filePath: string, response: http.ServerResponse, alternatePath: string = "", isDynamic: boolean = false) {
```
qimingj: Line too long (> 100).

sources/web/datalab/static.ts
```diff
@@ -63,22 +71,26 @@ function getContent(filePath: string, cb: common.Callback<Buffer>): void {
  * @param filePath the full path of the static file to send.
  * @param response the out-going response associated with the current HTTP request.
```
qimingj: Update param description?

qimingj:
LGTM. A couple of minor observations.

sources/web/datalab/static.ts
ojarjur: Done.

sources/web/datalab/static.ts
ojarjur: Done.

parthea:
@ojarjur  I couldn't locate the settings file `'datalab/.config/settings.json'`. I notice the following lines appear in my console: 

```
{"name":"app","hostname":"5f41dfc2dd4d","pid":63,"type":"jupyter","level":30,"msg":"[9000]: Using contents: services/contents\n","time":"2016-08-27T14:17:59.108Z","v":0}
Settings file /content/datalab/.config/settings.json not found.
Your active configuration is: [default]
```
Note the line: `'Settings file /content/datalab/.config/settings.json not found.'`

I tried deleting the `datalab` folder under `/content` so that a new configuration could be created. After the new configuration folder was created, I couldn't locate `'settings.json'`.

I found a file at `'/datalab/web/config/settings.json'`. Should I try to modify that file?

ojarjur:
> I couldn't locate the settings file 'datalab/.config/settings.json'

Sorry that I wasn't clearer in the description; that file does not exist by default, but if the user manually writes to it, then we will use that.

Alternatively, there is also a settings API which will create that file if it is missing.

That can be used by going to "/_setting?key=<key>&value=<value>', but I'm not expecting anyone to manually call that API. Instead, I hope to build a new UI page for listing and modifying the settings.

> I found a file at '/datalab/web/config/settings.json'

Those are just for build-time settings and shouldn't be changed (any changes to them will be lost when the Docker container is restarted).

parthea:
@ojarjur Thanks for explaining this! It's working! I created shortcuts to switch between themes. The dark theme is easier on the eyes at night.

[Dark theme](http://127.0.0.1:8081/_setting?key=theme&value=dark)
[Light theme](http://127.0.0.1:8081/_setting?key=theme&value=light)

techtonik:
Any screens?

parthea:
@techtonik 

Here is a screenshot of the dark theme:
screenshot from 2016-08-31 22-08-33

PR #957, opened 2016-08-30
Create the SSH tunnel as the logged in user rather than root.

ojarjur:
By default, everything inside of the Docker container runs as the user
'root'. Previously, we were setting up the SSH tunnel, we defaulted to
the user inside the VM matching the user inside the Docker
container. However, 'root' is special, and some VM images do not allow
SSH-ing as that user.

For example, this prevented the Datalab image from being able to connect
to a kernel gateway running inside of a Google Dataproc cluster.

This change fixes that be making the username inside of the VM match
the username of the logged-in account (e.g. "<username>@gmail.com").

### DIFF
--- a/containers/datalab/content/run.sh
+++ b/containers/datalab/content/run.sh
@@ -109,14 +109,15 @@ if [[ -n "${INSTANCE}" ]]; then
     fi
   fi
 
-  echo "Will connect to the kernel gateway running on the GCE VM ${INSTANCE}"
+  SSH_USER=`echo ${USER_EMAIL} | cut -d '@' -f 1`
+  echo "Will connect to the kernel gateway running on the GCE VM ${INSTANCE} as ${SSH_USER}"
   gcloud compute ssh --quiet \
     --project "${PROJECT_ID}" \
     --zone "${ZONE}" \
     --ssh-flag="-fNL" \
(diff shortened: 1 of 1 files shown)


containers/datalab/content/run.sh
```diff
@@ -109,14 +109,15 @@ if [[ -n "${INSTANCE}" ]]; then
     fi
   fi
 
-  echo "Will connect to the kernel gateway running on the GCE VM ${INSTANCE}"
+  SSH_USER=`echo ${USER_EMAIL} | cut -d '@' -f 1`
```
qimingj: It seems USER_EMAIL could be empty if user logs in through login.ts? Do we need to set USER_EMAIL between line 67 and 68?

chmeyers: What is the behavior when the user is not logged in?

ojarjur: > It seems USER_EMAIL could be empty if user logs in through login.ts? Do we need to set USER_EMAIL between line 67 and 68?

@qimingj Good catch; fixed.

> What is the behavior when the user is not logged in?

@chmeyers The code cannot reach to this point without the user logging in. Even if it did, it would fail during the SSH tunnel setup, since it needs the user's credentials to set up the SSH tunnel.

qimingj:
LGTM!

PR #958, opened 2016-08-31
Fix a bug in looking up the project number.

ojarjur:
Pull request #941 fixed a bug where our attempt to lookup the project
number would fail if the user was not signed in. However, that is just
a special case of the more general issue of not being able to look
up the project number.

We have seen reports of at least one other instance of this issue,
which is the case that a user is signed in but is not a project owner.

In that case, the check that the user is signed in will succeed, but
the attempt to lookup the project number will fail, causing the
Datalab startup to terminate.

This change fixes that by modifying the check to be more general.

Instead of checking if we believe that the lookup will succeed, we
try the lookup and gracefully handle any errors that occurred.


chmeyers:
LGTM

PR #959, opened 2016-09-01
Support logging in via the command line at startup.

ojarjur:
Previous versions of Datalab would trigger a CLI based OAuth flow at
startup if they needed the user's credentials to set up a connection
with the backend in a GCE VM.

We recently replaced that with a web-based OAuth flow that mimiced the
sign-in button in the Datalab UI.

However, both that login flow at startup and the sign-in button in the
UI require that the UI be accessed via the hostname 'localhost'.

That is not sufficient for the case of someone wanting to run Datalab
inside of the Google Cloud Shell. In that environment the user can still
access port 8081 via the "Web preview" feature, and that access is still
secure (since only that user can access it), but the hostname will
not be 'localhost'.

This change adds support for that use case by adding a "CLI_LOGIN"
environment variable to the Docker image. If that variable is set to
"true", then the startup script will trigger a CLI based OAuth flow if
the user is not already signed in.

### DIFF
--- a/containers/datalab/content/run.sh
+++ b/containers/datalab/content/run.sh
@@ -36,6 +36,21 @@ ERR_DEPLOY=6
 ERR_TUNNEL_FAILED=7
 ERR_GATEWAY_FAILED=7
 
+run_login() {
+  local login_cmd=${1:-"gcloud auth login"}
+
+  USER_EMAIL=`gcloud auth list --format="value(account)"`
+  if [[ -z "${USER_EMAIL}" ]]; then
+    local failed_login=""
+    ${login_cmd} || failed_login="true"
+    if [[ -n "${failed_login}" ]]; then
+      echo "Failed to log in to gcloud"
+      exit "${ERR_LOGIN}"
+    fi
+  fi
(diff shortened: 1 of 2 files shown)


containers/datalab/content/run.sh
```diff
@@ -56,6 +56,13 @@ if [[ -n "${GATEWAY_VM}" ]]; then
   fi
 fi
 
+if [[ "${CLI_LOGIN}" == "true" ]]; then
+  USER_EMAIL=`gcloud auth list --format="value(account)"`
+  if [[ -z "${USER_EMAIL}" ]]; then
+    gcloud auth login
+  fi
+fi
```
qimingj: Set USER_EMAIL after gcloud auth login so it is consistent with the flow below?

qimingj: Also, if auth failed it will exit gracefully with right message? Just double check.

ojarjur: @qimingj I've just gone ahead and split this out to a helper method, so that we have both flows operate the exact same way.

That should address both of your comments.

qimingj:
LGTM. Sorry for the delay!

PR #961, opened 2016-09-02
Pull the GCC source code from a mirror instead of FTP.

ojarjur:
The GNU project lists multiple mirrors for their source code and asks on their
site that one of the mirrors be used for downloading rather than reading
directly from their FTP server.

This change switches us from downloading directly from the GNU FTP server to
instead pulling from the first HTTP-based mirror listed in the US on this
site: https://www.gnu.org/prep/ftp.en.html


qimingj:
LGTM.

PR #965, opened 2016-09-08
Support hosts that block websockets.

ojarjur:
Previously, Datalab proxied web-socket connections over HTTP, so that
it could run in App Engine Flex (which does not support websockets).

That code was removed when we transitioned to the locally-run setup,
as we did not think that it was necessary anymore.

However, even in the locally-run setup, a user may want to run in
an environment that does not support web-sockets. One such example
is users who want to run Datalab using Google's Cloud Shell Web
Preview feature (https://cloud.google.com/shell/docs/using-web-preview).

This change adds support for such environments by conditionally
proxying web sockets over HTTP based on an environment variable that
can be set by the user.

The code for this is just the old proxy code from the App Engine
Flex version of Datalab, but wrapped in checks based on the new
user setting.

This works for the entirely-locally run version, but additional
work will be required for this to also work with kernel gateways,
as the proxied web socket connections look like cross-domain web
sockets, which are currently blocked.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -371,6 +371,10 @@ export function handleRequest(request: http.ServerRequest, response: http.Server
 
 function getBaseTemplateData(request: http.ServerRequest): common.Map<string> {
   var userId: string = userManager.getUserId(request);
+  var proxyWebSockets: string = process.env.PROXY_WEB_SOCKETS;
+  if (proxyWebSockets != 'true') {
+    proxyWebSockets = 'false';
+  }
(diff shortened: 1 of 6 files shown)


ojarjur:
FYI, I built and pushed an image with this change so that anyone who wants can try it.

To use this, open up the cloud shell and run this command:

    docker run -it -p 8081:8080 -v "${HOME}:/content" -e "PROXY_WEB_SOCKETS=true" gcr.io/cloud-datalab/datalab:local-cloud-shell

Then, click on the "Web Preview" button, select "Change port", and click on the "Port 8081" entry.

qimingj:
LGTM

PR #966, opened 2016-09-08
Make the left padding of cells consistent

ojarjur:
This change fixes a usability issue with the display of code and data cells
where they would be displayed with a different amount of left-padding depending
on their state (unselected, selected, or edit-mode).

This fixes that issue by setting the left padding to always be 4 pixels
regardless of what state the cell is in.


qimingj:
LGTM

PR #971, opened 2016-09-13
Add an API for getting the current user settings.

ojarjur:
This change extens the "_settings" endpoint so that it can be used
to fetch the current settings in addition to updating them.

The old behavior was:

    If both 'key' and 'value' are specified, then update that
    setting, otherwise return a 'bad-request' error.

The new behavior is:

    If 'key' and 'value' are specified then update that setting...

    Otherwise, if 'key' is specified then return the value of just
    that one setting...

    Otherwise, return the values of all the settings.


qimingj:
LGTM

Issue #964, opened 2016-09-07
Improve Datalab 'dark' theme 


parthea:
I noticed 2 issues while using the 'dark' theme from #954 

- When I highlight text in a cell, the background color of the text does not change. This prevents users from seeing which text is highlighted. Also, this makes cut and paste operations very difficult.
- When a DataFrame is displayed, odd rows are not visible when `%chart` is used. See screenshot. I've included code below to help reproduce the issue.

Run
```
data = [
        {'date': '2016/01/01', 'count': 10}, 
        {'date': '2016/01/02', 'count': 15}, 
        {'date': '2016/01/03', 'count': 12} 
       ]
%chart pie -d data
```
followed by
```
import pandas as pd
df = pd.DataFrame(data=data)
df
```

![odd rows in table not visible](https://cloud.githubusercontent.com/assets/5184014/18297599/d29edbf4-747f-11e6-8cdf-b443f3f62858.png)

parthea:
cc @ojarjur

ojarjur:
@parthea Yes, I agree that the dark theme needs a lot of work.

It's currently at the proof-of-concept stage and isn't yet ready for real use.

PR #975, opened 2016-09-15
Fix table and chart styling on dark theme

yebrahim:
Fixed dark theme styling for generated chart svgs as well as some dataframe tables.

Fixes issue #964

### DIFF
--- a/sources/web/datalab/static/dark.css
+++ b/sources/web/datalab/static/dark.css
@@ -155,16 +155,21 @@ code {
 .google-visualization-table {
   color: #000;
 }
-.gchart-table-cell {
-  background-color: #444;
-  color: #ddd;
-}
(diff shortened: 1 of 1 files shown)


sources/web/datalab/static/dark.css
```diff
@@ -155,16 +155,25 @@ code {
 .google-visualization-table {
   color: #000;
 }
-.gchart-table-cell {
+.gchart-table-cell, tr.google-visualization-table-tr-odd {
```
ojarjur: Can't this be combined with the style defined on line 166? They have the same body.

Regardless, I think that 'tr.google-visualization-table-tr-odd' should only occur in one of them.

yebrahim: Yep, this is duplicated. Will fix.

Issue #104, opened 2014-12-02
Support for themes / built in dark theme


Di-Ku:
This is definitely post-alpha and could be post-beta. An intangible delighter to create a more emotional attachment to the tool and the experience of using it. In particular, a sizeable segment prefers dark themes. Sublime Text is a great example that defaults to such a theme and gets a lot of mileage out of it. 

See http://www.damian.oquanta.info/posts/48-themes-for-your-ipython-notebook.html for some work that has been done.

yebrahim:
Fixed via #981.

PR #977, opened 2016-09-16
Exposing dark theme - Added a theme dropdown for controling light/dar…

yebrahim:
This change fixes the dark theme, and exposes it using a dropdown button under the user profile menu. It uses settings that are persisted by the server piece, and does XHR requests to load and save them. It reloads the stylesheet when a different theme is selected to avoid the user having to reload the whole pages, losing any changes they might have.

Issue #104


yebrahim:
More work needed here. Will close this one and send another.

PR #978, opened 2016-09-17
Fix an issue that 'custom.js' cannot be loaded.

qimingj:
I think the issue has been there for a while. And my recent commit might have changed the js load sequence, which exposes the issue.

According to https://github.com/googledatalab/datalab/blob/master/sources/web/datalab/templates/tree.html#L170, when require(['custom'], function()..) is called, it will load '/custom.js', which will return 404. If the line is executed later, jupyter will load 'custom.js' from '/custom/custom.js', which is served correctly.

### DIFF
--- a/sources/web/datalab/static.ts
+++ b/sources/web/datalab/static.ts
@@ -176,7 +176,7 @@ function requestHandler(request: http.ServerRequest, response: http.ServerRespon
       sendJupyterFile(path.substr(1), response);
     }
   }
-  else if (path.lastIndexOf('/custom.js') > 0) {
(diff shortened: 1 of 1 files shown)


qimingj:
Static handler kicks in only when the request path started with /static or /custom. See https://github.com/googledatalab/datalab/blob/master/sources/web/datalab/server.ts#L210. Usually >0 is fine, but /custom.js is special.

Another way to fix this is to change "custom : '/custom'," to "custom : '/custom/custom'," in all templates under https://github.com/googledatalab/datalab/tree/master/sources/web/datalab/templates. I don't see much difference though.

PR #980, opened 2016-09-20
Make it possible to use local pydatalab instead of git-cloning from pydatalab repo.

qimingj:
Right now it is hard to make changes to pydatalab locally and then have datalab container picking it up, because it always git-clone pydatalab from repo. The change adds a mechanism (build.sh <pydatalab location>) to optionally picking up pydatalab from a local directory.

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -130,15 +130,16 @@ RUN pip install google-cloud-dataflow==0.4.1
 ADD nbconvert /datalab/nbconvert
 ADD config/ipython.py /etc/ipython/ipython_config.py
 ADD config/nbconvert.py /etc/jupyter/jupyter_notebook_config.py
+# Directory "py" may be empty and in that case it will git clone pydatalab from repo
+ADD py /datalab/lib/pydatalab
 
 # Do IPython configuration and install build artifacts
 # Then link stuff needed for nbconvert to a location where Jinja will find it.
 # I'd prefer to just use absolute path in Jinja imports but those don't work.
 RUN ipython profile create default && \
     jupyter notebook --generate-config && \
-    mkdir -p /datalab/lib && \
-    cd /datalab/lib && \
-    git clone https://github.com/googledatalab/pydatalab.git && \
+    if [ -d /datalab/lib/pydatalab/.git ]; then echo "use local lib"; else \
+      git clone https://github.com/googledatalab/pydatalab.git /datalab/lib/pydatalab; fi && \
     cd /datalab/lib/pydatalab && \
     tsc --module amd --noImplicitAny --outdir datalab/notebook/static datalab/notebook/static/*.ts && \
     pip install . && \
--- a/containers/base/build.sh
+++ b/containers/base/build.sh
@@ -13,7 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Builds the Google Cloud DataLab base docker image
+# Builds the Google Cloud DataLab base docker image. Usage:
+#   build.sh [path_of_pydatalab_dir]
+# If [path_of_pydatalab_dir] is provided, it will copy the content of that dir into image.
+# Otherwise, it will get the pydatalab by "git clone" from pydatalab repo.
 
 # Build the docker image
+if [ -n "$1" ]; then rsync -avp $1 py; else mkdir -p py; fi
 docker build -t datalab-base .
+rm -rf py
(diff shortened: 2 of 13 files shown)


parthea:
This is a really helpful, and will definitely help with development!

I wanted to share another approach to using a local copy of pydatalab so that contributors have the option to choose based on their preference or use case. Contributors could add ` -v "${HOME}/Git/pydatalab/datalab/:/usr/local/lib/python2.7/dist-packages/datalab/"` to the ` docker run...` command and use their development copy of pydatalab in their datalab container.

qimingj:
Tony, thanks for sharing the great idea! I do prefer your approach of mapping in pydatalab, because it is faster (no need to rebuild) and you can change py files without restarting Datalab. It looks like this change is useful when creating containers for sharing with custom pydatalab.

containers/datalab/build.sh
```diff
@@ -29,7 +32,7 @@ rsync -avp ../../build/ build
 
 # Build the base docker image
 cd ../base
-docker build -t datalab-base .
+./build.sh $1
 cd ../datalab
```
yebrahim: It'd be good to also build the web server here

containers/base/Dockerfile
```diff
@@ -130,15 +130,16 @@ RUN pip install google-cloud-dataflow==0.4.1
...
 
 # Do IPython configuration and install build artifacts
 # Then link stuff needed for nbconvert to a location where Jinja will find it.
 # I'd prefer to just use absolute path in Jinja imports but those don't work.
 RUN ipython profile create default && \
     jupyter notebook --generate-config && \
-    mkdir -p /datalab/lib && \
-    cd /datalab/lib && \
-    git clone https://github.com/googledatalab/pydatalab.git && \
+    if [ -d /datalab/lib/pydatalab/.git ]; then echo "use local lib"; else \
```
ojarjur: Let's reformat these commands to split them over multiple lines so that they are more readable.

containers/base/Dockerfile
```diff
@@ -130,15 +130,16 @@ RUN pip install google-cloud-dataflow==0.4.1
 ADD nbconvert /datalab/nbconvert
 ADD config/ipython.py /etc/ipython/ipython_config.py
 ADD config/nbconvert.py /etc/jupyter/jupyter_notebook_config.py
+# Directory "py" may be empty and in that case it will git clone pydatalab from repo
+ADD py /datalab/lib/pydatalab
```
ojarjur: I *think* that if we rename the local directory to "pydatalab", then we can change the ADD instruction to:

    ADD py* /datalab/lib/

And then the base image will still be buildable with just "docker build .." (i.e. the build.sh script is optional unless you want to specify the pydatalib directory).

containers/base/build.sh
```diff
@@ -13,7 +13,12 @@
...
 # limitations under the License.
 
-# Builds the Google Cloud DataLab base docker image
+# Builds the Google Cloud DataLab base docker image. Usage:
+#   build.sh [path_of_pydatalab_dir]
+# If [path_of_pydatalab_dir] is provided, it will copy the content of that dir into image.
+# Otherwise, it will get the pydatalab by "git clone" from pydatalab repo.
 
 # Build the docker image
+if [ -n "$1" ]; then rsync -avp $1 py; else mkdir -p py; fi
```
ojarjur: If you go with the "ADD py* ..." suggestion, then this line would simply become:

    if [ -n "$1" ]; then rsync -avp "$1" pydatalab; fi

Either way, though, make sure that the "$1" in the rsync command is quoted (in case the directory contains whitespace characters).

containers/base/Dockerfile
qimingj: I wish that could work so we can simplify the build.sh. It seems "ADD py*" will fail the docker build if py* does not exist. Neither does "COPY py*" work:
https://github.com/docker/docker/issues/13045
So we may have to create an empty py dir. I'll keep looking for a better approach but if you have any ideas please let me know!

ojarjur: You're right; that doesn't work.

It seems that we need to require running the build.sh script now.

containers/base/build.sh
ojarjur: OK, the mkdir -p command needs to stay, but let's still wrap the "$1" in quotes.

containers/datalab/build.sh
qimingj: Perhaps in a different CL?

PR #982, opened 2016-09-20
Flexible pydatalab

qimingj:
(Sorry, I messed up the branch and then deleted it, so the comments are lost. But I think all of them are addressed).

Right now it is hard to make changes to pydatalab locally and then have datalab container picking it up, because it always git-clone pydatalab from repo. The change adds a mechanism (build.sh ) to optionally picking up pydatalab from a local directory.

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -130,15 +130,16 @@ RUN pip install google-cloud-dataflow==0.4.1
 ADD nbconvert /datalab/nbconvert
 ADD config/ipython.py /etc/ipython/ipython_config.py
 ADD config/nbconvert.py /etc/jupyter/jupyter_notebook_config.py
+# Directory "py" may be empty and in that case it will git clone pydatalab from repo
+ADD py /datalab/lib/pydatalab
 
 # Do IPython configuration and install build artifacts
 # Then link stuff needed for nbconvert to a location where Jinja will find it.
 # I'd prefer to just use absolute path in Jinja imports but those don't work.
 RUN ipython profile create default && \
     jupyter notebook --generate-config && \
-    mkdir -p /datalab/lib && \
-    cd /datalab/lib && \
-    git clone https://github.com/googledatalab/pydatalab.git && \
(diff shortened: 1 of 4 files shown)


containers/base/Dockerfile
```diff
@@ -130,15 +130,16 @@ RUN pip install google-cloud-dataflow==0.4.1
...
 
 # Do IPython configuration and install build artifacts
 # Then link stuff needed for nbconvert to a location where Jinja will find it.
 # I'd prefer to just use absolute path in Jinja imports but those don't work.
 RUN ipython profile create default && \
     jupyter notebook --generate-config && \
-    mkdir -p /datalab/lib && \
-    cd /datalab/lib && \
-    git clone https://github.com/googledatalab/pydatalab.git && \
+    if [ -d /datalab/lib/pydatalab/.git ]; then echo "use local lib"; else \
```
ojarjur: Please split this up into multiple lines so that it is easier to read.

containers/base/Dockerfile
```diff
@@ -130,15 +130,19 @@ RUN pip install google-cloud-dataflow==0.4.1
...
 # Then link stuff needed for nbconvert to a location where Jinja will find it.
 # I'd prefer to just use absolute path in Jinja imports but those don't work.
 RUN ipython profile create default && \
     jupyter notebook --generate-config && \
-    mkdir -p /datalab/lib && \
-    cd /datalab/lib && \
-    git clone https://github.com/googledatalab/pydatalab.git && \
+    if [ -d /datalab/lib/pydatalab/.git ]; then \
+        echo "use local lib"; \
+      else \
```
ojarjur: FYI, it looks like the indentation got a little messed up here and below on line 145 (off by 2 spaces).

containers/base/Dockerfile
qimingj: done.

containers/base/Dockerfile
qimingj: This is intentional. The convention is that no indentation indicates start of a new command. I consider "if ... fi" as a  single command, hence the 2 spaces before "else" and "fi".

PR #983, opened 2016-09-21
Improve accuracy the login/logout logic

ojarjur:
This change includes two improvements to the logic around logging in
and logging out:

1. Remove usage of the 'DATALAB_ENV' environment variable in the logic
   for determining if the user is signed-in, and instead directly
   query the 'gcloud' tool.
2. Remove the 'signout' button if the signed-in account is a service
   account. This is for running the Datalab UI inside of a GCE VM
   with a GCE service account. In that situation the service account
   is tied to the VM and cannot be detatched, so the 'signout' button
   did nothing in that case.

### DIFF
--- a/sources/web/datalab/auth.ts
+++ b/sources/web/datalab/auth.ts
@@ -53,7 +53,9 @@ function base64decodeSegment(str: string) {
 export function getGcloudAccount(): string {
   // Ask gcloud which account we are using.
   try {
-    var account = childProcess.execSync('gcloud config list --format "value(core.account)"', {env: process.env});
+    var account = childProcess.execSync(
+      'gcloud auth list --filter=status:ACTIVE --format "value(account)"',
+      {env: process.env});
     account = account.toString().trim();
     return account;
   } catch (err) {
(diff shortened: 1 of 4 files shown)


sources/web/datalab/auth.ts
```diff
@@ -146,8 +146,8 @@ function persistCredentials(tokens: any): string {
 }
 
 export function isSignedIn(): boolean {
-  // If not local, then we should have service account available so consider that signed in.
-  return (process.env.DATALAB_ENV != 'local' || fs.existsSync(appCredFile));
+  var gcloudAccount:string = getGcloudAccount();
```
qimingj: I tried the command in getGcloudAccount() ("gcloud config list --format "value(core.account)") when I already signed out, and it still returns my user id. Perhaps core.account does not reflect the sign-in status?

On the other hand, 'gcloud auth list --format="value(account)"' seems working for me.

ojarjur: You're right; the 'gcloud config' entry seems to be disjoint from login status.

I'll fix that to use 'gcloud auth list' instead.

ojarjur: Done.

PR #994, opened 2016-09-28
Add licenses for Sylvester and Parallel Coordinates

yebrahim:


containers/datalab/content/license.txt
```diff
@@ -4574,4 +4574,44 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+"""
+parcoords
```
qimingj: Same for divgrid? It is from https://github.com/syntagmatic/parallel-coordinates/blob/master/examples/lib/divgrid.js.

Issue #375, opened 2015-06-04
Better error handling for cell magics with single percent


gramster:
I am not sure we can necessarily fix this but if we can it would be good. I've been bitten by it a few times (at least now I'm aware) and I'm sure our users will be too.

If you enter a cell magic but use % instead of %% you get a syntax error. However, this can be very misleading - for example, if the cell is a SQL cell you may scratch your head looking for the syntax error in your SQL.

Here is an example, which shows how misleading this can look:

    %bigquery sql
    select count(*) from [googledata:sponge.target_1daysago]

      File "\<ipython-input-19-e228616ea1a6\>", line 2
        select count(*) from [googledata:sponge.target_1daysago]
                   ^
    SyntaxError: invalid syntax

gramster:
This has been taken care of by the changes for issue #551.

PR #997, opened 2016-09-29
With the exception of %sql, don't convert line magic to cell magic

### LINKED ISSUE #996: Line magic commands silently fail if they are the first line in a cell, and cell body is not empty
I noticed the following error in the Datalab console when using the `%storage` line magic command.

```
{"name":"app","hostname":"b86c9881eefb","pid":89,"type":"jupyter","level":30,"msg":"[9000]: ERROR:root:Cell magic `%%storage` not found (But line magic `%storage` exists, did you mean that instead?).\n","time":"2016-09-26T11:35:28.505Z","v":0}
```

The error appears in the console when the following code is executed, in an otherwise empty cell:

```
%storage read --object gs://cloud-datalab-samples/cars.csv --variable cars
print(cars)
```

Using a new cell, I can confirm that `cars` is not defined

```
>print(cars)
NameErrorTraceback (most recent call last)
<ipython-input-16-b33add6e2bbd> in <module>()
----> 1 print cars

NameError: name 'cars' is not defined
```

The code works as expected if I put `%storage` on the second line (first line can be empty, comment, or python code). For example, 

```
# This comment is here to work around the issue where %storage doesn't work if it is the first line in a cell. 
%storage read --object gs://cloud-datalab-samples/cars.csv --variable cars
print(cars)
```

The output is correctly shown as:

```
Year,Make,Model,Description,Price
1997,Ford,E350,"ac, abs, moon",3000.00
1999,Chevy,"Venture Extended Edition","",4900.00
1999,Chevy,"Venture Extended Edition",Very Large,5000.00
1996,Jeep,Grand Cherokee,"MUST SELL! air, moon roof, loaded",4799.00
```

parthea:
Fixes #996 

Restores the following functionality which would not work after #567:

```
%storage read --object gs://cloud-datalab-samples/cars.csv --variable cars
print(cars)
```

In #567 , line magic commands were automatically converted to cell magic if the cell body was not empty. Based on #375, it may be helpful to automatically convert line magic `%sql` to cell magic.

### DIFF
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -531,8 +531,8 @@ function initializeNotebookApplication(ipy, notebook, events, dialog, utils) {
 
     var originalExecute = kernel.prototype.execute;
     kernel.prototype.execute = function (code, callbacks, options) {
-      // If this is a line magic but has a non-empty cell body change it to a cell magic.
-      if (code.length > 2 && code[0] == '%' && code[1] != '%') {
+      // If this is a sql line magic but has a non-empty cell body change it to a cell magic.
+      if (code.startsWith('%sql')) {
         var lines = code.split('\n');
         if (lines.length > 1) {
           for (var i = 1; i < lines.length; i++) {


nikhilk:
I think we need to take a closer look at this -- we want to actively remove the differences between % and %% -- based on our usability studies, most users don't get the distinction.

We should make sure our magic implementations do the right thing regardless of how they are declared.

Specifically _all_ our magics should first of all work in both line mode and cell mode, with desirable behavior.

parthea:
@nikhilk Thanks for the explanation! I'm going to look into how to make magics work in both line and cell mode. I'm not quite sure how to solve this yet but I'll start experimenting.

gramster:
We adopted the convention in datalab that a cell can have just a single magic and it must start on the first line. This was needed both for some of the hybrid syntax highlighting we use and to allow % and %% to be used interchangeably. So to put this in a single cell is not valid:

    %storage read --object gs://cloud-datalab-samples/cars.csv --variable cars
    print(cars)

nikhilk:
As a starting point, I think we want all of our magics to be registered as line and cell magic (if not already).

What we may need to build is a whitelist of "native"/Jupyter-built-in magics, which are registered as only line magics to not be converted to %% even if the user has multiple lines in the cell. Or maybe the whitelist should contain the ones we add with datalab.

parthea:
@gramster @nikhilk Thanks for clarifying! I agree with all the comments. Assuming that (datalab specific) magic commands cannot co-exist in the same cell with non-magic code, I'll also look into displaying an error message to the user if this occurs.

PR #999, opened 2016-09-30
Check in the scripts for the release process.

ojarjur:
This adds to our source repository two scripts that were previously
just on a build server. By checking them in, we ensure that:

1. We can recreate the release process at any time without worrying
   about restoring from machine backups.
2. Our release process is version controlled so we can track
   when and why we change it.


tools/release/build.sh
```diff
@@ -0,0 +1,81 @@
...
+#
+# That involves a clean build (not using the Docker image cache) of
+# the datalab-base, datalab-gateway, and datalab images; tagging the
+# resulting images with the current date; and then pushing the
+# datalab and datalab-gateway images to the Google Container Registry.
+
+PROJECT_ID="${PROJECT_ID:-cloud-datalab}"
+TIMESTAMP=$(date +%Y%m%d)
+LABEL="${LABEL_PREFIX:-}${TIMESTAMP}"
+BACKEND_IMAGE="gcr.io/${PROJECT_ID}/datalab-gateway:${LABEL}"
```
qimingj: nit on the naming: standardize our naming for kernel gateway and datalab? Right now we have multiple names:
"backend", "kernel-gateway" for kernel gateway.
"datalab", "frondend" for datalab full (it is not really a frontend).
Maybe "kernel gateway image" and "datalab full image"? Maybe you have better names.

tools/release/build.sh
```diff
@@ -0,0 +1,81 @@
...
+# This script defines the automated build step of the Datalab release process.
+#
+# That involves a clean build (not using the Docker image cache) of
+# the datalab-base, datalab-gateway, and datalab images; tagging the
+# resulting images with the current date; and then pushing the
+# datalab and datalab-gateway images to the Google Container Registry.
+
+PROJECT_ID="${PROJECT_ID:-cloud-datalab}"
+TIMESTAMP=$(date +%Y%m%d)
+LABEL="${LABEL_PREFIX:-}${TIMESTAMP}"
```
qimingj: I cannot find LABEL_PREFIX definition?

tools/release/publish.sh
```diff
@@ -0,0 +1,44 @@
...
+gcloud docker -- pull ${BACKEND_IMAGE}
+docker tag -f ${BACKEND_IMAGE} gcr.io/${PROJECT_ID}/datalab-gateway:latest
+gcloud docker -- push gcr.io/${PROJECT_ID}/datalab-gateway:latest
+
+echo "Releasing the frontend image: ${FRONTEND_IMAGE}"
+gcloud docker -- pull ${FRONTEND_IMAGE}
+docker tag -f ${FRONTEND_IMAGE} gcr.io/${PROJECT_ID}/datalab:local
+gcloud docker -- push gcr.io/${PROJECT_ID}/datalab:local
+
+gsutil cp gs://${PROJECT_ID}/deploy/config_local.js ./config_local.js
```
qimingj: This is great!

tools/release/build.sh
ojarjur: It's optional.

The ":-" at the end means that it defaults to an empty string if the user has not defined it outside of the script.

I should probably add usage strings to these that outline what variables you can declare to modify their behavior.

tools/release/build.sh
ojarjur: I originally had these as just "GATEWAY_IMAGE" and "DATALAB_IMAGE", then changed them to backend/frontend to try to be more general.

The reasoning is that we've gotten some feedback that "Kernel Gateway" is jargon that won't be meaningful to anyone that isn't familiar with the internal workings of Jupyter.

As such, I was trying to avoid using that term unless necessary. That being said, I agree with the pushback that the main image isn't just a frontend.

I suppose it's also reasonable to expect anyone digging through the source code to be familiar with Jupyter terms, so I may switch back to "GATEWAY_IMAGE"/"DATALAB_IMAGE"

qimingj: "DATALAB_BACKEND" and "DATALAB_ALL"? Anyway, I am fine so long as we will be consistent about this everywhere.

tools/release/build.sh
qimingj: I see. Yes. Usage string will be helpful.

tools/release/build.sh
ojarjur: I've gone back to the earlier names.

PR #1000, opened 2016-09-30
Persist the last user path and set it as startup path

yebrahim:
Use the settings module to save and load the last path visited by the user so it's persistent across different visits

### DIFF
--- a/sources/web/datalab/server.ts
+++ b/sources/web/datalab/server.ts
@@ -47,6 +47,8 @@ let eulaDir = '/content/datalab/.config/eula';
  * The application settings instance.
  */
 var appSettings: common.Settings;
+var loadedSettings: common.Map<string> = null;
+var startup_path_setting = 'startuppath'
 
 /**
  * If it is the user's first request since the web server restarts,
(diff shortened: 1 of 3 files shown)


sources/web/datalab/server.ts
```diff
@@ -141,6 +152,11 @@ function handleRequest(request: http.ServerRequest,
       (path.indexOf('/files') == 0) ||
       (path.indexOf('/edit') == 0) ||
       (path.indexOf('/sessions') == 0)) {
+
+    if (path.indexOf('/tree') == 0) {
+        loadedSettings[startup_path_setting] = path
+        settings_.updateUserSetting(userId, startup_path_setting, path);
```
qimingj: updateUserSetting() does synchronous file writing so it is more expensive compared to other requests. Although each Datalab instance serves one user, I would avoid expensive synchronous operations per normal request ('/tree'), since it will block other requests.

Perhaps at least we should make the file writing asynchronous. Ideally there is a way to detect VM shutdown so we can write the user config to disk only once.

yebrahim: There's no reliable way to execute code before killing the process, as far as I know, so I'll just turn this into an async call.

PR #1003, opened 2016-10-04
Fix timing issues in the loading of the app bar

### LINKED ISSUE #998: File name not anymore displayed in the header
Datalab [Version: 0.5.20160929]

ojarjur:
The splitting out of the app bar into a single, reusable HTML file
introduced an unexpected timing issue where the Javascript code served
by the Jupyter notebook server was trying to modify DOM elements which
did not yet exist.

This change fixes that issue by ensuring that the app bar HTML is loaded
before the notebook server Javascript is loaded. Since the loading of
the app bar HTML requires JQuery, this required us to add JQuery as one
of the scripts loaded in each of the HTML templates.

This fixes #998

### DIFF
--- a/sources/web/datalab/templates/edit.html
+++ b/sources/web/datalab/templates/edit.html
@@ -11,6 +11,8 @@
   <link rel="stylesheet" href="/static/style/style.min.css" type="text/css" />
   <link rel="stylesheet" href="/static/style/datalab.css" type="text/css" />
   <link rel="stylesheet" id="themeStylesheet" href="/static/style/custom.css" type="text/css" />
+
+  <script src="/static/components/jquery/jquery.min.js"></script>
 </head>
 <body class="edit_app"
   data-project=""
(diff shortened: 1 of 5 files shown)


sources/web/datalab/templates/edit.html
```diff
@@ -11,6 +11,8 @@
   <link rel="stylesheet" href="/static/style/style.min.css" type="text/css" />
   <link rel="stylesheet" href="/static/style/datalab.css" type="text/css" />
   <link rel="stylesheet" id="themeStylesheet" href="/static/style/custom.css" type="text/css" />
+
+  <script src="/static/components/jquery/jquery.min.js"></script>
```
yebrahim: why do we need these explicitly? don't we get jquery for free from jupyter?

ojarjur: We need to add it explicitly because we are now using it before we load any Jupyter code.

PR #1009, opened 2016-10-08
Resolve issue where input text is hard to read in dark theme

### LINKED ISSUE #1008: Input text is hard to read when using input() or raw_input() in Dark Theme
See red rectangle around the text that is hard to read.

darkthemerawinput

parthea:
Fixes #1008

### DIFF
--- a/sources/web/datalab/static/dark.css
+++ b/sources/web/datalab/static/dark.css
@@ -113,6 +113,16 @@ div.output {
 div.output_stderr {
   background-color: #a94442;
 }
+input.raw_input {
+  color: #ddd;
+  background-color: #333;
+}
+input.raw_input::selection {
+  background: rgba(85, 85, 85, 0.99);
+}
+input.raw_input::-moz-selection {
+  background: rgba(85, 85, 85, 0.99);
+}
 .CodeMirror {
   color: #ddd;
   background: #333;


sources/web/datalab/static/dark.css
```diff
@@ -113,6 +113,16 @@ div.output {
...
 }
+input.raw_input {
+  color: #ddd;
+  background-color: #333;
+}
+input.raw_input::selection {
+  background: #ccc;
+}
+input.raw_input::-moz-selection {
+  background: #999;
```
yebrahim: Curious why this is different from the ::seletion rule

parthea: I've pushed a new version using `rgba()` based on [this](http://stackoverflow.com/questions/7224445/css3-selection-behaves-differently-in-ff-chrome) StackOverflow post.

`::selection` is not in the officially CSS spec, but it was published in a [draft of Pseudo-Elements Level 4](https://www.w3.org/TR/css-pseudo-4/).

From [Mozilla Developer Network](https://developer.mozilla.org/en-US/docs/Web/CSS/::selection),

> Gecko is the only engine requiring the prefix. Due to the fact that the CSS parsing rules require dropping the whole rule when encountering an invalid pseudo-element, two separate rules must be written: ::-moz-selection, ::selection {...}. The rule would be dropped on non-Gecko browsers as ::-moz-selection is invalid on them.

yebrahim: Awesome! Thanks for the investigation and fix.

parthea: > It would actually be great if we can merge the two CSS rules into one that takes both selectors. 

Unfortunately, the following didn't work in my testing.

```
input.raw_input::selection, input.raw_input::-moz-selection {
  background: rgba(85, 85, 85, 0.99);
}
```
The following snippet from [Mozilla Developer Network](https://developer.mozilla.org/en-US/docs/Web/CSS/::selection) also reaffirms the test result:

> Due to the fact that the CSS parsing rules require dropping the whole rule when encountering an invalid pseudo-element, two separate rules must be written..

ojarjur:
Sorry for the delay getting this merged; for some reason it seems that GitHub never triggered the webhook to trigger the automated checks, so the required commit statuses never got set.

PR #1010, opened 2016-10-08
Remove the keyboard shortcut to move cells up/down from the documentation

### LINKED ISSUE #777: Check and fix any incorrect key bindings
Shift-J/K are documented as being used to navigate between cells but are not working (at least on a Mac).

We need to go through all the documented key bindings and check which ones don't work and fix them or the documentation.

parthea:
Closes #777 

The following keyboard shortcuts are no longer supported in [Jupyter Notebook 4.2](https://github.com/jupyter/notebook/blob/4.2.x/notebook/static/notebook/js/keyboardmanager.js):
- Ctrl+K - move cell up
- Ctrl+J - move cell down

I've confirmed that the other keyboard shortcuts documented in Datalab are still functional.

### DIFF
--- a/sources/web/datalab/templates/nb.html
+++ b/sources/web/datalab/templates/nb.html
@@ -326,8 +326,6 @@
       <div><code>m</code>: to markdown</div>
       <div><code>Up</code>: select previous cell</div>
       <div><code>Down</code>: select next cell</div>
-      <div><code>Ctrl+K</code>: move cell up</div>
-      <div><code>Ctrl+J</code>: move cell down</div>
       <div><code>a</code>: insert cell above</div>
       <div><code>b</code>: insert cell below</div>
       <div><code>x</code>: cut cell</div>


parthea:
@yebrahim Thanks for reviewing this!

Yes, I was able to get the quick help to work on Ubuntu/Chrome. I haven't tried Mac. 

Can you try the following steps?

- In otherwise empty cell, run the following:
```
import datalab.bigquery as bq
```
- Type `bq.Query` in a new cell. With the cursor at the end of `bq.Query`, try the shift+tab shortcut.
quickhelp

nikhilk:
A notebook is in two modes - command mode (cell is selected) and edit mode (code cell is activated). Things like shortcuts to move cells only apply to command mode.

parthea:
Note: The keyboard shortcut to move cells up/down has been removed from Jupyter (ipython/ipython#7732).

yebrahim:
This is a shame, I wonder why they're taking it away! Thanks for the fix.

PR #1015, opened 2016-10-10
Add a mini-toolbar for each cell to expose various operations

### LINKED ISSUE #1004: Support for collapsing/expanding code cells

yebrahim:
This change adds a small toolbar that appears on each cell on hover that contains buttons that do:
- Run cell
- Clear cell output
- Collapse cell output
- Collapse entire cell
Collapse preferences are also persisted in the cell metadata and loaded on notebook load.

[Added screenshots for review]
image
image

Fixes #1004

### DIFF
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -281,17 +281,20 @@ var CELL_METADATA_COLLAPSED = 'hiddenCell';
 var CELL_METADATA_OUTPUT_COLLAPSED = 'hiddenOutput';
 
 function toggleCollapseCell(cell, hide) {
-  let isHidden = typeof(hide) !== 'undefined' ? !hide : cell.element.find('div.input')[0].style.display === 'none';
+  let inputDiv = cell.cell_type === 'code' ? 'div.input' : 'div.inner_cell';
+  let isHidden = typeof(hide) !== 'undefined' ? !hide : cell.element.find(inputDiv)[0].style.display === 'none';
   collapseSpan = cell.element.find('span.collapse-cell')[0];
   if (isHidden) {
-    cell.element.find('div.input').show();
-    cell.element.find('div.output_wrapper').show();
+    cell.element.find(inputDiv).show();
+    if (cell.cell_type == 'code')
+      cell.element.find('div.output_wrapper').show();
     cell.element.find('div.cellPlaceholder').hide();
     collapseSpan.classList.add('fa-eye-slash');
     collapseSpan.classList.remove('fa-eye');
   } else {
-    cell.element.find('div.input').hide();
-    cell.element.find('div.output_wrapper').hide();
+    cell.element.find(inputDiv).hide();
+    if (cell.cell_type == 'code')
+      cell.element.find('div.output_wrapper').hide();
     cell.element.find('div.cellPlaceholder').show();
     collapseSpan.classList.remove('fa-eye-slash');
     collapseSpan.classList.add('fa-eye');
@@ -314,41 +317,29 @@ function toggleCollapseCellOutput(cell, hide) {
   cell.metadata[CELL_METADATA_OUTPUT_COLLAPSED] = !isHidden;
 }
 
+function createCellMiniToolbarButton(classNames, title, callback) {
+  let buttonDiv = document.createElement('button');
+  buttonDiv.className = 'btn btn-default';
+  buttonDiv.innerHTML = '<span class="' + classNames + '"></span>';
+  buttonDiv.title = title;
+  buttonDiv.addEventListener('click', callback);
+  return buttonDiv;
+}
+
 /**
  * Patch the cell's element to add a collapse/expand button
  */
 function addCellMiniToolbar() {
   Jupyter.notebook.get_cells().forEach(function(cell) {
     
-    if (cell.cell_type !== 'code') {
-      return;
-    }
-
     let hoverableDiv = document.createElement('div');
     hoverableDiv.className = 'hoverable';
 
     // collapse cell button
-    let collapseButton = document.createElement('button')
-    collapseButton.className = 'btn btn-default';
-    collapseButton.innerHTML = '<span class="collapse-cell fa fa-eye-slash"></span>';
-    collapseButton.title = 'Collapse/Expand cell';
-    collapseButton.addEventListener('click', function() {
+    let collapseButton = createCellMiniToolbarButton('collapse-cell fa fa-eye-slash', 'Collapse/Expand cell', function() {
       toggleCollapseCell(cell);
     });
     hoverableDiv.appendChild(collapseButton);
-
-    // collapse output button
-    let hideButton = document.createElement('button')
-    hideButton.className = 'btn btn-default';
-    hideButton.innerHTML = '<span class="collapse-output fa fa-compress"></span>';
-    hideButton.title = 'Collapse/Expand output';
-    hideButton.addEventListener('click', function() {
-      toggleCollapseCellOutput(cell);
-    });
-    cell.element.find('div.output_collapsed')[0].addEventListener('click', function() {
-      toggleCollapseCellOutput(cell, false /*hide*/);
-    });
-    hoverableDiv.appendChild(hideButton);
 
     // cell collapse placeholder
     let placeholderDiv = document.createElement('div');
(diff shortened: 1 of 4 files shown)


sources/web/datalab/static/datalab.js
```diff
@@ -277,6 +277,166 @@ function initializePage(dialog, saveFn) {
...
+  buttonDiv.title = title;
+  buttonDiv.addEventListener('click', callback);
+  return buttonDiv;
+}
+
+/**
+ * Patch the cell's element to add the hoverable minitoolbar
+ */
+function addCellMiniToolbar() {
+  Jupyter.notebook.get_cells().forEach(function(cell) {
```
qimingj: Can all the new functions be moved to initializeNotebookApplication(), and instead of using "Jupyter.notebook" we could use "notebook" parameter of the function? It will be more consistent by following the requireJS pattern.

sources/web/datalab/static/datalab.js
```diff
@@ -277,6 +277,166 @@ function initializePage(dialog, saveFn) {
...
+    if (cell.cell_type === 'code') {
+      // collapse output button
+      let hideButton = createCellMiniToolbarButton(
+        'collapse-output fa ' + COLLAPSE_OUTPUT_BUTTON_CLASS,
+        'Collapse/Expand output',
+        function() {
+          toggleCollapseCellOutput(cell);
+        }
+      );
+      cell.element.find('div.output_collapsed')[0].addEventListener('click', function() {
```
qimingj: Multiple cell outputs are possible (I think multiple calls of IPython.core.display.*() will do it). Can we make sure all of them are added the minitool bar?

sources/web/datalab/static/datalab.js
```diff
@@ -277,6 +277,166 @@ function initializePage(dialog, saveFn) {
   });
 }
 
+// constants for minitoolbar operations
+const CELL_METADATA_COLLAPSED = 'hiddenCell';
+const CELL_METADATA_OUTPUT_COLLAPSED = 'hiddenOutput';
```
ojarjur: Why is this not just the existing Jupyter [cell metadata of "collapsed"](https://nbformat.readthedocs.io/en/latest/format_description.html#metadata)?

I would prefer to not define our own metadata key if there is an existing one that does the same thing.

sources/web/datalab/static/datalab.js
```diff
@@ -277,6 +277,166 @@ function initializePage(dialog, saveFn) {
...
+    cell.element.find(inputDiv).show();
+    // for code cells, also show the output div
+    if (cell.cell_type == 'code')
+      cell.element.find('div.output_wrapper').show();
+    cell.element.find('div.cellPlaceholder').hide();
+    collapseSpan.classList.add(UNCOLLAPSE_BUTTON_CLASS);
+    collapseSpan.classList.remove(COLLAPSE_BUTTON_CLASS);
+  } else {
+    cell.element.find(inputDiv).hide();
+    // for code cells, also hide the output div
```
ojarjur: I think there is a compelling use case for showing the output but not the input (e.g. for dashboards where the input is an implementation detail and what the user really cares about is the output).

As such, I'd prefer if the collapsing of the code and output could be separated.

sources/web/datalab/static/datalab.js
```diff
@@ -277,6 +277,166 @@ function initializePage(dialog, saveFn) {
...
+const UNCOLLAPSE_OUTPUT_BUTTON_CLASS = 'fa-expand';
+const CLEAR_BUTTON_CLASS = 'fa-minus-square-o';
+const RUN_BUTTON_CLASS = 'fa-play';
+const COLLAPSED_CELL_INNER_HTML = '-- Collapsed cell --';
+
+/**
+ * Collapse entire cell
+ */
+function toggleCollapseCell(cell, hide) {
+  // the cell is currently hidden if its input div is
```
ojarjur: It feels odd to me to base this on the view (the hidden state of the input div) instead of the model (the existing cell metadata).

What is the reason for doing it like that?

sources/web/datalab/static/datalab.js
yebrahim: I don't think that's the best spot for it. I need to be sure all the cells have rendered before I go about modifying them, that's why I'm relying on the event, does initializeNotebookApplication() guarantee this?

sources/web/datalab/static/datalab.js
yebrahim: They are. All of them are inside an `output_wrapper` div, and that's what the cell collapse code hides/shows. As for hiding the output only, it's actually done by a call to Jupyter's API `.collapse_output()` so it takes care of that as well.

sources/web/datalab/static/datalab.js
yebrahim: Sure.

sources/web/datalab/static/datalab.js
yebrahim: No good reason other than the lack of a full model behind our view, I missed this though. :) I fixed it to use the metadata.

sources/web/datalab/static/datalab.js
```diff
@@ -277,6 +277,168 @@ function initializePage(dialog, saveFn) {
...
+const COLLAPSED_CELL_INNER_HTML = '-- Collapsed cell --';
+
+/**
+ * Collapse entire cell
+ */
+function toggleCollapseCell(cell, hide) {
+  if (!(CELL_METADATA_COLLAPSED in cell.metadata)) {
+    return;
+  }
+  let isHidden = typeof(hide) !== 'undefined' ? !hide : cell.metadata[CELL_METADATA_COLLAPSED];
```
ojarjur: It took me a fair amount of time to understand this code because it wasn't clear to me that 'isHidden' was a description of the state prior to the function call.

In part, that was due to the fact that the 'isHidden' variable doesn't really represent that... if you call toggleCollapseCell(cell, true), then 'isHidden' will be false even if the cell was already collapsed.

I think this would be clearer if we got rid of the 'isHidden' variable and instead simply set the 'hide' variable to the negation of the 'cell.metadata[CELL_METADATA_COLLAPSED]' expression if it is undefined.

sources/web/datalab/static/datalab.js
```diff
@@ -277,6 +277,168 @@ function initializePage(dialog, saveFn) {
...
+}
+
+/**
+ * Collapse output of a cell
+ */
+function toggleCollapseCellOutput(cell, hide) {
+  if (!(CELL_METADATA_OUTPUT_COLLAPSED in cell.metadata)) {
+    return;
+  }
+  let isHidden = typeof(hide) !== 'undefined' ? !hide : cell.metadata[CELL_METADATA_OUTPUT_COLLAPSED];
```
ojarjur: Same comment here about replacing usage of 'isHidden' with references to the 'hide' variable (while dealing with the fact that those are negated).

sources/web/datalab/static/datalab.js
```diff
@@ -277,6 +277,168 @@ function initializePage(dialog, saveFn) {
...
+const UNCOLLAPSE_OUTPUT_BUTTON_CLASS = 'fa-expand';
+const CLEAR_BUTTON_CLASS = 'fa-minus-square-o';
+const RUN_BUTTON_CLASS = 'fa-play';
+const COLLAPSED_CELL_INNER_HTML = '-- Collapsed cell --';
+
+/**
+ * Collapse entire cell
+ */
+function toggleCollapseCell(cell, hide) {
+  if (!(CELL_METADATA_COLLAPSED in cell.metadata)) {
```
ojarjur: If we are short-circuiting this scenario, then how does the user initially add this metadata to the cell?

sources/web/datalab/static/datalab.js
```diff
@@ -277,6 +277,168 @@ function initializePage(dialog, saveFn) {
...
+    // collapse output button
+    let hideButton = createCellMiniToolbarButton(
+      'collapse-output fa ' + COLLAPSE_OUTPUT_BUTTON_CLASS,
+      'Collapse/Expand output',
+      function() {
+        toggleCollapseCellOutput(cell);
+      }
+    );
+    cell.element.find('div.output_collapsed')[0].addEventListener('click', function() {
+      toggleCollapseCellOutput(cell, false /*hide*/);
```
ojarjur: > false /\*hide\*/

This was initially confusing to me, as the variable is named 'hide', and we are setting that to false, so this means 'do not hide'. However, the comment just says 'hide', which (at first) reads as meaning the exact opposite of what it does.

Can we maybe split "expandCellOutput" and "collapseCellOutput" helper methods out of the current body of "toggleCollapseCellOutput", and then just directly call "expandCellOutput"?

yebrahim:
Closing this PR because requirements changed, will open a new one that addresses all comments and handles the new requirements.

PR #1019, opened 2016-10-13
Add code cell collapse/uncollapse toggle

### LINKED ISSUE #1004: Support for collapsing/expanding code cells

yebrahim:
This PR adds a small toolbar to the left of code cells that currently has only one button to collapse/uncollapse the cell. This works only with code cells. When collapsed, code cells show the first line of their input.

Fixes #1004

image

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,0 +1,4 @@
+cd sources/web
+./build.sh
+cd ../../containers/datalab
+./build.sh && ./run.sh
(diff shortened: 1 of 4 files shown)


brokenairplane:
Is it possible to set a cell to be collapsed by default? The use case is a tutorial where a cell containing the solution is hidden until the user wants to expand it and see the solution.

yebrahim:
The collapse/expand state is saved in the notebook itself, so after collapsing the notebook, make sure you save it, and it'll keep that state.
Note, however that after the user expands it, that will likely be saved too. We don't have a way of marking a cell to always show up collapsed.

PR #1028, opened 2016-10-20
Fix a build breakage introduced by a new version of Pillow.

ojarjur:
For now we simply pin Pillow to the previous, working version.


techtonik:
Where is the bug report for the breakage?

PR #1035, opened 2016-10-26
Add code collapse button to code cells

yebrahim:
This change adds a button to the end of the input div to collapse the input area and keep only output area (and widgets if any) visible. It works just like the cell collapse button, visible only on selected cells, and stays visible for collapsed cells even if not selected.

Screenshots:
image

image

### DIFF
--- a/sources/web/datalab/static/datalab.css
+++ b/sources/web/datalab/static/datalab.css
@@ -327,7 +327,7 @@ div.cell {
   border-radius: 0px;
   padding: 0px;
   padding-left: 0px;
-  margin-bottom: 10px;
+  margin-bottom: 15px;
   position: relative;
 }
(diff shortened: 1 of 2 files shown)


nikhilk:
Can you try moving the widget to the left in the gutter area, or a
variation to the right, so this doesn't interrupt the normal flow.

Are we also getting rid of the cell collapsing widget?

If not, what shows up when both input is collapsed and the cell is also
collapsed.

yebrahim:
The gutter area isn't easy to manipulate, and will be very crammed if we add the collapse widget there.

We're not getting rid of the cell collapse widget, I suggest we keep them both either on the left or on the right. If you collapse the cell it hides everything inside it, regardless whether code is collapsed.

yebrahim:
This is what I think is a better approach:

image

PR #1036, opened 2016-10-27
Add mathjax by default to all notebooks

yebrahim:
Also fix Dockerfile extra endings


nikhilk:
I believe this is a huge library (relatively speaking)... we should make this opt-in, rather than included by default IMO.

yebrahim:
It's 360 kb when added from the CDN so not that big. It's also worth noting that this is enabled by default in Jupyter, so this might be expected from Datalab as well.

PR #1037, opened 2016-10-28
Add more buttons to cell minitoolbar: code collapse, run, and clear

yebrahim:
Added more buttons to the minitoolbar and converted it to a dropdown menu with a hamburger button. The new options added are: code collapse, run, and clear cell. Hiding is done mostly in CSS, except for parts where JS code is needed (hiding ipywidgets).

image

### DIFF
--- a/sources/web/datalab/static/datalab.css
+++ b/sources/web/datalab/static/datalab.css
@@ -327,7 +327,7 @@ div.cell {
   border-radius: 0px;
   padding: 0px;
   padding-left: 0px;
-  margin-bottom: 10px;
+  margin-bottom: 15px;
   position: relative;
 }
 div.cell.selected, div.cell.text_cell.selected, div.cell.code_cell.selected {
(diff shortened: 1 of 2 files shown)


yebrahim:
Sure, shortening and changing labels according to states makes sense.

However, I don't think disabling hiding code/collapsing cell in certain situations improves user experience, it rather complicates it in my opinion. It's not immediately obvious to the user why collapsing works in some scenarios and not in others, unless we have a richer UI that can give some feedback to indicate that e.g. cells without output cannot be collapsed.

VelizarVESSELINOV:
Is it possible to display the content of the first row of the cell, if it is something like this:
"""Description of the cell content."""

It is very difficult to understand the content of cells, if you have long list of this:
image

yebrahim:
If you're using the `Collapse` option, you first line of the cell does appear in the collapsed div. The `Hide Code` option is meant to hide the entire input area of the cell. You use that when you, for example, want to show output only, or hide the code from the user in a tutorial or something similar.

You could also add another prior cell that explains what this cell is about.

PR #1046, opened 2016-11-08
Support running datalab from a custom root dir.

craigcitro:
This adds basic support for running datalab from a custom root dir (called
`DATALAB_ROOT` throughout), mainly for use **outside** a Docker container.
This allows us to start setting up some tests which don't require Docker to run.

The configuration here comes in two parts:

1. The shell scripts and config in `containers/base` and `containers/datalab`
now uniformly use a `${DATALAB_ROOT}` argument before all absolute paths. By
default, this is set to `""`, making this a no-op.

2. The Datalab webserver now respects a new argument in `settings.json`,
called `datalabRoot`. This is used throughout whenever creating or referring
to an absolute path; this required a few small tweaks, namely to move
top-level constants into functions (since we don't know the full path until
the app is initializing).

PTAL @ojarjur @yebrahim @qimingj

### DIFF
--- a/containers/base/config/nbconvert.py
+++ b/containers/base/config/nbconvert.py
@@ -15,7 +15,7 @@
 """IPython nbconvert configuration for Google Cloud DataLab."""
 
 import os
-nbconvert_dir = os.path.join(os.getenv('DATALB_ROOT', '/'), 'datalab/nbconvert')
+nbconvert_dir = os.path.join(os.getenv('DATALAB_ROOT', '/'), 'datalab/nbconvert')
 
 c = get_config()
 c.TemplateExporter.template_path.insert(0, nbconvert_dir)
--- a/containers/base/config/nbconvert.py
+++ b/containers/base/config/nbconvert.py
@@ -14,6 +14,9 @@
 
 """IPython nbconvert configuration for Google Cloud DataLab."""
 
+import os
+nbconvert_dir = os.path.join(os.getenv('DATALB_ROOT', '/'), 'datalab/nbconvert')
+
 c = get_config()
-c.TemplateExporter.template_path.insert(0, '/datalab/nbconvert')
+c.TemplateExporter.template_path.insert(0, nbconvert_dir)
 c.HTMLExporter.template_file = 'html'
--- a/sources/web/datalab/auth.ts
+++ b/sources/web/datalab/auth.ts
@@ -55,6 +55,7 @@ function appCredFile(): string {
   return path.join(gcloudDir(), '/application_default_credentials.json');
 }
 
+// Path for shared boto config (used by gsutil).
 function botoFile(): string {
   return path.join(appSettings.datalabRoot, '/etc/boto.cfg');
 }
@@ -158,7 +159,6 @@ function saveBotoFile(tokens: any) {
  * Save the tokens in a credentials file that Datalab and gcloud can both use.
  */
 function persistCredentials(tokens: any): string {
-  
   if (!fs.existsSync(gcloudDir())) {
     fs.mkdirSync(gcloudDir());
   }
--- a/sources/web/datalab/auth.ts
+++ b/sources/web/datalab/auth.ts
@@ -22,6 +22,7 @@ import fs = require('fs');
 import google = require('googleapis');
 import http = require('http');
 import logging = require('./logging');
+import path = require('path');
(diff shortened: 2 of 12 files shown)


craigcitro:
(Note -- as I sent this, I realized that a few of the internal functions on the TypeScript side still need docstrings. I'll take care of that in the morning.)

containers/base/config/nbconvert.py
```diff
@@ -14,6 +14,9 @@
 
 """IPython nbconvert configuration for Google Cloud DataLab."""
 
+import os
+nbconvert_dir = os.path.join(os.getenv('DATALB_ROOT', '/'), 'datalab/nbconvert')
```
chmeyers: You misspeelled DATALAB_ROOT here.

craigcitro: good catch! thanks.

craigcitro:
PTAL @chmeyers @qimingj @ojarjur @yebrahim 

Based on some offline conversation with Bradley, I've slightly tweaked the original plan, in the following way:

* I no longer try to set `${DATALAB_ROOT}` in `containers/datalab/content/run.sh`. This was slightly inconsistent with the other shell scripts there, and wasn't actually necessary for our use case.
* I've tweaked the logic in a few more places to correctly thread in and use the global `settings` object.
* I added the request url to an error message, since it was hard to find the source without it. If that's a bad idea (say because these paths can be sensitive and get copy-pasted around), let me know.

(I'll squash and clean up the commit message once we've got an LGTM, but leaving as separate commits now for easier comparison.)

sources/web/datalab/auth.ts
```diff
@@ -30,16 +31,35 @@ var oauth2Client: any = undefined;
...
+}
+
+// Path for application default credentials.
+function appCredFile(): string {
+  return path.join(gcloudDir(), '/application_default_credentials.json');
+}
+
+// Path for shared boto config (used by gsutil).
+function botoFile(): string {
+  return path.join(appSettings.datalabRoot, '/etc/boto.cfg');
```
qimingj: Will the boto file still work for gsutil if it is under a different dir? This can be tested easily by running gsutil on a GCS object that requires permissions to access. Do we need to set "BOTO_CONFIG" env var? It is unfortunate that gsutil does not use gcloud credentials so we have to deal with it separately.

sources/web/datalab/auth.ts
```diff
@@ -30,16 +31,35 @@ var oauth2Client: any = undefined;
...
   'https://www.googleapis.com/auth/cloud-platform',
 ];
 
+// Path to the root of the gcloud configuration.
+function gcloudDir(): string {
+  return path.join(appSettings.datalabRoot, '/content/datalab/.config');
+}
+
+// Path for user-specific credentials.
+function userCredFile(): string {
```
qimingj: Does Datalab's auth flow works with this change? Here we save credential files to a custom dir but in pydatalab (https://github.com/googledatalab/pydatalab/blob/master/datalab/context/_utils.py#L53), we get credentials first from default credentials, and then try CLOUDSDK_CONFIG dir (which is not set unless we run setup-env.sh), and neither will retrieve credentials from the custom dir?

Or we actually require users to run gcloud auth login before running datalab outside docker, so that we never need to go through the auth flow?

craigcitro: Yes, a caller wanting to set a custom root dir will need to ensure that they've set `CLOUDSDK_CONFIG` appropriately, or that ADC are available (since that will override in both cases).

In particular, I've confirmed that if I start datalab with credentials at `userCredFile()`, everything works as expected. But in that sense I'm sidestepping this code for *setting* credentials.

sources/web/datalab/auth.ts
craigcitro: (This is basically tied up with the one above.)

sources/web/datalab/auth.ts
qimingj: Ah. Okay. Thanks for explaining this. Maybe someday we will have another run_no_docker.sh.

craigcitro:
Awesome -- @qimingj do you want to merge, or does anyone else want to chime in?

qimingj:
I will squash and merge now. If anyone has more comments we can discuss further.

PR #1047, opened 2016-11-10
Run ungit inside the datalab container

yebrahim:
This PR adds [ungit](https://github.com/FredrikNoren/ungit) inside the datalab container. It runs on port 8083 inside the container, and gets its requests reverse proxied so only the one port used by datalab is open outside the container.

I added a git icon in the app bar to open ungit in the current location. It finds the current directory path and points ungit to it. Note that ungit will ignore any trailing subdirectories after the git root dir.

image
image

### DIFF
--- a/sources/web/datalab/reverseProxy.ts
+++ b/sources/web/datalab/reverseProxy.ts
@@ -48,12 +48,22 @@ export function getRequestPort(request: http.ServerRequest, path: string): strin
 }
 
 /**
+ * Checks if a request should be exempted from reverse proxying to the gateway
+ */
+function shouldSkipGateway(port: String) {
+  // skip requests to 8083 for ungit
+  if (port === '8083')
+    return true;
+  return false;
+}
+
+/**
  * Handle request by sending it to the internal http endpoint.
  */
 export function handleRequest(request: http.ServerRequest,
                               response: http.ServerResponse,
                               port: String) {
-  if (process.env.KG_URL) {
+  if (process.env.KG_URL && !shouldSkipGateway(port)) {
     proxy.web(request, response, { target: process.env.KG_URL });
   }
   else {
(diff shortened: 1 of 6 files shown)


sources/web/datalab/reverseProxy.ts
```diff
@@ -48,12 +48,22 @@ export function getRequestPort(request: http.ServerRequest, path: string): strin
 }
 
 /**
+ * Checks if a request should be exempted from reverse proxying to the gateway
+ */
+function shouldSkipGateway(port: String) {
```
ojarjur: It took me a long time to realize why this logic was doing what it was doing, and I think that a slightly different structure would have made it easier to understand the reasoning.

I'd suggest the following:

1. Change the name of this function to something like 'isLocalPort'
2. Reorder the conditional branches in the 'handleRequest' method. With this change, what was previously the `else` branch is conceptually the default branch, so it feels like it should come first in the order.

Alternatively, we could make the 'isLocalPort' method check the 'KG_URL' environment variable.

yebrahim: I agree the naming isn't good. I think it's a better idea to make the function check the KG_URL though, but it should be checking if traffic should be proxied, not local. Please take a look and let me know what you think.

sources/web/datalab/static/appbar.html
```diff
@@ -13,6 +13,11 @@
   </span>
   <div class="btn-toolbar pull-right">
     <div class="btn-group">
+      <button id="ungitButton" title="Open ungit here" class="toolbar-btn">
```
chmeyers: One comment I heard from Nikhil: Should this button (and ungit) only be there when running on a VM?

yebrahim: Why not keep a consistent experience for all three supported scenarios?

PR #1048, opened 2016-11-14
Added a release rollback script

yebrahim:
The rollback relies on a 'previous' variable defined in the config_local.js file. This PR updates the publish script to put the current release as previous before publishing the new one. This way the rollback script can read the previous release version and copy its config_local.js to be the current one, and release its corresponding images as well.

### DIFF
--- a/tools/release/rollback.sh
+++ b/tools/release/rollback.sh
@@ -27,19 +27,23 @@
 # 1. "PROJECT_ID": Sets the name of the target project where the
 #    images will be pulled from and pushed to. Defaults to "cloud-datalab"
 # 2. "ROLLBACK_BUILD": The label of the build to pull down and rollback to.
-#    If omitted, this will extract the 'last' field from the config_local.js
+#    If omitted, this will extract the 'porevious' field from the config_local.js
 
 PROJECT_ID="${PROJECT_ID:-cloud-datalab}"
 
 gsutil cp gs://${PROJECT_ID}/deploy/config_local.js ./config_local.js
-REGEX='latest: ([0-9]+),.*last: ([0-9]+),'
-[[ $(cat config_local.js) =~ $REGEX ]]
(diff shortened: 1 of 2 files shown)


tools/release/rollback.sh
```diff
@@ -0,0 +1,68 @@
...
+# previous release name from the config_local.js file in Google Cloud
+# Storage. The desired rollback can also be specified as an argument.
+
+# The script supports two (optional) environment variables that can be
+# defined externally to modify its behavior:
+#
+# 1. "PROJECT_ID": Sets the name of the target project where the
+#    images will be pulled from and pushed to. Defaults to "cloud-datalab"
+# 2. "ROLLBACK_BUILD": The label of the build to pull down and rollback to.
+#    If omitted, this will extract the 'porevious' field from the config_local.js
```
ojarjur: s/porevious/previous/

tools/release/rollback.sh
```diff
@@ -0,0 +1,68 @@
...
+DATALAB_IMAGE="gcr.io/${PROJECT_ID}/datalab:local-${PREVIOUS_BUILD}"
+
+read -p "Proceed to release ${GATEWAY_IMAGE} and ${DATALAB_IMAGE} as latest? [Y/n]: " answer
+if echo $answer | grep -iq -v '^y'; then
+  exit 1
+fi
+
+echo "Moving config_local.js to config_local_${LATEST_BUILD}.js"
+echo "Updating config_local_${PREVIOUS_BUILD}.js to be config_local.js"
+gsutil cp gs://${PROJECT_ID}/deploy/config_local.js gs://${PROJECT_ID}/deploy/config_local_${LATEST_BUILD}.js
```
ojarjur: I think it is safe to drop this line and assume the `release.sh` script saved the config_local.js file with that timestamp.

yebrahim: Fair.

PR #1056, opened 2016-11-24
Add an allowOriginOverride setting to override the Access-Control-Allow-Origin header

corrieann:
Add an allowOriginOverride config setting to workaround Datalab's default behavior of deleting the Access-Control-Allow-Origin that is sent back from Jupyter.  When this setting is overridden with another frontend's URL, that frontend will be able to make XHR requests to the Datalab frontend.

PTAL @yebrahim @ojarjur @qimingj

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -425,15 +425,13 @@ function sendTemplate(key: string, data: common.Map<string>, response: http.Serv
 
 function responseHandler(proxyResponse: http.ClientResponse,
                          request: http.ServerRequest, response: http.ServerResponse) {
-  if (proxyResponse.headers['access-control-allow-origin'] !== undefined) {
-    if (appSettings.allowOriginOverride) {
-      proxyResponse.headers['access-control-allow-origin'] = appSettings.allowOriginOverride;
-    } else {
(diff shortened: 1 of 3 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -426,10 +426,14 @@ function sendTemplate(key: string, data: common.Map<string>, response: http.Serv
 function responseHandler(proxyResponse: http.ClientResponse,
                          request: http.ServerRequest, response: http.ServerResponse) {
   if (proxyResponse.headers['access-control-allow-origin'] !== undefined) {
-    // Delete the allow-origin = * header that is sent (likely as a result of a workaround
-    // notebook configuration to allow server-side websocket connections that are
-    // interpreted by Jupyter as cross-domain).
-    delete proxyResponse.headers['access-control-allow-origin'];
+    if (appSettings.allowOriginOverride) {
```
qimingj: I wonder if we should set the header outside the "if (proxyResponse.headers['access-control-allow-origin'] !== undefined)" block? Not sure if Jupyter always sets the header but in case it doesn't, should we set the overriding header anyway?

corrieann: Good point.  Done.

Issue #889, opened 2016-06-30
UX: Responsive design


VelizarVESSELINOV:
On small screen there are a lot of challenges (laptop with debugging window open use case):
1. The main action menu is not visible
2. A lot of space is used for nothing on the right side, it will be nice to be optional
3. There are big offset which is maybe good for big screens, but for small screen it is not very useful

![image](https://cloud.githubusercontent.com/assets/8209940/16494728/3ef8137c-3e9f-11e6-9e9a-40692ec178b9.png)

Jupyter online is more usable even in a smaller screen:
![image](https://cloud.githubusercontent.com/assets/8209940/16494526/85f6a384-3e9e-11e6-9c65-8002d3c91b08.png)

yebrahim:
Fixed by https://github.com/googledatalab/datalab/pull/1235.

PR #1059, opened 2016-11-29
Add 'Open as Text' Button in Toolbar

yebrahim:
This PR adds a button that appears when (exactly) one file is selected in the notebook list view, and opens a Jupyter editor for that file when clicked. Selecting no files, more than one file, or folders hides the button.

image

### DIFF
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -1241,10 +1241,11 @@ function initializeNotebookList(ipy, notebookList, newNotebook, events, dialog, 
     });
   }
 
+  // Extend the selection changed method to show/hide the editor button
   notebookSelectedFn = notebookList._selection_changed;
   notebookList._selection_changed = function() {
     notebookSelectedFn.apply(this, arguments);
-    if (notebookList.selected.length === 1)
+    if (notebookList.selected.length === 1 && notebookList.selected[0].type !== 'directory')
       $('#editorButton').show();
     else
       $('#editorButton').hide();
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -240,7 +240,7 @@ function initializePage(dialog, saveFn) {
         path = path.substr('/notebooks'.length);
         path = path.substr(0, path.lastIndexOf('/'));
       }
-      path = '/content' + path;
+      path = encodeURIComponent('/content' + path);
       prefix = window.location.protocol + '//' + window.location.host;
 
       window.open(prefix + '/_proxy/8083/#/repository?path=' + path);
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -240,7 +240,7 @@ function initializePage(dialog, saveFn) {
         path = path.substr('/notebooks'.length);
         path = path.substr(0, path.lastIndexOf('/'));
       }
-      path = encodeURIComponent('/content' + path);
+      path = '/content' + path;
       prefix = window.location.protocol + '//' + window.location.host;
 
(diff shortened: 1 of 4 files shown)


parthea:
This feature is really useful. Thanks for adding this! I've used the `http://127.0.0.1:8081/edit/...` trick before to edit files so it's great to have easy access to the text editor directly in Cloud Datalab.

I have a few observations that I want to share:

1. I notice the font color is different for the **Text Editor** button. I feel that the **Text Editor** button may feel out of place in its present state with the button border and the different font color. 
texteditorbutton

2. When there are unsaved changes in a notebook, I see the `'(unsaved changes)'` text beside the notebook name. I noticed that I don't see this behavior in the file editor when there are unsaved changes; although there is an asterisk symbol in the page title which helps. Separately, an autosave feature would also be very helpful for the **Text Editor** (I'm haven't played with this enough to see if it already exists).
**Notebook**
unsavedchanges2
**Text Editor**
texteditorunsavedchanges

3. In the **Text Editor**, clicking **File->Rename** did not seem to do anything in my environment.

4. In the **Text Editor**, clicking **File->Download** allowed me to download the file. This worked as expected on the first attempt. If I click **File->Download** again, I noticed that a new tab opens in my browser instead of downloading a file. Here is what I see in the navigation bar: http://127.0.0.1:8081/files/my_text_file.txt#?download=true

Separately, I have a question: Do we want to allow users to edit notebooks with running sessions? The reason I ask is that I notice that the **Rename** button is not visible when a notebook has a running session, but the **Text Editor** button is still visible.

I hope this feedback is helpful!

yebrahim:
For 1, sure that makes sense. I'll modify the looks to be in-line with the rest of the icons.
For 2, 3, and 4, please feel free to open issues for those and I'll see if I can get to them soon.

The user can editing notebooks with running sessions anyway, through another tab or through command line. It makes sense to make it harder to rename a notebook with a running session since this might be confusing if you have the session running with the old name. Not so much with editing the file itself, you've probably already got it open as a notebook so it's editable.

As for the toolbar real estate, you're right. We have another issue (#889) open for making the page's design responsive, in which case icons will collapse under a menu when the window isn't big enough for all of them.

Thanks @parthea, this feedback is really appreciated, please keep it coming.

sources/web/datalab/templates/tree.html
```diff
@@ -86,11 +86,14 @@
                             <span class="fa fa-files-o"></span> Copy
                           </button>
                           <button id="renameButton" type="button" class="toolbar-btn rename-button" title="Rename selected item(s)">
-                            <span class="fa fa-edit"></span> Rename
+                            <span class="fa fa-text-height"></span> Rename
```
ojarjur: I don't see how 'fa-text-height' indicates renaming.

How about, instead, keeping the 'fa-edit' icon here, and for the text editor use the 'file-text-o' icon. That way, we are indicating that button opens the file as a text file.

sources/web/datalab/static/datalab.js
```diff
@@ -240,7 +240,7 @@ function initializePage(dialog, saveFn) {
         path = path.substr('/notebooks'.length);
         path = path.substr(0, path.lastIndexOf('/'));
       }
-      path = encodeURIComponent('/content' + path);
+      path = '/content' + path;
```
ojarjur: Why did the call to `encodeURIComponent` get removed?

yebrahim: It's actually resulting in the wrong string in this case, where ungit expects an unencoded path.

ojarjur: That makes sense, but can you send it as a separate PR?

Otherwise, the reason behind this change will get lost.

yebrahim: Sure.

PR #1062, opened 2016-11-30
Automatically close auth windows loaded with "referer=popup"

corrieann:
Provide a way to automatically close an auth window in case Datalab's /signin flow is loaded from a popup window.  Instead of referer=URL-TO-REDIRECT-TO, issue referer=popup.

PTAL @yebrahim @ojarjur @qimingj

### DIFF
--- a/sources/web/datalab/auth.ts
+++ b/sources/web/datalab/auth.ts
@@ -196,7 +196,9 @@ function setOauth2Client(request: http.ServerRequest): void {
 
 function redirect(response: http.ServerResponse, referer: string) {
   if (referer == 'popup') {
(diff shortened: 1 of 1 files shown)


sources/web/datalab/auth.ts
```diff
@@ -194,6 +194,20 @@ function setOauth2Client(request: http.ServerRequest): void {
   }
 }
 
+function redirect(response: http.ServerResponse, referer: string) {
+  if (referer == 'popup') {
+    // If auth is triggered from a pop-up window, close the window.
```
qimingj: Assuming 'popup' is the state set by Colab? If so add that to comment so easier to understand and maintain?

PR #1065, opened 2016-12-02
Use gsutil's md5 hash to skip unnecessary backups

yebrahim:
This is related to automatic directory backup using GCS.

This PR changes the backup script to skip backing up directories when no changes are made. Without this, if we set up automatic backups every one hour for the last 10 hours for example, we end up with 10 copies of the same archive if no changes were made in the last 10 hours.


yebrahim:
It's worth noting that using `gsutil`'s md5 means that any changes to the files, including their metadata, will result in a different hash, and a new backup. So just by `touch`ing files we get a new hash. We should if this is good enough or if we need to only hash file contents.

PR #1070, opened 2016-12-08
Initial command line surface for a Datalab management tool

ojarjur:
This is meant to automate and simplify the manual steps currently documented
[here](https://cloud.google.com/datalab/docs/quickstarts/quickstart-gce-frontend).

This relies on the `gcloud` tool being installed on the same machine,
and the user already being authenticated.

### DIFF
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -81,7 +81,7 @@ def connection_flags(parser):
         dest='port',
         type=int,
         default=8081,
-        help='local port from which to forward traffic to Datalab')
+        help='local port on which Datalab is accessible')
     parser.add_argument(
         '--max-reconnects',
         dest='max_reconnects',
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -111,19 +111,12 @@ def connection_flags(parser):
             'the log level for the SSH command.'
             '\n\n'
             'The default log level is "error".'))
-
-    browser_group = parser.add_mutually_exclusive_group()
-    browser_group.add_argument(
+    parser.add_argument(
         '--no-launch-browser',
-        dest='launch_browser',
-        action='store_false',
-        help='do not open a browser connected to Datalab')
-    browser_group.add_argument(
-        '--launch-browser',
-        dest='launch_browser',
+        dest='no_launch_browser',
         action='store_true',
-        default=True,
-        help='open a browser connected to Datalab')
+        default=False,
+        help='do not open a browser connected to Datalab')
 
     return
 
@@ -204,7 +197,7 @@ def connect(args, gcloud_compute):
     def on_ready():
         """Callback that handles a successful connection."""
         print('You can now connect to Datalab at ' + datalab_address)
-        if args.launch_browser:
+        if not args.no_launch_browser:
             try:
                 webbrowser.open(datalab_address)
             except webbrowser.Error as e:
@@ -273,7 +266,7 @@ def connect(args, gcloud_compute):
         print('Attempting to reconnect...')
         remaining_reconnects -= 1
         # Don't launch the browser on reconnect...
-        args.launch_browser = False
+        args.no_launch_browser = True
     return
 
 
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -14,6 +14,8 @@
 
 """Methods for implementing the `datalab connect` command."""
 
+import os
+import os.path
 import subprocess
 import tempfile
 import threading
@@ -89,6 +91,26 @@ def connection_flags(parser):
             'maximum number of times to reconnect.'
             '\n\n'
             'A negative value means no limit.'))
+    parser.add_argument(
+        '--log-file',
+        dest='log_file',
+        default=os.devnull,
+        help=(
+            'log file to write messages from the SSH command.'
+            '\n\n'
+            'This may be useful for debugging issues with an SSH connection.'
+            '\n\n'
+            'By default, all such messages are discarded.'))
+    parser.add_argument(
+        '--log-level',
+        dest='log_level',
+        choices=['quiet', 'fatal', 'error', 'info', 'verbose',
+                 'debug', 'debug1', 'debug2', 'debug3'],
+        default='error',
+        help=(
+            'the log level for the SSH command.'
+            '\n\n'
+            'The default log level is "error".'))
 
     browser_group = parser.add_mutually_exclusive_group()
     browser_group.add_argument(
@@ -118,6 +140,39 @@ def connect(args, gcloud_compute):
 
     datalab_address = 'http://localhost:{0}/'.format(str(args.port))
 
+    def ensure_sshable():
+        """Ensure that the instance has been configured for SSH access.
+
+        This method is separate from the `create_tunnel` method, as the
+        first time we SSH in to an instance may involve interactive
+        prompts that the user wants to interact with, even if they
+        otherwise want the output from an SSH command to go to a log file.
+
+        Raises:
+          KeyboardInterrupt: When the end user kills the connection
+        """
+        print('Ensuring that {0} can be connected to via SSH').format(instance)
+
+        cmd = ['ssh']
+        if args.quiet:
+            cmd.append('--quiet')
+        if args.zone:
+            cmd.extend(['--zone', args.zone])
+        cmd.extend([
+            '--ssh-flag=-4',
+            '--ssh-flag=-o',
+            '--ssh-flag=LogLevel=' + args.log_level])
+        cmd.append('datalab@{0}'.format(instance))
+        cmd.extend(['--', 'true'])
+        with open(os.devnull, "a") as dn:
+            while True:
+                try:
+                    gcloud_compute(args, cmd, stderr=dn)
+                    return
+                except subprocess.CalledProcessError:
+                    pass
+        return
+
     def create_tunnel():
         """Create an SSH tunnel to the Datalab instance.
 
@@ -136,13 +191,14 @@ def connect(args, gcloud_compute):
         cmd.extend([
             '--ssh-flag=-4',
             '--ssh-flag=-o',
-            '--ssh-flag=LogLevel=error',
+            '--ssh-flag=LogLevel=' + args.log_level,
             '--ssh-flag=-N',
             '--ssh-flag=-L',
             '--ssh-flag=' + port_mapping])
         cmd.append('datalab@{0}'.format(instance))
-        with tempfile.TemporaryFile() as tf:
-            gcloud_compute(args, cmd, stdout=tf, stderr=tf)
+        log_file = os.path.expandvars(os.path.expanduser(args.log_file))
+        with open(log_file, "a") as lf:
+            gcloud_compute(args, cmd, stdout=lf, stderr=lf)
         return
 
     def on_ready():
@@ -205,6 +261,7 @@ def connect(args, gcloud_compute):
             health_check_thread.join()
         return
 
+    ensure_sshable()
     remaining_reconnects = args.max_reconnects
     while True:
         try:
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -14,7 +14,6 @@
 
 """Methods for implementing the `datalab connect` command."""
 
-import argparse
 import subprocess
 import tempfile
 import threading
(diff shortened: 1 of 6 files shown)


tools/cli/commands/connect.py
```diff
@@ -0,0 +1,263 @@
...
+        get_cmd.extend(['--zone', args.zone])
+        start_cmd.extend(['--zone', args.zone])
+    status = 'UNKNOWN'
+    with tempfile.TemporaryFile() as tf:
+        gcloud_compute(args, get_cmd + ['--format', 'value(status)', instance], stdout=tf)
+        tf.seek(0)
+        status = tf.read().strip()
+    if status != 'RUNNING':
+        print('Restarting the instance {0} with status {1}'.format(instance, status))
+        gcloud_compute(args, start_cmd + [instance])
```
ojarjur: The `gcloud_compute` method raises a subprocess.CalledProcessError exception if the exit code is anything other than 0.

We handle that in the outermost `run` method.

tools/cli/commands/connect.py
```diff
@@ -0,0 +1,263 @@
...
+gets dropped. However, that connection will only exist while
+this command is running.
+""")
+
+
+examples=(
+"""
+To connect to 'example-instance' in zone 'us-central1-a', run:
+
+    $ {0} {1} example-instance --zone us-central1-a"""
```
nikhilk: I wonder if its interesting to consider a single arg:
[[project.]zone.]instance

So fully-qualified name is xyzproject.us-central1-a.example-instance, and one that picks up the current configured project/zone is simply example-instance.

tools/cli/commands/connect.py
```diff
@@ -0,0 +1,263 @@
...
+
+def flags(parser):
+    """Add command line flags for the `connect` subcommand.
+
+    Args:
+      parser: The argparse parser to which to add the flags.
+    """
+    parser.add_argument(
+        'instance',
+        metavar='INSTANCE',
```
nikhilk: I think metavar should be 'name' -- as mentioned in the help ... 'name of ...'

tools/cli/commands/connect.py
```diff
@@ -0,0 +1,263 @@
...
+        '--quiet',
+        dest='quiet',
+        action='store_true',
+        help='do not issue any interactive prompts')
+    parser.add_argument(
+        '--port',
+        dest='port',
+        type=int,
+        default=8081,
+        help='local port from which to forward traffic to Datalab')
```
nikhilk: Simplify the help to say "the local port on which Datalab is accessible"

tools/cli/commands/connect.py
```diff
@@ -0,0 +1,263 @@
...
+        '--max-reconnects',
+        dest='max_reconnects',
+        type=int,
+        default=-1,
+        help=(
+            'maximum number of times to reconnect.'
+            '\n\n'
+            'A negative value means no limit.'))
+
+    browser_group = parser.add_mutually_exclusive_group()
```
nikhilk: A single flag is sufficient. The default should be to launch a browser, so all that is needed is --no-browser.

tools/cli/commands/connect.py
```diff
@@ -0,0 +1,263 @@
...
+        cmd.extend([
+            '--ssh-flag=-4',
+            '--ssh-flag=-o',
+            '--ssh-flag=LogLevel=error',
+            '--ssh-flag=-N',
+            '--ssh-flag=-L',
+            '--ssh-flag=' + port_mapping])
+        cmd.append('datalab@{0}'.format(instance))
+        with tempfile.TemporaryFile() as tf:
+            gcloud_compute(args, cmd, stdout=tf, stderr=tf)
```
nikhilk: Maybe interesting to have a --log file arg that someone can use if they are having issues, and in that case have a verbose output go to that log file. Otherwise, I think tmp file (isn't there a way to use /dev/null instead?) should work just fine.

tools/cli/commands/create.py
```diff
@@ -0,0 +1,375 @@
...
+_DATALAB_NETWORK_DESCRIPTION = 'Network for Google Cloud Datalab instances'
+
+_DATALAB_FIREWALL_RULE = 'datalab-network-allow-ssh'
+_DATALAB_FIREWALL_RULE_DESCRIPTION = 'Allow SSH access to Datalab instances'
+
+_DATALAB_DEFAULT_DISK_SIZE_GB = 200
+_DATALAB_DISK_DESCRIPTION = (
+    'Persistent disk for a Google Cloud Datalab instance')
+
+_DATALAB_STARTUP_SCRIPT = """#!/bin/bash
```
nikhilk: Does it make sense to write this script file as a .sh and package it as a file in the tool, rather than embed in the code.

[not for now] but think this is an interesting user customization point - to be able to provide a modified startup script for the VM.

tools/cli/commands/create.py
```diff
@@ -0,0 +1,375 @@
...
+  mkdir -p "${MOUNT_DIR}"
+  ${MOUNT_CMD} || format_disk
+  mount -o discard,defaults /dev/disk/by-id/google-datalab-pd /mnt/disks/datalab-pd || format_disk
+  chmod a+w "${MOUNT_DIR}"
+}
+
+mount_disk
+"""
+
+_DATALAB_CONTAINER_SPEC = """
```
nikhilk: Similar comment about making this an external file, but still packaged within the tool.

tools/cli/commands/create.py
```diff
@@ -0,0 +1,375 @@
...
+            'If not specified, this defaults to having a name based\n'
+            'on the instance name.'))
+    parser.add_argument(
+        '--disk_size_gb',
+        type=int,
+        dest='disk_size_gb',
+        default=_DATALAB_DEFAULT_DISK_SIZE_GB,
+        help='size of the persistent disk in GB.')
+
+    connect_group = parser.add_mutually_exclusive_group()
```
nikhilk: Similar comment about pick the default, and offer a single flag for opposite behavior.

tools/cli/commands/delete.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Args:
+      parser: The argparse parser to which to add the flags.
+    """
+    parser.add_argument(
+        'instance',
+        metavar='INSTANCE',
+        help='name of the instance to delete')
+
+    auto_delete_override = parser.add_mutually_exclusive_group()
```
nikhilk: Same -- single flag.

tools/cli/commands/list.py
```diff
@@ -0,0 +1,111 @@
...
+
+examples=("""
+To list all of the available Datalab instances in a project:
+
+    $ {0} {1}
+
+To only list the Datalab instances in the zones
+'us-central1-a' and  'us-central1-b':
+
+    $ {0} {1} --zones us-central1-a,us-central1-b
```
nikhilk: So I personally like my single fully qualified name pattern, and if you agree, I think this would just be either:

xyzproject.us-central1-a to list all in a zone or
xyzproject to list all in a project and configured default zone
or nothing to list all in configured project and zone

So just:

    datalab list
    datalab list xyzproject.us-central1-a

Thoughts?

nikhilk:
Looks great -- I had comments, mostly on the CLI surface of the tool.

On that question of installation - we will get this to be installed with the sdk, so I don't think this needs to be pip installable by itself.

tools/cli/commands/delete.py
ojarjur: This one is a bit different from the others, as a user might have to override to get back the default behavior.

Specifically, they could create an instance, and then (via the console) change the attached disk to be deleted. Alternatively, we may make the disk keep/delete config a flag in the create command.

In those scenarios, if they want the `datalab delete` command to not delete the disk, they need a flag to get back to the default behavior.

yebrahim: I think generally having more flags helps even if they're redundant, users won't always remember what the default behavior is.

tools/cli/commands/create.py
```diff
@@ -0,0 +1,379 @@
...
+        dest='connect',
+        action='store_false',
+        help='do not connect to the newly created instance')
+    connect_group.add_argument(
+        '--connect',
+        dest='connect',
+        default=True,
+        action='store_true',
+        help='connect to the newly created instance')
+
```
yebrahim: It'd be great to have a --image option here to create custom images

tools/cli/commands/connect.py
```diff
@@ -0,0 +1,264 @@
...
+        get_cmd.extend(['--zone', args.zone])
+        start_cmd.extend(['--zone', args.zone])
+    get_cmd.extend(['--format', 'value(status)', instance])
+    start_cmd.extend([instance])
+    status = 'UNKNOWN'
+    with tempfile.TemporaryFile() as tf:
+        gcloud_compute(args, get_cmd, stdout=tf)
+        tf.seek(0)
+        status = tf.read().strip()
+    if status != 'RUNNING':
```
yebrahim: What if a non-existent instance name is supplied here (e.g. by mistake)? Does this result in forever trying to connect?

ojarjur: > Does this result in forever trying to connect?

No, it doesn't get to the point of comparing the status.

What happens is the call to `gcloud_compute` on line 242 raises a subprocess.CalledProcessError exception, and that cascades into the `run` method in the datalab.py file, which then reports that to the end user.

tools/cli/commands/connect.py
ojarjur: I've added a '--log-file' flag

tools/cli/commands/connect.py
```diff
@@ -0,0 +1,263 @@
...
+        Args:
+          cancelled_event: A threading.Event instance that indicates we should
+            give up on the instance becoming reachable.
+        """
+        health_url = '{0}_info/'.format(datalab_address)
+        healthy = False
+        print('Waiting for Datalab to be reachable at ' + datalab_address)
+        while not cancelled_event.is_set():
+            try:
+                health_resp = urllib2.urlopen(health_url)
```
ojarjur: I've thought about it, and I don't think it would add much.

This command always keeps running until the user explicitly kills it, so I can't think of any benefit a user would get from a timeout.

tools/cli/commands/connect.py
ojarjur: Done.

tools/cli/commands/connect.py
ojarjur: Done

tools/cli/commands/connect.py
ojarjur: I'm worried about drifting too far from the conventions of the existing command line tools.

As far as I am aware, none of the existing tools (e.g. gcloud, gsutil, etc) support referring to instances in this way.

tools/cli/commands/list.py
ojarjur: This makes me nervous because it is so different from the existing tools.

In particular, I don't think any of the existing tools support a single operation crossing multiple projects.

For instance, the 'gsutil ls' command only shows you buckets in a single project.

ojarjur:
@nikhilk I believe I got to all of your comments.

PTAL

tools/cli/commands/create.py
ojarjur: Done

tools/cli/commands/connect.py
nikhilk: OK, fair enough -- except BQ does use something like that.

tools/cli/commands/list.py
nikhilk: To be clear, I didn't suggest spanning multiple projects/zones. Either use default project and zone, or use specified project and default zone, or use specified project and specified zone.

But if we're not doing a compact syntax, then I suppose this is moot.

tools/cli/commands/create.py
ojarjur: Added https://github.com/googledatalab/datalab/issues/1088 for this

tools/cli/commands/create.py
ojarjur: This is also covered by https://github.com/googledatalab/datalab/issues/1088

tools/cli/commands/list.py
ojarjur: Fair enough; I misread your earlier comment. Sorry about that.

tools/cli/commands/connect.py
ojarjur: I didn't realize bq supports that. I'm still not sure whether or not it makes sense for us, but we should at least track it.

I added https://github.com/googledatalab/datalab/issues/1089 to do that.

PR #1072, opened 2016-12-09
Set the root path for ungit

ojarjur:
The reverse proxy logic will route requests correctly if either:

1. Their path specifies a resource behind the reverse proxy, or...
2. The referrer specifies a path behind the reverse proxy.

However, for the ungit UI sometimes neither of these were true.

This change attempts to fix that by making the URLs used by ungit always specify a reverse-proxy path.

### DIFF
--- a/containers/datalab/content/run.sh
+++ b/containers/datalab/content/run.sh
@@ -205,7 +205,7 @@ then
 fi
 
 # Start the ungit server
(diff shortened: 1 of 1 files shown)


ojarjur:
Scratch that; this PR is not the fix for the issue I am seeing (sporadic request failures due to content-length mismatch).

Issue #1073, opened 2016-12-10
Datalab erroneously redirecting to '/login'


ojarjur:
We found this during our attempt at releasing today and had to rollback the release because of it.

Bradley is investigating to identify the root cause.

qimingj:
It seems not related to any check-ins. the daily builds before (including) 12/8 work well. 12/9 build is screwed and we don't have checkins in datalab (and one unrelated checkin in pydatalab).

It seems more likely that apt-get update or some packages that we do not specify version got updated within 24 hours and datalab docker build picked it up. This is nasty and I have to dig further.

qimingj:
Found the diff between 12/8 and 12/9 image. On 12/8, notebook = 4.2.3. On 12/9, notebook = 4.3.0. From log:
...
Collecting notebook>=4.2.0 (from widgetsnbextension>=1.2.6->ipywidgets==5.2.2)
  Downloading notebook-4.3.0-py2.py3-none-any.whl (6.9MB)
...
We probably need to pin notebook = 4.2.3.

yebrahim:
It already is. It's getting updated by ipywidgets. We should find an ipywidgets version that doesn't require updating notebook.

yebrahim:
Looks like this PR is what broke it: https://github.com/jupyter/notebook/pull/1831. I'll see if it's an easy fix.

qimingj:
Thanks for chasing it down!

PR #1075, opened 2016-12-10
Pin notebook version to 4.2.3. 4.3.0 seems not working well with Datalab

qimingj:
Need more investigation on how to tweak notebook 4.3.0 to work with Datalab.

This fixes https://github.com/googledatalab/datalab/issues/1073.

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -138,6 +138,10 @@ RUN gsutil cp gs://cloud-ml/sdk/cloudml-0.1.6-alpha.tar.gz cloudml-0.1.6-alpha.t
     pip install --no-cache-dir cloudml-0.1.6-alpha.tar.gz && \
     rm cloudml-0.1.6-alpha.tar.gz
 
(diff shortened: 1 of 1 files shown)


qimingj:
ipwidgets should work as it does before. The problem is ipywidgets depend on notebook>=4.2.0, and sometime yesterday notebook 4.3.0 was published on pypi. In Datalab ipywidgets has been working with 4.2.* for a while.

The cause is notebook 4.3.0 includes a breaking change that we need to investigate where.

yebrahim:
You're right for this case, but generally I think we should pin all the dependencies, rather than hope that `notebook==4.3` won't break `ipywidgets==5.2.2`.

As an alternative, we can just `ipywidget`'s dependencies at working versions. Checking `ipywidget`'s dependencies, we're only missing `widgetsnbextension==1.2.6`. If we pip install this one we can then `pip install notebook` and it shouldn't update.

qimingj:
yebrahim has a better fix #1076 that takes care of this.

PR #1077, opened 2016-12-10
Update node to v6.9.2

yebrahim:
It's good to bring our packages up to date, but this is needed now to fix `npm install` issues in Docker on MacOS, where it hangs forever or throws errors with older node versions. This happens if `npm install` is run in a mounted directory.

Context:
https://forums.docker.com/t/unable-to-perform-any-npm-install-inside-a-osxfs-mounted-directory/12590/17
https://forums.docker.com/t/npm-install-doesnt-complete-inside-docker-container/12640
https://github.com/npm/npm/issues/9863

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -80,7 +80,7 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
 
 # Setup Node.js
     mkdir -p /tools/node && \
-    wget -nv https://nodejs.org/dist/v4.3.2/node-v4.3.2-linux-x64.tar.gz -O node.tar.gz && \
+    wget -nv https://nodejs.org/dist/v6.9.2/node-v6.9.2-linux-x64.tar.gz -O node.tar.gz && \
     tar xzf node.tar.gz -C /tools/node --strip-components=1 && \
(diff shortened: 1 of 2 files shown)


containers/base/Dockerfile
```diff
@@ -80,7 +80,7 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
 
 # Setup Node.js
```
nikhilk: Probably good to clearly indicate we are using the LTS build (rather than cutting edge/latest).

yebrahim: Well we're only choosing to pin to a version that works, do we have any other reason? It's just like any other installation in the Dockerfile.

parthea: I'm new to the Datalab development environment so my personal preference is to have more comments in the Datalab Dockerfile (and potentially external links in the comments to obtain more information). The following notes could be helpful for maintenance purposes:
- `Use the Node LTS build for stability reasons. See `https://nodejs.org/en/download/
- `Maintenance end date for Node 6.x LTS is 2019-04-18. See ` https://github.com/nodejs/LTS/

yebrahim:
Closing this for now as unnecessary, will reopen when we get a development Docker image.

PR #1087, opened 2016-12-14
Make Docker image more dev-friendly

yebrahim:
This PR makes the development cycle easier by building the Datalab frontend inside the Docker container, and adding an option to mount the `.ts` sources from host disk inside the container so they can be iterated on.

- Typescript compiler is now installed inside the image, reducing build requirements on host to be only Docker
- Sources are now compiled as part of building the container, and every time the `content/run.sh` script is executed.
- Removed build script from frontend sources. These should no longer be built outside the container.
- Added an event listener to kill the frontend on `SIGINT` to allow reiteration.

The following development cycle is now possible:

- Build the Datalab image: `./containers/datalab/build.sh`
- Mount the sources inside the run script (see [here](https://github.com/googledatalab/datalab/compare/master...yebrahim:yebrahim-devimage?expand=1#diff-db9db80a3ffb0b3c57d84ed6fa28f6f4R47))
- Run a container and open a shell into it: `./containers/datalab/run.sh shell`
- Inside the container, change to `/datalab`, and run the server from there: `./run.sh`
- Change `.ts` sources on the host disk, kill the server inside the container and restart it


nikhilk:
On surface this doesn't make sense -- lets sync up...

yebrahim:
Sync'ed. We're not proceeding with this as it will change the Datalab container environment. We'll explore adding an additional docker image for development cycle.

PR #1090, opened 2016-12-19
Use supervisord instead of forever to keep datalab servers running

yebrahim:
We're switching to supervisor to be able to spawn multiple processes inside the docker container.


containers/gateway/content/run.sh
```diff
@@ -35,11 +35,13 @@ n=0; while true; do
...
   echo "Starting Datalab Kernel Gateway in silent mode, for debug output, rerun with an additional '-e DATALAB_DEBUG=true' argument"
-  FOREVER_CMD="${FOREVER_CMD} -s"
+else
+  # The Datalab job logs to this file, make sure it exists and tail it
+  touch /var/log/gateway.stdout.log
+  tail -f /var/log/gateway.stdout.log &
 fi
 
-${FOREVER_CMD} /datalab/web/kernelproxy/app.js
+supervisord -c /etc/supervisor/conf.d/gateway.conf
```
qimingj: Can we do the same for jupyter command in the same file?

containers/base/Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
     apt-get update -y && \
     apt-get install --no-install-recommends -y -q python unzip ca-certificates build-essential \
     libatlas-base-dev liblapack-dev gfortran libpng-dev libfreetype6-dev libxft-dev libxml2-dev \
-    python-dev python-setuptools python-zmq openssh-client wget curl git pkg-config && \
+    python-dev python-setuptools python-zmq openssh-client wget curl git pkg-config supervisor && \
```
qimingj: Not sure if we need to do anything given the license here https://github.com/Supervisor/supervisor/blob/master/LICENSES.txt?

yebrahim: Need @Di-Ku input on this

tools/supervisor/datalab.conf
```diff
@@ -0,0 +1,10 @@
+[supervisord]
```
nikhilk: I think this file should go into the containers directory.
Assumption is that this is only going to be used within the docker container, not outside.

tools/supervisor/gateway.conf
```diff
@@ -0,0 +1,10 @@
+[supervisord]
```
nikhilk: And this one too.

containers/base/build.sh
```diff
@@ -18,12 +18,17 @@
...
+# Copy local pydatalab if supplied
 if [ -n "$1" ]; then
   rsync -avp "$1" pydatalab;
 else
   # Create empty dir to make docker build happy.
   mkdir -p pydatalab;
 fi
+
+# Copy supervisord files
+rsync -avp ../supervisor/ supervisor
```
ojarjur: Adding more rsync calls to the build script makes me think the files are in the wrong place.

I wouldn't put any supervisor configs in the base image, but rather put the supervisor config for the Datalab frontend in the 'datalab' image and the other configs in the 'gateway' image.

That, in turn, means we would just have those files in the 'containers/datalab/content' and 'containers/gateway/content' directories.

nikhilk: I agree - I also wasn't sure why all these rsync calls. Would be simpler to have them in the right place to begin with.

yebrahim:
Closing this for now until we sort out the licensing issues.

PR #1094, opened 2016-12-21
Run GCSBackup.sh script periodically from the nodejs server

yebrahim:
This PR adds periodic GCS backup functionality into the Datalab frontend:
- Added a backupUtility.ts module that runs as part of the server initialization, and periodically calls the GCSBackup.sh script.
- Added optional arguments to GCSBackup.sh script to specify machine id and project id to be able to test it locally (not within a GCE VM).
- Added config settings to turn on/off GCS backup, and specify number of hourly, daily, and weekly backups to keep on GCS if the feature is on.
- Removed README from GCSBackup.sh script as it's not meant to be run by the user.


sources/web/datalab/backupUtility.ts
```diff
@@ -0,0 +1,100 @@
...
+  return fs.writeFileSync(backup_log_path, JSON.stringify(log_history), 'utf8');
+}
+
+/**
+ * Run the GCSbackup shell script and call the callback function with its exit code
+ */
+function runBackup(tag: string, callback: Function) {
+  var cmd = childProcess.spawn('/bin/bash', ['/datalab/GCSbackup.sh',
+                               '-t', tag,
+                               '-p', '/content/datalab',
```
ojarjur: I think this should be the entire '/content/' directory.

sources/web/datalab/backupUtility.ts
```diff
@@ -0,0 +1,100 @@
...
+      var log_history = readBackupLog();
+      // test hourly backup
+      if (!log_history.lastHourlyRun || (Date.now()-log_history.lastHourlyRun)/1000 > 60*60) {
+        runBackup('hourly', () => {
+          writeBackupLog(log_history);
+          log_history.lastHourlyRun = Date.now();
+        });
+      }
+      if (!log_history.lastDailyRun || (Date.now()-log_history.lastDailyRun)/1000 > 60*60*24) {
+        runBackup('daily', () => {
```
ojarjur: I'm surprised to see this running a new backup instead of using the most recent hourly backup.

sources/web/datalab/backupUtility.ts
```diff
@@ -0,0 +1,100 @@
...
+          writeBackupLog(log_history);
+          log_history.lastDailyRun = Date.now();
+        });
+      }
+      if (!log_history.lastWeeklyRun || (Date.now()-log_history.lastWeeklyRun)/1000 > 60*60*24*7) {
+        runBackup('weekly', () => {
```
ojarjur: Same comment here about running a new backup instead of reusing an existing one.

containers/datalab/content/GCSbackup.sh
```diff
@@ -62,6 +62,14 @@ while [[ $# -gt 1 ]]; do
         log_file="$2"
         shift
         ;;
+      -m)
+        machine_id="$2"
+        shift
+        ;;
+      -P)
```
yebrahim: Sure. Btw, I added this as for testing, as it should be detected automatically on GCE VMs. I added a comment explaining this.

sources/web/datalab/backupUtility.ts
yebrahim: Agree. Forgot to replace after testing.

sources/web/datalab/backupUtility.ts
yebrahim: Here's my reasoning why I think all three should be separate:
- Simplifies the implementation since each run only has to look for previous runs with the same tag.
- If one backup fails it doesn't affect the others.
- If backups are not complete because VM was stopped, these backups can go out of sync, and that should be fine.

tools/gcsbackup/README.md
```diff
@@ -1,19 +0,0 @@
-# GCSbackup
```
yebrahim: Where do you think it should go? Note that I moved the GCSBackup script into the content directory, which has several unrelated scripts as of now.

ojarjur: My bad, I somehow overlooked that the Bash script was converted to a TS and moved into the main codebase.

sources/web/datalab/common.d.ts
```diff
@@ -87,6 +87,26 @@ declare module common {
      * If true, allow HTTP requests via websockets.
      */
     allowHttpOverWebsocket: boolean;
+    
+    /**
+     * Whether to automatically back up user's contents dir to GCS
+     */
+    enableAutoGCSBackups: boolean;
```
ojarjur: By putting these inside of the settings.json file, they get tied to the source code and a user won't be able to override them.

Instead, these should go into the user settings and be read using the ['loadUserSettings'](https://github.com/googledatalab/datalab/blob/master/sources/web/datalab/settings.ts#L90) method.

yebrahim: Makes sense.

However I think these should still be part of the global settings, with the userSettings optionally overriding them. This way we can keep our defaults in one place, and the user starts out with an empty settings file that they modify however they need. We then modify the loadUserSettings function to first try loading from the user's config dir, then default to the global settings file. Thoughts?

yebrahim:
I think with https://github.com/googledatalab/datalab/pull/1100 the settings issue is resolved, is this ready to be merged?

sources/web/datalab/backupUtility.ts
```diff
@@ -0,0 +1,100 @@
...
+import logging = require('./logging');
+import path = require('path');
+import fs = require('fs');
+
+var appSettings: common.Settings;
+var backupIntervalVar: number = null;
+var backupLogPath: string = '';
+
+// check for backups every 10 minutes so we can retry failed backups
+const BACKUP_INTERVAL: number = 10*60*60;
```
ojarjur: This is not every 10 minutes.

(I believe) that would be 10*60*1000

ojarjur:
@yebrahim I think there is still more work to do on the settings, but that shouldn't block the rest of this PR

(in particular, there is currently no way to disable the backups if you use the CLI tool to create a Datalab instance).

Let's change the default to having backups disabled, and then we can get this submitted and flip it back later once we have a UI to change that setting.

yebrahim:
Rebased and fixed.

PR #1098, opened 2016-12-28
Use dot:mega progress meter option for wget

neilsh:
In the Dockerfile, the default `--progress=dot` style used when wget has non-TTY output uses 1K per dot, 50 dots (50K) per line. This ends up being quite noisy when downloading the 86MB gcc source file, so I've added the `--progress=mega` option which uses 64K per dot instead, which is 3M per line. This allows the progress to fit onto a single screen, instead of many screens.

I also considered using `--progress=giga`, but on slower connections it might be unacceptably long before the first indication of download speed appears.

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -35,7 +35,7 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
     mkdir -p /srcs && \
     cd /srcs && \
     apt-get source -d wget git python-zmq ca-certificates pkg-config libpng-dev && \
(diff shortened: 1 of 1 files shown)


neilsh:
I signed it!

yebrahim:
Thanks! This was annoying indeed. Can we add it to other `wget` commands as well?

neilsh:
Sure, I added it to the tensorflow wget. I didn't add it to the other two because they already have the `-nv` option, which prevents the progress bar from rendering at all.

yebrahim:
Great, thanks.

PR #1100, opened 2017-01-03
Allow overriding individual config settings from outside docker container

blois:
This allows patching specific config settings from the environment so the settings can be tweaked without modifying the docker container.


blois:
Updated with latest changes.

Thanks!

PR #1101, opened 2017-01-03
Add a Travis config

ojarjur:
This does two things:

1. Ensures that the 'datalab' container can be built
2. Looks for lint errors in the CLI

This change also includes two fixes to lint errors in the CLI that
were not caught earlier (when we did not have this check).


yebrahim:
Thanks! This is a good start.

PR #1105, opened 2017-01-06
Cleanups for GET request handlers:

ojarjur:
1. Make the '_restart' handler require a 'POST' request to do anything.
2. Make the '_setting' handler require a 'POST' request in order to
   update a setting. This also meant it should support reading the
   form data from either the URL query or from the request body.

Similarly, the various places where we update settings were changed
to use 'POST' requests instead of 'GET' requests.

### DIFF
--- a/sources/web/datalab/server.ts
+++ b/sources/web/datalab/server.ts
@@ -175,6 +175,9 @@ function handleRequest(request: http.ServerRequest,
   //       this into a real feature, that involves a confirmation prompt, as
   //       well validation to require a POST request.
   if (path.indexOf('/_restart') == 0) {
+    if ('POST' != request.method) {
+      return;
+    }
     setTimeout(function() { process.exit(0); }, 0);
     response.statusCode = 200;
     response.end();
--- a/sources/web/datalab/settings.ts
+++ b/sources/web/datalab/settings.ts
@@ -219,7 +219,6 @@ function postSettingsHandler(userId: string, request: http.ServerRequest, respon
     }
     formHandler(userId, formData, request, response);
   });
-  request.resume();
 }
 
 /**
--- a/sources/web/datalab/settings.ts
+++ b/sources/web/datalab/settings.ts
@@ -229,6 +229,8 @@ function postSettingsHandler(userId: string, request: http.ServerRequest, respon
  */
 function formHandler(userId: string, formData: any, request: http.ServerRequest, response: http.ServerResponse): void {
   if (!(('key' in formData) && ('value' in formData))) {
+    response.writeHead(400, { 'Content-Type': 'text/plain' });
+    response.end('Missing one or more required fields');
     return;
   }
   var key = formData['key'];
--- a/sources/web/datalab/settings.ts
+++ b/sources/web/datalab/settings.ts
@@ -21,6 +21,7 @@ import fs = require('fs');
 import http = require('http');
 import uuid = require('node-uuid');
 import path = require('path');
+import querystring = require('querystring');
 import url = require('url');
 import util = require('util');
 import userManager = require('./userManager');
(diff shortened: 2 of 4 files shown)


sources/web/datalab/server.ts
```diff
@@ -175,6 +175,9 @@ function handleRequest(request: http.ServerRequest,
   //       this into a real feature, that involves a confirmation prompt, as
   //       well validation to require a POST request.
   if (path.indexOf('/_restart') == 0) {
+    if ('POST' != request.method) {
```
ojarjur: Since this now requires a post request, we should also add a button in the app bar to send such a request (it's not reasonable to require the user to manually send one).

ojarjur: Done; I added a reset button to the app bar.

sources/web/datalab/settings.ts
```diff
@@ -168,39 +169,82 @@ export function updateUserSetting(userId: string, key: string, value: string, as
...
     } else {
-      response.writeHead(200, { 'Content-Type': 'application/json' });
-      response.end(JSON.stringify(userSettings));
-      return;
+      var parsedUrl = url.parse(request.url, true);
+      formData = parsedUrl.query;  
+    }
+    formHandler(userId, formData, request, response);
+  });
+  request.resume();
```
qimingj: Not sure if the resume() call is needed? It seems if you handle 'data' event you may not.

sources/web/datalab/settings.ts
```diff
@@ -168,39 +169,82 @@ export function updateUserSetting(userId: string, key: string, value: string, as
...
+
+/**
+ * Handles updating a setting given the parsed form contents.
+ * @param formData the form data parsed from the incoming http request.
+ * @param request the incoming http request.
+ * @param response the outgoing http response.
+ */
+function formHandler(userId: string, formData: any, request: http.ServerRequest, response: http.ServerResponse): void {
+  if (!(('key' in formData) && ('value' in formData))) {
+    return;
```
yebrahim: Return an error instead of dropping the request?

sources/web/datalab/static/datalab.js
```diff
@@ -180,7 +194,8 @@ function initializePage(dialog, saveFn) {
       '<span class="fa fa-external-link-square">&nbsp;</span><a href="/static/about.txt" target="_blank">License and software information</a><br />' +
       '<span class="fa fa-external-link-square">&nbsp;</span><a href="https://cloud.google.com/terms/" target="_blank">Terms of Service</a><br />' +
       '<span class="fa fa-external-link-square">&nbsp;</span><a href="http://www.google.com/intl/en/policies/" target="_blank">Privacy Policy</a><br />' +
-      '<span class="fa fa-external-link-square">&nbsp;</span><a href="/static/reporting.html?enabled=' + reportingEnabled + '" target="_blank">Usage Statistics</a><br />';
+      '<span class="fa fa-external-link-square">&nbsp;</span><a href="/static/reporting.html?enabled=' + reportingEnabled + '" target="_blank">Usage Statistics</a><br />' +
+      '<span class="fa fa-recycle">&nbsp;</span><a href="javascript:restartDatalab()">Restart Server</a><br />';
```
ojarjur: I'm hesitant to call it that, since it does not restart all of Datalab (e.g. the Docker container).

Instead, this just kills the NodeJS server, which then gets restarted by forever.

So, for instance, custom setup scripts will not re-run, nor will will it recreate things like SSH connections to a kernel gateway.

sources/web/datalab/static/datalab.js
```diff
@@ -157,14 +157,28 @@ function showHelp(markup) {
 function xhr(url, callback) {
   let request = new XMLHttpRequest();
   request.onreadystatechange = callback.bind(request);
-  request.open("GET", url);
+  request.open("POST", url);
```
ojarjur: Done

sources/web/datalab/settings.ts
ojarjur: Apparently, it isn't.

I've removed that call.

sources/web/datalab/settings.ts
ojarjur: Done.

Issue #1106, opened 2017-01-07
The 'create' subcommand for the CLI requires the 'compute/zone' setting to be set


ojarjur:
There is a bug in the current CLI where the `create` command:

1. Captures the stderr when calling `gcloud compute disks describe`.
2. Does not pass a `--zone` argument to that command.

If you have the `compute/zone` setting set (as I did when previously testing this), then that works fine, but otherwise the gcloud command will prompt you to select a zone (via stderr), that prompt will never be answered, and the command will hang.

We should (at a minimum) do two things to fix this:

1. Do not capture stderr when calling a gcloud command that potentially needs to prompt the user.
2. If the `--zone` argument is passed to the `datalab create` command, then also pass that down to the `gcloud compute disks describe` command.

Even better would be if we prompted the user to select a zone if none was yet specified, so that we could ensure the disk and VM were created in the same zone.

PR #1109, opened 2017-01-07
Proactively prompt the user for a zone instead of relying on gcloud to do so

ojarjur:
This change makes the CLI tool detect if `gcloud` needs to prompt the
user for a zone, and then takes over responsibility for doing that.

There are multiple reasons we want to do this:

1. For the `create` command, we have to do this since otherwise the
   user would be prompted multiple times, and they would have to
   answer the same way every time in order to avoid issues.
2. For other commands, we would like the zone prompt to look similar
   to what we do for the `create` command.
3. For the `connect`, `stop`, and `delete` commands, it makes sense
   to simply default to the zone the instance is in if there is only
   one matching instance in the project.

### DIFF
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -17,10 +17,11 @@
 import os
 import os.path
 import subprocess
-import tempfile
 import threading
 import urllib2
 import webbrowser
+
+import utils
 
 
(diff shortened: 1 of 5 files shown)


ojarjur:
This improves upon the earlier fix for #1106

PR #1115, opened 2017-01-11
Datalab test framework

yebrahim:
This is based on @gramster's PR https://github.com/googledatalab/datalab/pull/696, which uses Selenium to open notebooks and make sure they produce expected results. I changed it to test that the cells run without errors, but I do not compare against results as this was causing lots of issues, and it means notebooks have to always contain their run results in the cells.
I also added the following:

- Switched to just running cells in notebooks and making sure they don't produce errors
- Skipped EULA for test containers
- Running browser sessions in parallel
- Used only three tests to begin with, all run Python only, no cloud functionality tested yet
- Used headless Firefox (with geckodriver and xvfb) to navigate to test notebooks
- Added a dedicated test image that builds and runs the configuration above instead of modifying the Datalab image on the fly, and bundled tests inside that image
- Added travi-ci modifications to build and run a test container

Added packages:
- [Firefox](https://www.mozilla.org/en-US/foundation/licensing/) and geckodriver are [MPL](https://github.com/mozilla/geckodriver/blob/master/LICENSE)
- xvfb is [MIT](https://github.com/cgoldberg/xvfbwrapper/blob/master/LICENSE)


containers/test/run.sh
```diff
@@ -0,0 +1,35 @@
...
+  fi
+  if [ "$1" == "shell" ]; then
+    ENTRYPOINT="/bin/bash"
+  fi
+fi
+
+# disable EULA for the sake of tests
+mkdir -p $CONTENT/datalab/.config/eula
+
+# On linux docker runs directly on host machine, so bind to 127.0.0.1 only
```
ojarjur: This bit is obsolete. New versions of Docker support 127.0.0.1 port binding on all platforms, so we should just make the portmap fixed.

yebrahim: I don't think this was working anyway. It's missing one more '[' and ']' right?

ojarjur: >  It's missing one more '[' and ']' right?

I don't think it's using any of the additional functionality that "[["/"]]" provides over "["/"]"

parthea: @ojarjur 
Just to clarify,  based on the comments above I understand that port binding to IP 127.0.0.1 is still needed. The obsolete part is that binding to 127.0.0.1 it is no longer limited to Linux only.  Can you confirm?

ojarjur: @parthea Sorry that I was ambiguous.

Yes, binding to 127.0.0.1 is still required. Docker now supports that on all platforms it supports, so the check for linux is now obsolete and we can hardcode the portmap to "127.0.0.1:8081:8080"

gramster:
In case this is blocked by needing my approval, you certainly have it. I'm disappointed that the output comparison part has been dropped but happy for this finally to be merged.

OTOH I have toyed occasionally with taking my original change and trying to make it into a Jupyter extension; how would the project owners feel about that? Not that I have a lot of time these days for such things...

yebrahim:
@gramster I think we'll need to add something similar in the future, but for the initial change I wanted something as simple as possible, as I did see some flakiness caused by MIME types. In addition, it requires that the cell outputs are saved in the tested notebooks, which is not always the case, and sometimes cannot be done (e.g. cell returns PID for started tensorboard). I do think it'll be useful in the future to add it for certain cells.

containers/test/run.sh
```diff
@@ -0,0 +1,34 @@
...
+else
+  PORTMAP="8081:8080"
+fi
+# Use this flag to map in web server content during development
+#  -v $REPO_DIR/sources/web:/sources \
+docker run -it --entrypoint=$ENTRYPOINT \
+  -v "$CONTENT:/content" \
+  -p $PORTMAP \
+  -p 8910:8910 \
+  -e "DATALAB_ENV=local" \
```
ojarjur: We should also add a line of:

    -e "ENABLE_USAGE_REPORTING=false" \

So that test runs do not get reported like real usage.

containers/datalab/run.sh
```diff
@@ -32,20 +32,10 @@ if [ "$1" != "" ]; then
...
-if [ "$OSTYPE" == "linux"* ]; then
-  PORTMAP="127.0.0.1:8081:8080"
-else
-  PORTMAP="8081:8080"
-fi
 # Use this flag to map in web server content during development
 #  -v $REPO_DIR/sources/web:/sources \
 docker run -it --entrypoint=$ENTRYPOINT \
-  -p $PORTMAP \
+  -p 8081:8080 \
```
ojarjur: This should be 127.0.0.1:8081:8080

containers/gateway/run.sh
```diff
@@ -32,20 +32,10 @@ if [ "$1" != "" ]; then
...
-if [ "$OSTYPE" == "linux"* ]; then
-  PORTMAP="127.0.0.1:8082:8080"
-else
-  PORTMAP="8082:8080"
-fi
 # Use this flag to map in web server content during development
 #  -v $REPO_DIR/sources/web:/sources \
 docker run -it --entrypoint=$ENTRYPOINT \
-  -p $PORTMAP \
+  -p 8082:8080 \
```
ojarjur: This should be 127.0.0.1:8082:8080

containers/test/run.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+# disable EULA for the sake of tests
+mkdir -p $CONTENT/datalab/.config/eula
+
+# Use this flag to map in web server content during development
+#  -v $REPO_DIR/sources/web:/sources \
+docker run -it --entrypoint=$ENTRYPOINT \
+  -v "$CONTENT:/content" \
+  -p 8081:8080 \
```
ojarjur: I think both of these port mappings should be limited to 127.0.0.1

yebrahim:
@ojarjur can you help merge this please?

PR #1117, opened 2017-01-11
Add a flag to the CLI to proxy websocket connections.

ojarjur:
This is needed for users who want to connect to their
Datalab instances from the Cloud Shell, since the Cloud
Shell's "web preview" feature does not support websockets.


ojarjur:
I'm going to drop this PR in favor of the approach used in https://github.com/googledatalab/datalab/pull/1120

Hopefully, once that one is merged and a new release goes out, connections from Cloud Shell will work regardless of how the user creates their instance.

PR #1120, opened 2017-01-12
Automatically detect if websockets need to be proxied.

ojarjur:
This change removes the previously used 'PROXY_WEB_SOCKETS' environment
variable, and instead detects on the client whether or not that client
needs to proxy websocket connections via socket.io.

This gives us the benefits that:

1. The user no longer needs to decide ahead of time whether or not
   their instance will rely on websockets.
2. A single instance can both rely on websockets for some clients
   and proxy them for others (e.g. depending on whether an SSH
   tunnel is coming from a Cloud Shell instance or from something else).

For now the logic for this decision is based on whether or not the
hostname ends in '.appspot.com' (indicating that it is hosted by
App Engine which does not support websockets).

However, in the future we may want to make the logic more sophisticated
by first trying to create a websocket connection, and then proxying
them if that fails.

### DIFF
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -18,7 +18,7 @@ var debug = {
 };
 
 function shouldShimWebsockets() {
-    return location.host.substr(-12) === '.appspot.com';
+    return location.host.toLowerCase().substr(-12) === '.appspot.com';
 }
 
 // Override WebSocket
@@ -106,7 +106,7 @@ function shouldShimWebsockets() {
 
     if (shouldShimWebsockets()) {
         debug.log('Replacing native websockets with socket.io');
-        var nativeWebSocket = window.WebSocket;
+        window.nativeWebSocket = window.WebSocket;
         window.WebSocket = WebSocketShim;
     }
 })();
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -12,13 +12,17 @@
  * the License.
  */
 
+var debug = {
+  enabled: true,
+  log: function() { console.log.apply(console, arguments); }
+};
+
+function shimWebSockets() {
+    return location.host.substr(-12) === '.appspot.com';
+}
+
 // Override WebSocket
 (function() {
-    var proxyWebSockets = (document.body.getAttribute('data-proxy-web-sockets') == 'true');
-    if (!proxyWebSockets) {
-	return;
-    }
-
     if (!window.io) {
 	// If socket.io was not loaded into the page, then do not override the existing
 	// WebSocket functionality.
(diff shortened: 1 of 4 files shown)


sources/web/datalab/static/datalab.js
```diff
@@ -12,13 +12,17 @@
  * the License.
  */
 
+var debug = {
+  enabled: true,
+  log: function() { console.log.apply(console, arguments); }
+};
+
+function shimWebSockets() {
```
yebrahim: add a verb, e.g. shouldShimWebSockets, or canUseWebSockets (returns not equal in this case)

sources/web/datalab/static/datalab.js
```diff
@@ -12,13 +12,17 @@
...
+    return location.host.substr(-12) === '.appspot.com';
```
yebrahim: this doesn't catch all cases though. we need it disabled for tests for example

yebrahim:
Why are we removing the option? I'd keep it even when we have this change.

sources/web/datalab/static/datalab.js
ojarjur: Why would we need to disable websockets for tests?

qimingj: location.host.endsWith('.appspot.com')?

sources/web/datalab/static/datalab.js
ojarjur: Done.

sources/web/datalab/static/datalab.js
ojarjur: I wanted to avoid 'endsWith' because its relatively new and I was worried about users with older browsers (e.g. versions of IE prior to IE 12)

sources/web/datalab/static/datalab.js
```diff
@@ -100,15 +104,13 @@
     WebSocketShim.CLOSED = 0;
     WebSocketShim.OPEN = 1;
 
-    var nativeWebSocket = window.WebSocket;
-    window.WebSocket = WebSocketShim;
+    if (shouldShimWebsockets()) {
+        debug.log('Replacing native websockets with socket.io');
+        var nativeWebSocket = window.WebSocket;
```
nikhilk: This creates a local var, that won't be accessible outside. Should window.nativeWebSocket be set instead?

sources/web/datalab/static/datalab.js
```diff
@@ -12,13 +12,17 @@
  * the License.
  */
 
+var debug = {
+  enabled: true,
+  log: function() { console.log.apply(console, arguments); }
+};
+
+function shouldShimWebsockets() {
+    return location.host.substr(-12) === '.appspot.com';
```
nikhilk: The answer is probably not required, but should this be lower-cased before comparison?

sources/web/datalab/static/datalab.js
yebrahim: websockets didn't work with SauceLabs, but we changed that. I just tested that it works with headless Firefox, so we're fine.

sources/web/datalab/static/datalab.js
ojarjur: Done.

sources/web/datalab/static/datalab.js
ojarjur: Done.

PR #1128, opened 2017-01-17
Support configuring the author email address for git.

ojarjur:
With this change, Datalab will configure git with an author email
address at startup. That can also be overridden via an environment
variable.

### DIFF
--- a/containers/datalab/content/setup-env.sh
+++ b/containers/datalab/content/setup-env.sh
@@ -27,3 +27,6 @@ export CLOUDSDK_CONFIG=/content/datalab/.config
 export PROJECT_ID=${PROJECT_ID:-`gcloud config list -q --format 'value(core.project)' 2> /dev/null`}
 export ZONE=${ZONE:-`gcloud config list -q --format 'value(compute.zone)' 2> /dev/null`}
 
+# Lookup the author email address to use for git commits, and then configure git accordingly
+export GIT_AUTHOR=${DATALAB_GIT_AUTHOR:-`gcloud auth list --format 'value(account)' --filter 'status:ACTIVE'`}
+git config --global user.email "${GIT_AUTHOR}"
--- a/containers/datalab/content/setup-env.sh
+++ b/containers/datalab/content/setup-env.sh
@@ -28,5 +28,6 @@ export PROJECT_ID=${PROJECT_ID:-`gcloud config list -q --format 'value(core.proj
 export ZONE=${ZONE:-`gcloud config list -q --format 'value(compute.zone)' 2> /dev/null`}
 
 # Lookup the author email address to use for git commits, and then configure git accordingly
-export GIT_AUTHOR=${DATALAB_GIT_AUTHOR:-`gcloud auth list --format 'value(account)' --filter 'status:ACTIVE'`}
-git config --global user.email "${GIT_AUTHOR}"
+export DATALAB_GIT_AUTHOR=${DATALAB_GIT_AUTHOR:-`gcloud auth list --format 'value(account)' --filter 'status:ACTIVE'`}
+git config --global user.email "${DATALAB_GIT_AUTHOR}"
+git config --global user.name "${DATALAB_GIT_AUTHOR}"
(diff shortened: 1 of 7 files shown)


containers/datalab/content/setup-env.sh
```diff
@@ -27,3 +27,6 @@ export CLOUDSDK_CONFIG=/content/datalab/.config
 export PROJECT_ID=${PROJECT_ID:-`gcloud config list -q --format 'value(core.project)' 2> /dev/null`}
 export ZONE=${ZONE:-`gcloud config list -q --format 'value(compute.zone)' 2> /dev/null`}
 
+# Lookup the author email address to use for git commits, and then configure git accordingly
+export GIT_AUTHOR=${DATALAB_GIT_AUTHOR:-`gcloud auth list --format 'value(account)' --filter 'status:ACTIVE'`}
```
yebrahim: Should this fail instead of using the service account for git, which is a bad config in this case?

tools/cli/datalab.py
```diff
@@ -93,6 +98,10 @@
 environment variable CLOUDSDK_COMPUTE_ZONE.
 """)
 
+_EMAIL_HELP = ("""The email address associated with the instance.
+
+This should represent the user of Datalab.""")
```
yebrahim: I think we should add something about using the email for git here, otherwise it's not clear why this is suddenly required

tools/cli/datalab.py
```diff
@@ -170,6 +192,12 @@ def run():
                 dest='zone',
                 default=None,
                 help=_ZONE_HELP)
+        if command_config['require-email']:
+            subcommand_parser.add_argument(
+                '--email',
+                dest='email',
+                default=get_email_address(),
```
yebrahim: Shouldn't this be a required argument instead of using service account?

tools/cli/datalab.py
```diff
@@ -170,6 +192,12 @@ def run():
                 dest='zone',
                 default=None,
                 help=_ZONE_HELP)
+        if command_config['require-email']:
+            subcommand_parser.add_argument(
+                '--email',
```
nikhilk: A question:
- Is this only for git user -- if so git-user instead of email is likely much more descriptive/appropriate.

Thought:
- Do we actually need this? Won't the default always suffice, esp. with cloud source repository?

containers/datalab/content/setup-env.sh
ojarjur: I don't agree that using the service account is a bad config.

It may be ugly, but it is technically correct. Additionally, advanced users may create descriptively named service accounts for which the service account is actually the best email address to use.

More importantly, we don't expect that case to actually happen.

Users will usually use the CLI to create Datalab instances running in GCE VMs, and that tool will set the right environment variable.

That means the case where this will happen is for someone running Datalab locally, and in that case their gcloud credentials should be the right email address to use.

tools/cli/datalab.py
ojarjur: This wouldn't default to a service account unless the user runs `datalab create` from inside of a VM, and that seems like a rare scenario.

tools/cli/datalab.py
ojarjur: This is just a temporary solution I am using to test the changes to the Docker image.

This change isn't ready for review yet.

tools/cli/commands/connect.py
```diff
@@ -293,7 +293,7 @@ def maybe_start(args, gcloud_compute, instance):
     return
 
 
-def run(args, gcloud_compute):
+def run(args, gcloud_compute, **unused_kwargs):
```
yebrahim: what does unused_kwargs add here if we're not passing it around?

ojarjur: Keyword args are being passed in. There has to be a corresponding **_something_ parameter, or that call will fail.

tools/cli/datalab.py
ojarjur: FYI, this is ready for review now.

nikhilk: I think my comment still holds ... about this being worded as email in the flags, as opposed to git-user or something that indicates use.

ojarjur: To clarify, the '--email' flag is now gone.

PR #1130, opened 2017-01-18
Add option to shutdown VM from Datalab

yebrahim:
This PR adds a link in the profile menu to manage the VM hosting Datalab and a button to shut it down. The needed information is stored in environment variables that are computed in the startup script. The VM section appears only if Datalab is hosted inside a google cloud VM.

image

### DIFF
--- a/containers/gateway/content/run.sh
+++ b/containers/gateway/content/run.sh
@@ -25,11 +25,6 @@ then
   . ~/startup.sh
 fi
 
-# Get VM information if running on google cloud
-machine_metadata_url="http://metadata.google.internal/computeMetadata/v1/instance"
-export VM_NAME=$(curl -s "${machine_metadata_url}/hostname" -H "Metadata-Flavor: Google" | cut -d '.' -f 1)
-export VM_ZONE=$(curl -s "${machine_metadata_url}/zone" -H "Metadata-Flavor: Google" | sed 's/.*zones\///')
-
 # Start the DataLab kernel gateway.
 jupyter kernelgateway --JupyterWebsocketPersonality.list_kernels=True --KernelGatewayApp.port=8081 --KernelGatewayApp.ip=0.0.0.0 &
 n=0; while true; do
--- a/containers/gateway/content/run.sh
+++ b/containers/gateway/content/run.sh
@@ -25,6 +25,11 @@ then
   . ~/startup.sh
 fi
 
+# Get VM information if running on google cloud
+machine_metadata_url="http://metadata.google.internal/computeMetadata/v1/instance"
+export VM_NAME=$(curl -s "${machine_metadata_url}/hostname" -H "Metadata-Flavor: Google" | cut -d '.' -f 1)
+export VM_ZONE=$(curl -s "${machine_metadata_url}/zone" -H "Metadata-Flavor: Google" | sed 's/.*zones\///')
+
 # Start the DataLab kernel gateway.
 jupyter kernelgateway --JupyterWebsocketPersonality.list_kernels=True --KernelGatewayApp.port=8081 --KernelGatewayApp.ip=0.0.0.0 &
 n=0; while true; do
--- a/sources/web/datalab/server.ts
+++ b/sources/web/datalab/server.ts
@@ -35,6 +35,7 @@ import url = require('url');
 import userManager = require('./userManager');
 import wsHttpProxy = require('./wsHttpProxy');
 import backupUtility = require('./backupUtility');
+import childProcess = require('child_process');
 
 var server: http.Server;
 var healthHandler: http.RequestHandler;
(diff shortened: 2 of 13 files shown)


sources/web/datalab/server.ts
```diff
@@ -185,6 +186,21 @@ function handleRequest(request: http.ServerRequest,
     return;
   }
 
+  if (path.indexOf('/_killvm') == 0) {
```
ojarjur: This should be "/_stopvm" rather than "/_killvm"

sources/web/datalab/server.ts
```diff
@@ -185,6 +186,21 @@ function handleRequest(request: http.ServerRequest,
...
+    if ('POST' != request.method) {
+      return;
+    }
+    try {
+      let vminfo = info.getVmInfo();
+      childProcess.execSync(
+        'gcloud compute instances stop ' + vminfo.VM_NAME + ' --zone ' + vminfo.VM_ZONE,
```
ojarjur: This should explicitly set the project rather than assuming that the current project (which the user can change) is the hosting project.

sources/web/datalab/server.ts
```diff
@@ -185,6 +186,21 @@ function handleRequest(request: http.ServerRequest,
     return;
   }
 
+  if (path.indexOf('/_killvm') == 0) {
+    if ('POST' != request.method) {
```
ojarjur: I think it would be cleaner if this whole block was moved to a helper method called something like "stopHandler"

sources/web/datalab/static/datalab.js
```diff
@@ -183,30 +186,80 @@ function restartDatalab() {
...
+        '<span class="fa fa-external-link-square">&nbsp;</span><a href="/static/about.txt" target="_blank">License and software information</a><br />' +
+        '<span class="fa fa-external-link-square">&nbsp;</span><a href="https://cloud.google.com/terms/" target="_blank">Terms of Service</a><br />' +
+        '<span class="fa fa-external-link-square">&nbsp;</span><a href="http://www.google.com/intl/en/policies/" target="_blank">Privacy Policy</a><br />' +
+        '<span class="fa fa-external-link-square">&nbsp;</span><a href="/static/reporting.html?enabled=' + reportingEnabled + '" target="_blank">Usage Statistics</a><br />' +
+        '<span class="fa fa-recycle">&nbsp;</span><a href="javascript:restartDatalab()">Restart Server</a><br />';
+
+      if (vminfo && vminfo.VM_NAME) {
+        dialogContent += '<hr><div>This Datalab instance is running from a Google Cloud VM: <b>' + vminfo.VM_NAME + '</b></span></div>' +
+        '<div><a href="javascript:manageVm()">Click here to manage the VM</a></div>' +
+        '<br><button id="killVmButton" title="Shutdown VM" class="btn btn-danger" onclick="killVm()">Shutdown VM</button>';
```
ojarjur: Change "Shutdown VM" to "Stop VM".

nikhilk:
Some tangential comments on this dialog... its a mix of fyi links and user actions.

Can you put the ones that are user actions together (manage, vm details link) on top and fyi stuff below? Also, what is usage statistics? Why is it in the user's view?

Restart server -- this wasn't meant to be exposed to users either.

yebrahim:
@nikhilk This is the info dialog though, this is why I think information should appear first. Eventually we should move this button to a settings pane anyway, it'll always feel out of place here. What do you think?

Usage statistics is actually usage reporting opt in/out, I can rename it to "Usage reporting", or "Usage reporting preferences"
For restart server, where should this go? I think it's a useful functionality to be exposed.

sources/web/datalab/static/datalab.js
```diff
@@ -160,7 +160,10 @@ function xhr(url, callback, method) {
   method = method || "GET";
 
   let request = new XMLHttpRequest();
-  request.onreadystatechange = callback.bind(request);
+  request.onreadystatechange = function() {
+    if (request.readyState === 4 && request.status === 200)
```
ojarjur: This should have opening and closing brackets.

containers/gateway/content/run.sh
```diff
@@ -25,6 +25,11 @@ then
   . ~/startup.sh
 fi
 
+# Get VM information if running on google cloud
+machine_metadata_url="http://metadata.google.internal/computeMetadata/v1/instance"
```
ojarjur: Are these variables actually used anywhere?

sources/web/datalab/server.ts
```diff
@@ -250,6 +256,22 @@ function uncheckedRequestHandler(request: http.ServerRequest, response: http.Ser
...
+function stopVmHandler(request: http.ServerRequest, response: http.ServerResponse) {
+  if ('POST' != request.method) {
+    return;
+  }
+  try {
+    let vminfo = info.getVmInfo();
+    childProcess.execSync(
+      'gcloud compute instances stop ' + vminfo.vm_name +
+         ' --project ' + vminfo.vm_project + ' --zone ' + vminfo.vm_zone,
+      {env: process.env});
```
ojarjur: After the gcloud command, we should run 'sudo shutdown -P now' to make sure the VM shuts itself off.

yebrahim: Can't do this inside of a docker container.

dovy:
error: Forever detected script exited with code: 0
error: Script restart attempt #1

With new image in gateway mode...

ojarjur:
@dovy I'm not able to reproduce that error (using the current release of the datalab and datalab-gateway images).

Can you please file a separate issue with more details?

yebrahim:
@dovy When does this happen? Why is it related to this PR? The shutdown button shouldn't appear in the gateway mode. Please file a new issue with more details.

dovy:
Had to end up deleting my local image and specifying the full version. If I did a pull on local, but ran it without the :local-xxx it failed.

Resolved.

PR #1132, opened 2017-01-19
Automate integration with Google Cloud Source Repositories

ojarjur:
This commit changes the CLI to automatically do two things:

1. Create the `datalab-notebooks` repository if it does not exist
2. Clone that repository into the `datalab/notebooks` directory on
   the attached persistent disk when creating a new disk.

With this change (and the next Datalab release), users will be
able to use the ungit UI to manage their notebooks without having
to do any manual configuration.


OrielResearchCure:
Hi, I would like to work with a different Google repository. What will be the easiest way to have it cloned on any new datalab machine?
Many thanks,
eilalan

PR #1145, opened 2017-01-23
Upgrade the version of the 'six' library to 1.10.0

ojarjur:
This fixes a build breakage that just showed up today (builds
were working this morning, and failing now).

I was not able to isolate exactly what changed, but it seems that
our earlier builds were installing 1.10.0 as a dependency of
something else, and then the explicit install of 1.9.0 was skipped
since the existing 1.10.0 install satisified that dependency.

The failing builds seem to not have the 1.10.0 install, so the
1.9.0 install actually goes through, and later on the 'future'
library fails because it requires 1.10.0.

This change fixes that build breakage by switching our explicit
dependency to be the one that we were implicitly depending on.


yebrahim:
FWIW, I looked at the diff between one of the failing builds and an older successful build, these have changed:

- nbconvert from 5.0.0 to 5.1.1
- setuptools from 33.1.1 to 34.0.0
- packaging, appdirs getting installed (not so before)

PR #1148, opened 2017-01-24
Restart the SSH tunnel if it dies

ojarjur:
This change spins up a background process that continuously watches the SSH process to see if it is still running.

If it finds that it is not, it tries to recreate the SSH tunnel.

This change also includes a fix for an issue where Datalab would not start if it could not communicate with a metadata server (e.g. if it was running locally). That has been sent in a separate PR, and once it is merged I will rebase this PR.

### DIFF
--- a/containers/datalab/content/run.sh
+++ b/containers/datalab/content/run.sh
@@ -81,6 +81,15 @@ setup_tunnel() {
     exit "${ERR_GATEWAY_FAILED}"
   fi
   echo "Successfully established SSH tunnel to ${instance}"
+
+  watch_and_restart_tunnel &
+}
+
+watch_and_restart_tunnel() {
+  while ps --ppid $$ -C ssh 2>&1 > /dev/null; do
+    sleep 1
+  done
+  setup_tunnel > /dev/null
 }
 
 source /datalab/setup-env.sh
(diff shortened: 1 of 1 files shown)


ojarjur:
@yebrahim @qimingj 

The logic for shutting down turned out to be more complicated than originally anticipated, but it seems to work now.

PTAL

containers/datalab/content/run.sh
```diff
@@ -63,33 +80,39 @@ setup_tunnel() {
...
 
 watch_and_restart_tunnel() {
-  while ps --ppid $$ -C ssh 2>&1 > /dev/null; do
+  project_id=$1
+  zone=$2
+  instance=$3
+  ssh_user=$4
+  
+  while [ "$(ps --ppid $$ -o comm= | grep ssh)" != "" ]; do
     sleep 1
```
qimingj: It is probably not a big deal but 5 seconds as interval might save some CPU cycles on the VM (don't know the cost of ps command). Establishing ssh tunnel itself takes more than several seconds anyway.

ojarjur: Done.

jiajiayb:
Do we need to use a specific daily build of gcr.io/cloud-datalab for restarting ssh tunnel? I am using datalab:local-20170110 and just wonder if above changes integrated into this specific daily build? Thanks!

PR #1149, opened 2017-01-25
Fix license file headers.

ojarjur:
This change makes improvements to the license headers in the assorted
source files in the Datalab repository by doing four things:

1. Separating license headers from other comments so that they
   do not get conflated together.
2. Move the externs under a third_party directory to make it explicit
   that they should not have license headers added to them.
3. Adding license headers to a few recently added source files
   that were missing them.
4. Move the license.txt file to under third_party, since it documents
   the third-party licenses.


containers/datalab/build.sh
```diff
@@ -36,6 +36,9 @@ cd ../../containers/datalab
 # Copy build outputs as a dependency of the Dockerfile
 rsync -avp ../../build/ build
 
+# Copy the license file into the container
+cp ../../third_party/license.txt content/license.txt
```
yebrahim: Can we add this to gitignore as well?

ojarjur: Done.

PR #1152, opened 2017-01-26
Cleanup backup tar after copying to GCS

yebrahim:
Also, try using project_id as bucket name if other options have failed.

### DIFF
--- a/containers/datalab/content/GCSbackup.sh
+++ b/containers/datalab/content/GCSbackup.sh
@@ -123,7 +123,11 @@ fi
 # test and create bucket if necessary
 gsutil ls gs://${gcs_bucket} &>/dev/null || {
   echo "Bucket '${gcs_bucket}' was not found. Creating it.."
-  gsutil mb gs://"${gcs_bucket}"
+  gsutil mb gs://"${gcs_bucket}" || {
+    echo "Bucket '${gcs_bucket}' coult not be created. Will try bucket name ${project_id}"
+    gcs_bucket="${project_id}"
+    gsutil mb gs://"${gcs_bucket}" || true
+  }
 }
 
 # create an archive of the backup path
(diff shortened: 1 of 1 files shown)


containers/datalab/content/GCSbackup.sh
```diff
@@ -123,7 +123,11 @@ fi
 # test and create bucket if necessary
 gsutil ls gs://${gcs_bucket} &>/dev/null || {
   echo "Bucket '${gcs_bucket}' was not found. Creating it.."
-  gsutil mb gs://"${gcs_bucket}"
+  gsutil mb gs://"${gcs_bucket}" || {
```
ojarjur: I'd suggest a slightly different ordering of the logic where we only try to run `gsutil mb` once:

Basically, we'd look at the "gcs_bucket" (without any default), and "project_id" environment variables to figure out what the bucket name should be.

If "gcs_bucket" is defined, then that is the bucket name. Otherwise we run `gsutil ls gs://${project_id}.appspot.com` to see if that bucket exists, and if so set the bucket name to it. Finally, if neither of the two previous tests passed, then we set the bucket name to match the project ID.

yebrahim: Do you mean we shouldn't create the `gs://${project_id}.appspot.com` bucket if it doesn't exist? Other than this, I don't see much difference between the comment and the changes.

ojarjur: In addition to not trying to create the "...appspot.com" bucket (which we know would always fail), the other difference between what I proposed and the current approach is that we would not try to create the project bucket if the user has explicitly specified a gcs_bucket, regardless of whether or not that gcs_bucket exists.

containers/datalab/content/GCSbackup.sh
```diff
@@ -96,10 +96,26 @@ timestamp=$(date "+%Y%m%d%H%M%S")
...
+  gsutil ls "gs://${default_bucket}" &>/dev/null && gcs_bucket="${default_bucket}" || {
+    # We cannot create the $project_id.appspot.com bucket, so don't try that
+    gcs_bucket=${project_id}
+    echo "Could not list bucket ${default_bucket}. Next will use ${gcs_bucket} for bucket name"
+    gsutil mb gs://"${gcs_bucket}" || true
+  }
+else
+  gsutil ls gs://"${gcs_bucket}" &>/dev/null || {
+    echo "Could not list bucket '${gcs_bucket}'. Will try to create it.."
+    gsutil mb gs://"${gcs_bucket}" || true
```
ojarjur: Should this have an "|| true"?

If this mb call fails, then the script should exit.

Issue #1157, opened 2017-01-27
Make the nodejs server logs when running in a VM show up in StackDriver logs


ojarjur:
Right now, the `datalab create` command creates a VM instance running Datalab in 'silent' mode (with stderr and stdout from the NodeJS server being discarded).

That means any errors reported by the NodeJS server will not show up in the StackDriver logs. We should fix this so that the errors are logged appropriately.

PR #1158, opened 2017-01-27
Add a config setting for the console log level.

ojarjur:
Previously, if console logging was enabled, then the log threshold
level was hard-coded to 'info'.

This change both makes that configurable, and changes the default
from 'info' (which is very noisy) to 'warn' (which should only
include things the user really needs to know).

This is part of the work to address #1157

### DIFF
--- a/sources/web/datalab/config/settings.json
+++ b/sources/web/datalab/config/settings.json
@@ -1,5 +1,6 @@
 {
   "consoleLogging": true,
+  "consoleLogLevel": "warn",
   "release": "beta",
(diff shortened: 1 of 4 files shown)


sources/web/datalab/config/settings.json
```diff
@@ -1,5 +1,6 @@
 {
   "consoleLogging": true,
+  "consoleLogLevel": 'warn',
```
yebrahim: Can we add a comment in the run scripts as an example how to set this to verbose?

ojarjur: Done

PR #1159, opened 2017-01-28
Configure StackDriver logging for Datalab instances.

### LINKED ISSUE #1157: Make the nodejs server logs when running in a VM show up in StackDriver logs

ojarjur:
Previously, Datalab instances wrote a very small amount of
information to StackDriver logging, and then went silent.

With this change, the logs from the Datalab server that are
higher than a configurable threshold will be written to StackDriver
logging.

This change also fixes some badly-named SSH flags that are
used to configure the verbosity of the SSH output to indicate
that those flags are specific to just the SSH connection.

This fixes #1157

### DIFF
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -93,8 +93,8 @@ def connection_flags(parser):
             '\n\n'
             'A negative value means no limit.'))
     parser.add_argument(
-        '--log-file',
-        dest='log_file',
+        '--ssh-log-file',
+        dest='ssh_log_file',
         default=os.devnull,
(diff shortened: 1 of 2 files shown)


ojarjur:
The automated tests were failing for this because of the build breakage from last week, but now that is fixed and this is ready for review.

PR #1164, opened 2017-01-30
Install Datalab Python library using pip

yebrahim:
The Datalab python library is now on PyPI.

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -163,7 +163,7 @@ RUN ipython profile create default && \
         pip install --upgrade-strategy only-if-needed --no-cache-dir . && \
         rm datalab/notebook/static/*.js; \
       else \
-        pip install datalab; \
+        pip install --upgrade-strategy only-if-needed --no-cache-dir datalab==0.1.1701301349; \
       fi && \
     jupyter nbextension install --py datalab.notebook && \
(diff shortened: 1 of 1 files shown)


containers/base/Dockerfile
```diff
@@ -155,18 +155,18 @@ ADD nbconvert /datalab/nbconvert
...
+        echo "use local lib" && \
+        cd /datalab/lib/pydatalab && \
+        /tools/node/bin/npm install -g typescript && \
+        tsc --module amd --noImplicitAny --outdir datalab/notebook/static datalab/notebook/static/*.ts && \
+        /tools/node/bin/npm uninstall -g typescript && \
+        pip install --upgrade-strategy only-if-needed --no-cache-dir . && \
+        rm datalab/notebook/static/*.js; \
       else \
-        git clone https://github.com/googledatalab/pydatalab.git /datalab/lib/pydatalab; \
+        pip install --upgrade-strategy only-if-needed --no-cache-dir datalab==0.1.1701301349; \
```
ojarjur: I'm not sure that we should pin the version of the 'datalab' library.

Our dependencies should absolutely be pinned, but for our own library, I worry that we will always want the latest version. If that's the case, then pinning it will just add a manual step we always have to remember to perform.

yebrahim: Sure.

PR #1166, opened 2017-01-31
Update TensorFlow and CloudML SDK. Fix a build bug.

qimingj:
Update TensorFlow and CloudML SDK.

We can remove protobuf installation because installation of TensorFlow includes protobuf. We can remove DataFlow installation because installation of CloudML SDK includes DataFlow.

Fix a bug that when it fails to build the container, an intermediate dir is not cleaned up.

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -78,6 +78,7 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
     pip install -U --upgrade-strategy only-if-needed --no-cache-dir pillow==3.4.1 && \
     find /usr/local/lib/python2.7 -type d -name tests | xargs rm -rf && \
 
+
 # Setup Node.js
     mkdir -p /tools/node && \
     wget -nv https://nodejs.org/dist/v4.3.2/node-v4.3.2-linux-x64.tar.gz -O node.tar.gz && \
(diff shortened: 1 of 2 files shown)


containers/base/Dockerfile
```diff
@@ -123,20 +124,6 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
     rm -rf /usr/share/i18n/locales/* && \
     cd /
 
-# Install ProtoBuf, TensorFlow
```
nikhilk: Seems incorrect for pydatalab to bring in TensorFlow or Dataflow. We don't want all users of pydatalab depending on those. The specific features in pydatalab should handle missing dependencies with appropriate error, and we should continue to install these explicitly.

qimingj: That requires more changes. For example, datalab.mlalpha.* will need to be changed to add try catch in importing tensorflow. And the same thing when we import cloudml SDK since it references DataFlow.

What do we lose if we require TF and DataFlow from pydatalab? These are added to setup.py so that Jupyter+pydatalab works. We could do conditional import, but isn't the same can be said for almost all imports?

nikhilk: Those are relatively "big" dependencies. Someone using pydatalab for just working with BigQuery for example shouldn't have to bring them down as well.

qimingj: Done.

qimingj:
PTAL

ojarjur:
The PR description needs to be updated. Right now it's not clear what the scope of this change is meant to be.

qimingj:
PR description updated.

qimingj:
Yes. Description was updated.

PR #1168, opened 2017-02-01
Report errors when creating a disk to the user if the error is terminal.

ojarjur:
This fixes a bug where the errors reported by gcloud when creating
a disk were sometimes discarded without being shown to the user if
the command failed the first time it is run, but will not be retried
(either because the zone has been explicitly set, or the --quiet
flag has been specified, causing us to not prompt for a zone).

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -388,21 +388,21 @@ def ensure_disk_exists(args, gcloud_compute, disk_name, report_errors=False):
         try:
             create_disk(args, gcloud_compute, disk_name, report_errors)
         except:
-            if args.zone or args.quiet:
-                if report_errors:
-                    raise
-                # We know the command failed (and will almost certainly
-                # fail again), but we did not forward the errors that
-                # it reported to the user. To work around this, we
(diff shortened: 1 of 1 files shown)


tools/cli/commands/create.py
```diff
@@ -389,7 +389,13 @@ def ensure_disk_exists(args, gcloud_compute, disk_name, report_errors=False):
             create_disk(args, gcloud_compute, disk_name, report_errors)
         except:
             if args.zone or args.quiet:
-                raise
+                if report_errors:
```
yebrahim: Why not just add this to the first if?

ojarjur: I don't want to bake-in assumptions based on the current structure of the code.

Right now, `report_errors` is only true when retrying a previous, failed invocation. However, that might not be the case in the future, and so I was trying to structure the code based on what the fields mean rather than the dataflow of the method.

ojarjur: That being said, the logic could probably still be flattened to keep it simpler.

I will do that.

PR #1172, opened 2017-02-01
Allow configuring the socket.io port.

blois:
This adds a config setting for the socket IO port so the server can be
executed in environments where only specific ports are allowed.

### DIFF
--- a/sources/web/datalab/server.ts
+++ b/sources/web/datalab/server.ts
@@ -303,8 +303,8 @@ export function run(settings: common.Settings): void {
   jupyter.init(settings);
   auth.init(settings);
   noCacheContent.init(settings);
-  reverseProxy.init();
-  sockets.init();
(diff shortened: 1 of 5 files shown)


sources/web/datalab/server.ts
```diff
@@ -303,8 +303,8 @@ export function run(settings: common.Settings): void {
   jupyter.init(settings);
   auth.init(settings);
   noCacheContent.init(settings);
-  reverseProxy.init();
-  sockets.init();
+  reverseProxy.init(settings.socketioPort);
```
ojarjur: This should probably just be 'settings' instead of 'settings.socketioPort', since we'll soon want to add an 'ungitPort' value to the settings object too.

blois: Updated.

Issue #1139, opened 2017-01-20
Don't try to launch a browser if run inside of Google Cloud Shell


ojarjur:
The default behavior of the `datalab create` and `datalab connect` commands is to open a browser pointing at the local endpoint of the SSH tunnel.

There is an option to not launch a browser, but in the case of running in Google Cloud Shell we should be able to detect that and have a smarter default.

Specifically, the tool should probably just output something like:

    The connection to Datalab is now open and will remain until this command is killed.

    Click on the *Web Preview* (up-arrow button at top-left), select *port 8081*, and start using Datalab.

PR #1175, opened 2017-02-02
Do not try to launch the browser if it may not support Datalab.

ojarjur:
This change stops the CLI from automatically launching the browser
if the type of browser launched is one that may not work with the
Datalab UI.

In particular, this detects a large class of text-only browsers,
including the one that is provided in Google Cloud Shell.

This mitigates iessue #1139, but does not completely fix it, as
we still do not tell the user to launch the web-preview feature.

However, even though this approach still leaves room for improving
the usage under Cloud Shell, it also greatly improves the experience
when running under other environments where a text-only browser
would be opened instead of a browser with a GUI.

### DIFF
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -52,6 +52,15 @@ To connect to 'example-instance' in zone 'us-central1-a', run:
     $ {0} {1} example-instance --zone us-central1-a""")
 
 
+# The list of web browsers that we don't want to automatically open.
+#
+# This is a subset of the canonical list of python browser types
+# defined here: https://docs.python.org/2/library/webbrowser.html
+unsupported_browsers = [
+    'BackgroundBrowser', 'Elinks', 'GenericBrowser', 'Grail'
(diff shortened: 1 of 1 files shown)


tools/cli/commands/connect.py
```diff
@@ -52,6 +52,15 @@
     $ {0} {1} example-instance --zone us-central1-a""")
 
 
+# The list of web browsers that we don't want to automatically open.
+#
+# This is a subset of the canonical list of python browser types
+# defined here: https://docs.python.org/2/library/webbrowser.html
+unsupported_browsers = [
```
nikhilk: Instead of a blacklist, could we have a list of supported browsers - Chrome, FF, Edge, Safari?

ojarjur: There are two reasons I didn't do that:

1. That list is a lot longer
2. According to the Python docs, the combined list is hard-coded, so we don't have much risk of new ones showing up that we don't support.

PR #1178, opened 2017-02-04
Prevent users from accidentally sharing a Datalab instance.

### LINKED ISSUE #1165: Warn users if they try to use someone else's Datalab instance
Datalab instances are single-user environments (so each user should have their own), but projects can have multiple members.

As such, the tool should warn a user if they try to connect to a Datalab instance that was created for someone else.

ojarjur:
The overall goal of this change is to make it clear to users that
the Datalab instances they create should be tied to a single user.

We do this by recording for which user the Datalab instance was
created, and then checking that when a user tries to connect to
the instance.

By default, the Datalab instance is created for the user who runs
the `datalab create` command, but there is a new `--for-user` flag
that can be used to override that.

When a user tries to connect to a Datalab instance that was created
for someone else, the code prints out an error message explaining
the user mismatch.

The user still has the option of disabling that check by passing
in the new `--no-user-checking` flag, but this should prevent
users from accidentally stomping on each others' toes.

This fixes #1165

### DIFF
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -57,10 +57,10 @@ wrong_user_message = (
     'are attempting to connect to it as {1}.'
     '\n\n'
     'Datalab instances are single-user environments, and trying '
-    'to share one will cause problems.'
+    'to share one is not supported.'
     '\n\n'
-    'If you are sure you want to do this, then re-run the command '
-    'with the `--no-user-checking` flag.')
+    'To override this behavior, re-run the command with the '
+    '--no-user-checking flag.')
 
 
 def flags(parser):
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -52,6 +52,17 @@ To connect to 'example-instance' in zone 'us-central1-a', run:
     $ {0} {1} example-instance --zone us-central1-a""")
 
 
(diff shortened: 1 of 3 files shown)


tools/cli/commands/connect.py
```diff
@@ -52,6 +52,17 @@
     $ {0} {1} example-instance --zone us-central1-a""")
 
 
+wrong_user_message = (
+    'The specified Datalab instance was created for {0}, but you '
+    'are attempting to connect to it as {1}.'
+    '\n\n'
+    'Datalab instances are single-user environments, and trying '
+    'to share one will cause problems.'
```
nikhilk: Suggest:
to share one is not supported.

tools/cli/commands/connect.py
```diff
@@ -52,6 +52,17 @@
...
+    '\n\n'
+    'If you are sure you want to do this, then re-run the command '
+    'with the `--no-user-checking` flag.')
```
nikhilk: Suggest:
To override this behavior, re-run the command with the --no-user-check flag.

ojarjur: Done

tools/cli/commands/connect.py
ojarjur: Done

PR #1183, opened 2017-02-07
Detect when the CLI is run from Google Cloud Shell 

### LINKED ISSUE #1139: Don't try to launch a browser if run inside of Google Cloud Shell

ojarjur:
This change teaches the CLI to detect when it is running inside
of the Google Cloud Shell, and then handle successful connections
differently for that case.

The purpose of this is that the user cannot connect to the 'local'
port in Cloud Shell via the 'localhost' domain name, but rather
must use the 'Web Preview' feature of Cloud Shell to connect to
their Datalab instance. As such, the tool should tell the user
to connect using the 'Web Preview' feature.

This fixes #1139

### DIFF
--- a/tools/cli/datalab.py
+++ b/tools/cli/datalab.py
@@ -224,7 +224,8 @@ def run():
         _SUBCOMMANDS[args.subcommand]['run'](
             args, gcloud_compute,
             gcloud_repos=gcloud_repos,
-            email=get_email_address())
+            email=get_email_address(),
+            in_cloud_shell=('DEVSHELL_CLIENT_PORT' in os.environ))
     except subprocess.CalledProcessError:
         print('A nested call to gcloud failed.')
 
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -240,8 +240,8 @@ def connect(args, gcloud_compute, email, in_cloud_shell):
 
     def on_ready():
         """Callback that handles a successful connection."""
-        print('The connection to Datalab is now open and will '
-              'remain until this command is killed.\n\n')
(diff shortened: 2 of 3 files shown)


tools/cli/datalab.py
```diff
@@ -224,7 +224,8 @@ def run():
         _SUBCOMMANDS[args.subcommand]['run'](
             args, gcloud_compute,
             gcloud_repos=gcloud_repos,
-            email=get_email_address())
+            email=get_email_address(),
+            in_cloud_shell=('DEVSHELL_CLIENT_PORT' in os.environ))
```
yebrahim: is using an environment var reliable enough? why don't we also make sure the VM metadata server is reachable?

ojarjur: > is using an environment var reliable enough? 

I think it is.

This seems to be the established pattern for detecting running in the Cloud Shell.

In particular, this is the pattern that is followed by the [oauth2client](https://github.com/google/oauth2client/blob/master/oauth2client/contrib/devshell.py) library.

The corresponding concept in the oauth2client library would be raising a 'NoDevshellServer' error, and that only gets raised if the environment variable is not set.

PR #1213, opened 2017-02-24
Switch the default image tag to be "latest" instead of "local".

ojarjur:
We were previously not using the "latest" tag because that had
been used by oure beta1 release, which ran in App Engine.

However, that has been deprecated for so long that we can now
take back the "latest" tag.

This change updates the default for the CLI and the release
script.

The release script will continue to also update the "local" tag,
so that anyone using it can still get updates.


chmeyers:
Please make sure you create one with the latest tag.

PR #1219, opened 2017-02-28
Install inception and structured data packages in container. 

qimingj:
Also change the docs dir to "ga" and pull notebooks from ga branch.

### DIFF
--- a/containers/datalab/content/run.sh
+++ b/containers/datalab/content/run.sh
@@ -206,20 +206,20 @@ mkdir -p /content/datalab/notebooks
 
 # Fetch docs and tutorials. This should not abort startup if it fails
 {
-if [ -d /content/datalab/docs ]; then
+if [ -d /content/datalab/ga ]; then
   # The docs directory already exists, so we have to either update or initialize it as a git repository
   pushd ./
-  cd /content/datalab/docs
-  if [ -d /content/datalab/docs/.git ]; then
-    git fetch origin master; git reset --hard origin/master
+  cd /content/datalab/ga
+  if [ -d /content/datalab/ga/.git ]; then
+    git fetch origin ga; git reset --hard origin/ga
   else
     git init; git remote add origin https://github.com/googledatalab/notebooks.git; git fetch origin; 
   fi
   popd
 else
-  (cd /content/datalab; git clone -n --single-branch https://github.com/googledatalab/notebooks.git docs)
+  (cd /content/datalab; git clone -n --branch ga --single-branch https://github.com/googledatalab/notebooks.git ga)
 fi
-(cd /content/datalab/docs; git config core.sparsecheckout true; echo $'intro/\nsamples/\ntutorials/\n*.ipynb\n' > .git/info/sparse-checkout; git checkout master)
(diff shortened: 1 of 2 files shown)


containers/datalab/content/run.sh
```diff
@@ -206,20 +206,20 @@ mkdir -p /content/datalab/notebooks
 
 # Fetch docs and tutorials. This should not abort startup if it fails
 {
-if [ -d /content/datalab/docs ]; then
+if [ -d /content/datalab/ga ]; then
```
ojarjur: I don't understand the reasoning for changing the directory name.

containers/datalab/content/run.sh
```diff
@@ -206,20 +206,20 @@ mkdir -p /content/datalab/notebooks
...
   # The docs directory already exists, so we have to either update or initialize it as a git repository
   pushd ./
-  cd /content/datalab/docs
-  if [ -d /content/datalab/docs/.git ]; then
-    git fetch origin master; git reset --hard origin/master
+  cd /content/datalab/ga
+  if [ -d /content/datalab/ga/.git ]; then
+    git fetch origin ga; git reset --hard origin/ga
```
ojarjur: I wonder if we should make the branch an ENV entry in the Dockerfile.

containers/datalab/content/run.sh
qimingj: Without changing the directory, we need to deal with the case where the docs dir is checked out to master (since almost all of our dev machine has ~/datalab/docs and is checked out to master), and we probably to fetch ga, checkout ga, and then rest local ga. It will be different from the release, where it simply pulls all notebook from master. Compared to adding an extra command, it seems doing it in a different dir is "closer" to the final release. No?

Also, if we want to keep master clean, we can create a new branch of datalab that pulls notebooks from ga branch. I am trying to avoid an extra branch though, but that means master from now on will pull notebooks from ga until notebooks are merged back to master.

containers/datalab/content/run.sh
qimingj: That is a great idea. Would it work with Datalab CLI? Does the CLI tool support passing extra params? If so, we should certainly do that to avoid the issue above.

containers/datalab/content/run.sh
qimingj: Just tried it without changing the directory. Here is what happened:

1. I have my old ~/datalab/docs/ dir, which is a git clone with --single-branch of master.
2. Now it tries to run "git fetch origin ga; git reset --hard origin/ga;" on Datalab startup from docs dir. It fails on "git reset --hard origin/ga" command, with "fatal: ambiguous argument 'origin/ga': unknown revision or path not in the working tree."

I'll need to investigate why but it might be because of the --single-branch when the repo was cloned. This is why I think a new dir is better -- it saves us efforts in dealing with switching to a new branch, which we don't have to after release.

qimingj: The problem is indeed --single-branch. We need to add a command:
git config remote.origin.fetch "+refs/heads/ga:refs/remotes/origin/ga"
to be able to fetch and then switch to ga branch locally.

containers/datalab/content/run.sh
ojarjur: Unfortunately, no, the CLI does not support passing in arbitrary environment variables.

qimingj: Cancel this. I will send two changes.

PR #1225, opened 2017-02-28
Fix visuals for ipywidgets Widgets menu.

jimmc:
Commenting out the height line makes the Widgets menu vertically aligned with the other menus. Probably not the best solution, but seems to work.

### DIFF
--- a/sources/web/datalab/static/datalab.css
+++ b/sources/web/datalab/static/datalab.css
@@ -215,14 +215,14 @@ div.btn-group a {
 div.btn-toolbar {
   margin: 0px;
 }
-div.btn-toolbar .toolbar-btn {
+div.btn-toolbar .toolbar-btn, #help_menu_container .dropdown-toggle {
   background-color: transparent;
   border: none;
   font-family: 'Open Sans', Helvetica, sans-serif;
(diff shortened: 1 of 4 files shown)


sources/web/datalab/static/datalab.css
```diff
@@ -215,14 +215,14 @@ div.btn-group a {
...
-div.btn-toolbar .toolbar-btn {
+div.btn-toolbar .toolbar-btn, #help_menu_container .dropdown-toggle {
   background-color: transparent;
   border: none;
   font-family: 'Open Sans', Helvetica, sans-serif;
   font-weight: normal;
   font-size: 14px;
   outline: solid 1px transparent !important;
-  height: 46px;
+  /* height: 46px; */
```
ojarjur: Was this just temporarily commented out for testing?

We should not have commented out lines in the final commit.

jimmc: I left this in in hopes someone else might make an alternate suggestion.

At this point it looks like we're not going to merge this branch, so I will wait until later (if ever) to clean it up.

jimmc:
Abandoning this PR in favor of #1232.

Issue #1206, opened 2017-02-24
Git button should point to the repo


nikhilk:
It should always open ungit to the top of where the source repo is located - not to the current folder.

In the current approach, when I am on the root directory in the file view, and open ungit, I am prompted with an ungit UI that says I need to setup a repo. We want users to just use the cloud source repo we clone.

yebrahim:
I'm not sure https://github.com/googledatalab/datalab/issues/1231 actually fixes this. Do we want users to always go to the cloud source repo we clone, or the repo root of whatever dir they're on when they click the ungit icon? I'd say the latter makes more sense.

jimmc:
Discussed with yebrahim. Going to the root of the repo they are in will require a more extensive change that includes a message to the service to figure out where that root is. We agreed that, if we want to make that change, we will open a new issue for that.

PR #1231, opened 2017-03-01
Always open ungit at /content/datalab/notebooks.

jimmc:
Fix for #1206 
This assumes a constant path for the repo location; it does not pull 'datalab' from the current location.

### DIFF
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -293,8 +293,8 @@ function initializePage(dialog, saveFn) {
     });
     $('#ungitButton').click(function() {
       // Always open at the root of the notebooks repository
-      path = '/content/datalab/notebooks';
-      prefix = window.location.protocol + '//' + window.location.host;
+      const path = '/content/datalab/notebooks';
+      const prefix = window.location.protocol + '//' + window.location.host;
 
       window.open(prefix + '/_proxy/8083/#/repository?path=' + path);
     });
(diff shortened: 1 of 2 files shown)


sources/web/datalab/static/datalab.js
```diff
@@ -292,16 +292,8 @@ function initializePage(dialog, saveFn) {
...
-      // Extract the current directory name
-      if (path.indexOf('/tree') == 0) {
-        path = path.substr('/tree'.length);
-      } else if (path.indexOf('/notebooks') == 0) {
-        path = path.substr('/notebooks'.length);
-        path = path.substr(0, path.lastIndexOf('/'));
-      }
-      path = '/content' + path;
+      // Always open at the root of the notebooks repository
+      path = '/content/datalab/notebooks';
```
ojarjur: This should still include "var " to declar the `path` variable.

sources/web/datalab/static/datalab.js
```diff
@@ -292,16 +292,8 @@ function initializePage(dialog, saveFn) {
...
       prefix = window.location.protocol + '//' + window.location.host;
```
ojarjur: This should have a "var" to declare the variable.

jimmc: Fixed.

sources/web/datalab/static/datalab.js
jimmc: Fixed.

PR #1265, opened 2017-03-13
Don't convert line to cell magics

yebrahim:
This code was added for the sake of convenience, but it's causing undesired behavior. For example, the following code doesn't work properly:
```
%timeit -n 1 print('line1')
print('line2')
```
Because it's getting converted to a cell magic, which means consecutive code will be part of the cell.

This fixes https://github.com/googledatalab/pydatalab/issues/71.

### DIFF
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -847,27 +847,6 @@ function initializeNotebookApplication(ipy, notebook, events, dialog, utils) {
     };
   });
 
-  require(["services/kernels/kernel"], function(ipy) {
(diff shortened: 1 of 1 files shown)


chmeyers:
I'm assuming this is non-backwards compatible and will break any notebook that uses line magics as if they are cell magics?

nikhilk:
Can we just say write this?

    %timeit -n 1
    print('line1')

yebrahim:
Yes this is a breaking change, how much do we want the convenience of using one `%` to mean two?

yebrahim:
Closing this now until we discuss this more.

PR #1269, opened 2017-03-15
Don't let the `datalab` tool manage non-datalab instances

### LINKED ISSUE #1268: Make the `datalab delete` command error out if the instance was not created by `datalab create`
Right now, the `datalab delete` command just forwards to `gcloud compute instances delete`.

As such, someone could erroneously run this against an arbitrary VM. We should add some checks so that scenario results in an error message rather than performing the delete.

ojarjur:
With this change, every command that is meant to operate on an
existing instance (connect, stop, delete), will raise an error
if the user specifies an instance that was not created by the
datalab command line tool.

This fixes #1268

### DIFF
--- a/tools/cli/commands/utils.py
+++ b/tools/cli/commands/utils.py
@@ -155,7 +155,7 @@ def _check_datalab_tag(instance, tags):
       InvalidInstanceException: If the check fails.
     """
     items = tags.get('items', [])
-    if not 'datalab' in items:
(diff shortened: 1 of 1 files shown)


ojarjur:
@yebrahim You can manually circumvent the check by adding a `datalab` tag to your old instance.

yebrahim:
Good point!

PR #1270, opened 2017-03-15
Add .gitignore and README.txt when creating notebook repo.

### LINKED ISSUE #1209: Git repo setup
When creating a repo, it would be a nice-to-have to have it be initialized with some starter content, esp. with a .gitignore that filters notebook checkpoints.

jimmc:
Fixes #1209. Adds a two-line .gitignore and a one-line README.txt file to new notebook repos.

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -59,6 +59,26 @@ clone_repo() {{
     gcloud source repos clone {1} /content/datalab/notebooks
 }}
 
+repo_is_populated() {{
+  cd ${{MOUNT_DIR}}/content/datalab/notebooks
+  git show-ref --quiet
+}}
+
+populate_repo() {{
+  echo "Populating datalab-notebooks repo"
+  docker run --rm -v "${{MOUNT_DIR}}/content:/content" \
+    --workdir=/content/datalab/notebooks \
+    --entrypoint "/bin/bash" {0} -c "\
+        echo '.ipynb_checkpoints' >> .gitignore; \
+        echo '*.pyc' >> .gitignore; \
+        echo '# Project Notebooks' >> README.md; \
+        git add .gitignore README.md; \
+        git -c user.email=nobody -c user.name=Datalab \
+          commit --message='Set up initial notebook repo.'; \
+        git push origin master; \
+    "
+}}
+
 format_disk() {{
   echo "Formatting the persistent disk"
   mkfs.ext4 -F \
@@ -66,6 +86,9 @@ format_disk() {{
     ${{PERSISTENT_DISK_DEV}}
   ${{MOUNT_CMD}}
   clone_repo
+  if ! repo_is_populated; then
+    populate_repo
+  fi
 }}
 
 mount_and_prepare_disk() {{
@@ -457,6 +480,9 @@ def create_repo(args, gcloud_repos, repo_name):
     Raises:
       subprocess.CalledProcessError: If the `gcloud` command fails
     """
+    if utils.print_info_messages(args):
+        print('Creating the repository {0}'.format(repo_name))
+
     create_cmd = ['create', repo_name]
     utils.call_gcloud_quietly(args, gcloud_repos, create_cmd)
 


tools/cli/commands/create.py
```diff
@@ -447,9 +447,53 @@ def create_repo(args, gcloud_repos, repo_name):
...
+    utils.call_gcloud_quietly(args, gcloud_repos, clone_cmd)
+
+    # Add a couple of files
+    if utils.print_info_messages(args):
+      print('Adding files to repo using local dir {0}'.format(datalab_dir))
+    with open(datalab_dir + '/.gitignore', 'w') as f:
+      f.write('.ipynb_checkpoints\n')
+      f.write('*.pyc\n')
+
+    with open(datalab_dir + '/README.txt', 'w') as f:
```
nikhilk: Can you make this readme.md for better rendering on github?

jimmc: Done.

tools/cli/commands/create.py
```diff
@@ -447,9 +447,53 @@ def create_repo(args, gcloud_repos, repo_name):
...
 
+    # Clone the repo into a local scratch directory
+    datalab_dir = tempfile.mkdtemp(prefix='datalab-')
+    clone_cmd = ['clone', repo_name, datalab_dir]
+    utils.call_gcloud_quietly(args, gcloud_repos, clone_cmd)
+
+    # Add a couple of files
+    if utils.print_info_messages(args):
+      print('Adding files to repo using local dir {0}'.format(datalab_dir))
+    with open(datalab_dir + '/.gitignore', 'w') as f:
```
ojarjur: This should use `os.path.join` so that it works on Windows as well.

tools/cli/commands/create.py
```diff
@@ -447,9 +447,53 @@ def create_repo(args, gcloud_repos, repo_name):
...
+      f.write('*.pyc\n')
+
+    with open(datalab_dir + '/README.md', 'w') as f:
+      f.write('# Project Notebooks\n')
+
+    # Commit the new files and push the changes back to the master
+    if utils.print_info_messages(args):
+      print('Commit changes and push repo')
+    try:
+      cmd = 'cd {0} && git add .gitignore README.md'.format(datalab_dir)
```
ojarjur: This might be better passing in the directory as a the `cwd` keyword argument to the subprcess invocation.

ojarjur:
Please check that this works on windows before we submit it.

yebrahim:
We should add the `flake8` command to the build script. We should also add it to the pydatalab repo.

tools/cli/commands/create.py
jimmc: Done.

tools/cli/commands/create.py
jimmc: Done.

tools/cli/commands/create.py
```diff
@@ -447,9 +447,56 @@ def create_repo(args, gcloud_repos, repo_name):
...
+    # Clone the repo into a local scratch directory
+    datalab_dir = tempfile.mkdtemp(prefix='datalab-')
+    clone_cmd = ['clone', repo_name, datalab_dir]
+    utils.call_gcloud_quietly(args, gcloud_repos, clone_cmd)
+
+    # Add a couple of files
+    if utils.print_info_messages(args):
+      print('Adding files to repo using local dir {0}'.format(datalab_dir))
+    with open(os.path.join(datalab_dir, '.gitignore'), 'w') as f:
+      f.write('.ipynb_checkpoints\n')
```
yebrahim: I think this needs an asterisk (`*.ipynb...`)?

nikhilk: Just .ipynb_checkpoints works, because that is the name of the directory.

jimmc: .ipynb_checkpoints is a directory. This line matches the directory name, so everything under that directory gets ignored.

tools/cli/commands/create.py
```diff
@@ -447,9 +447,56 @@ def create_repo(args, gcloud_repos, repo_name):
...
     create_cmd = ['create', repo_name]
     utils.call_gcloud_quietly(args, gcloud_repos, create_cmd)
 
+    # Clone the repo into a local scratch directory
+    datalab_dir = tempfile.mkdtemp(prefix='datalab-')
+    clone_cmd = ['clone', repo_name, datalab_dir]
+    utils.call_gcloud_quietly(args, gcloud_repos, clone_cmd)
+
+    # Add a couple of files
+    if utils.print_info_messages(args):
```
ojarjur: I think the first message about creating the repo deserves to be 'info' level, but the rest of them feel more like debugging messages than info.

Consider using `if args.verbosity == 'debug':` instead of `if utils.print_info_messages(args):`

jimmc:
I was having too much trouble trying to get the old code to run on all platforms, so I rewrote it to populate the repo from the container rather than the machine running the datalab command.

jimmc:
Tested successfully on linux, windows, and mac osx.
Please review the new code.
Note that I had to specify an email and user name for the initial commit, so please take a look at those values.

PR #1281, opened 2017-03-18
Map the entire /content directory to the disk instead of just /conten…

### LINKED ISSUE #1280: The entire '/content' directory should be mapped to the notebooks disk
The current version of the `datalab` command line tool creates instances where the `/content/datalab` directory is mapped to the `datalab` subdirectory of the notebooks disk.

However, there are two issues with this:

1. Why even have the `/content` directory if it isn't writeable in a persistent way?
2. We then loose (with caveats) any data that a user writes to the `/content` directory.

The caveat on the second item is that even though only the `/content/datalab` directory is persisted, the entire `/content` directory is backed up to GCS.

This means we aren't being consistent with our own behavior.

I think the appropriate solution is to map the entire `/content` directory to the notebooks persistent disk.

However, this requires backwards compatibility with existing notebook disks.

To address that, I am thinking of the following logic for newly created instances:

1. After mounting the notebook disk, check if there is a `content` subdirectory on it.
2. If there isn't, then create that directory.
3. If there is a `datalab` directory in the notebook disk, but no `content/datalab` subdirectory, then move the `datalab` directory under the `content` directory.
4. Otherwise, if there is no `datalab` directory under the `content` directory, then create one.
5. When launching the Docker container, always map the `content` directory of the notebooks disk to the `/content` directory in the container.

ojarjur:
…t/datalab

This change only affects new instances, but it is backwards compatible
with existing notebook disks.

This fixes #1280

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -78,7 +78,7 @@ mount_and_prepare_disk() {{
   old_dir="${{MOUNT_DIR}}/datalab"
   new_dir="${{MOUNT_DIR}}/content/datalab"
   if [ -d "${{old_dir}}" ] && [ ! -d "${{new_dir}}" ]; then
-    echo "Copying ${{old_dir}} to ${{new_dir}}"
+    echo "Moving ${{old_dir}} to ${{new_dir}}"
     mv "${{old_dir}}" "${{new_dir}}"
   else
     echo "Creating ${{new_dir}}"
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -52,9 +52,9 @@ MOUNT_CMD="mount -o discard,defaults ${{PERSISTENT_DISK_DEV}} ${{MOUNT_DIR}}"
 
 clone_repo() {{
   echo "Creating the datalab directory"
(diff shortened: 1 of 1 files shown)


tools/cli/commands/create.py
```diff
@@ -68,12 +68,22 @@
...
   mkdir -p "${{MOUNT_DIR}}"
   ${{MOUNT_CMD}} || format_disk
   chmod a+w "${{MOUNT_DIR}}"
-  mkdir -p ${{MOUNT_DIR}}/datalab
+  mkdir -p "${{MOUNT_DIR}}/content"
+
+  old_dir="${{MOUNT_DIR}}/datalab"
+  new_dir="${{MOUNT_DIR}}/content/datalab"
+  if [ -d "${{old_dir}}" ] && [ ! -d "${{new_dir}}" ]; then
+    echo "Copying ${{old_dir}} to ${{new_dir}}"
```
jimmc: This should probably say "Moving" or "Renaming" rather than "Copying".

ojarjur: Good catch; fixed.

Issue #1051, opened 2016-11-16
Feature Request: Add ability to search and navigate to a specific notebook


parthea:
It would be interesting to have a search text box in Datalab to help navigate to specific notebooks when the directory structure is very deep (lot's of clicking!). 

I am imagining a text box with autocomplete functionality. If I type in part of a notebook name, the autocomplete dropdown menu will show me notebooks that match my search pattern. Clicking on the autocomplete item would navigate to the corresponding notebook.

This is just a suggestion! Please feel free to close this issue if this is not helpful.

VelizarVESSELINOV:
Today using GitHub to search to find a specific file or content inside the ipynb files, the power of Google search is missing inside Datalab.

PR #1282, opened 2017-03-20
Add "Jump to file" functionality

yebrahim:
This PR adds a "Jump to file" input box in the file list UI that shows a list of files matching the entered substring to help with quick navigation.

Files on the user's content directory are indexed at the start of the webserver and cached in a global file index with a cap of a million files, which are then filtered as the user types a substring. Each 500ms, a GET xhr request is sent to the server containing the characters typed by the user to filter files. Only 20 search results are sent back to the client.

[Chokidar](https://github.com/paulmillr/chokidar) (MIT) is used to index and watch the file system for changes.

For now, exact match of substrings is used, next we can look into making search fuzzy.

I also modified tests to start 10 seconds after starting the datalab server, this helps fix an issue when indexing slows down the server for the first 30 seconds or so (although it still responds to GET requests), which was causing tests to fail.

image

When more than 20 matches are found, a message is shown saying only 20 are shown:
image

When there are too many files (> 1M) to index on the server, the find functionality is disabled:
image

Fixes https://github.com/googledatalab/datalab/issues/1051.


sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,106 @@
...
+var appSettings: common.Settings;
+var fileIndex: string[] = [];
+
+/**
+ * Implements the file search/filter request handling.
+ * @param request the incoming search request.
+ * @param response the outgoing search response.
+ */
+function requestHandler(request: http.ServerRequest, response: http.ServerResponse): void {
+  var parsedUrl = url.parse(request.url, true);
```
jimmc: Please use const for fields that will not be updated.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,106 @@
...
+/// <reference path="common.d.ts" />
+
+import http = require('http');
+import url = require('url');
+import fs = require('fs');
+import path = require('path');
+import logging = require('./logging');
+
+var appSettings: common.Settings;
+var fileIndex: string[] = [];
```
jimmc: Probably doesn't matter at the file level, but in general I prefer seeing let and const rather than var.

sources/web/datalab/static/datalab.css
```diff
@@ -469,6 +469,46 @@ div.cellPlaceholder {
...
+#find-file-container {
+  padding: 2px;
+  font-weight: normal;
+}
+#tree-filter {
+  margin: 0px;
+  padding: 5px;
+}
+.ui-autocomplete {
+  position: absolute !important;
```
jimmc: Can you use a more specific selector rather than 'important'? Using 'important' can make things more difficult later.

sources/web/datalab/static/datalab.js
```diff
@@ -1366,8 +1366,38 @@ function initializeNotebookList(ipy, notebookList, newNotebook, events, dialog,
...
+
   })();
 
+  searchDiv = $('#tree-filter');
+  searchDiv.autocomplete({
+    appendTo: '.tree-filter-complete',
+    position: { of: '.tree-filter-complete' },
+    source: (request, response) => {
+
+      fileSearchPath = 'http://localhost:8081/_filesearch?';
```
jimmc: Should this be using location.protocol and location.host?

sources/web/datalab/templates/tree.html
```diff
@@ -114,6 +114,11 @@
                           </li>
                         </ul>
                       </div>
+                      <div id="find-file-container" class="btn-group pull-right">
+                        <input id="tree-filter" class="form-control" placeholder="Jump to file" />
+                        <div class="tree-filter-complete"></div>
+                      </div>
+                    </div>
```
jimmc: This looks to me like an extra &lt;/div&gt; line.

sources/web/datalab/fileSearch.ts
yebrahim: Yes, I'll keep the global vars, everything else is fixed to `let`/`const`.

sources/web/datalab/static/datalab.css
yebrahim: Narrower selector wouldn't help here unfortunately , these are overriding inline styles.

sources/web/datalab/static/datalab.js
yebrahim: Done.

sources/web/datalab/templates/tree.html
yebrahim: Done.

yebrahim:
I fixed the implementation to use an async function to index the files so the server isn't stuck in initialization. I'll work on showing an indicator when too many files exist instead of returning a partial result.

yebrahim:
Now added a div that says "Showing top 20 results" that appears at the top of the autocomplete list (see images in description) if 20 results are returned from server, which means it's likely there are more.

This is now ready for review.

nikhilk:
Not sure why we're showing the count of results at the top... at most its something shown at the bottom, so it doesn't interfere with the top result, but why do we need it at the bottom either?

And I thought we were switching to doing things on the server, so it doesn't matter how many matches there are -- only the top N are returned to the client.

yebrahim:
I could move it to bottom, I don't have a preference.

Yes we're doing things on the server, but we wanted an indication to the user that there are probably more than the list of results shown, since we're only showing 20 max, so they can maybe narrow down their search.

nikhilk:
If the searching is done on the server, then why does one of the screenshots say "Find disabled. Too many files"?

yebrahim:
I have a one million cap on the file indexing. If the user has more than that, I wouldn't want the server to go spinning for a long time before serving requests, but I don't want it to return partial results either. This looked like the safest option, considering one million files in the content disk is unlikely.

nikhilk:
Agree, with the specific # one is unlikely to hit the special case, but even if you do, you could search over the first 1M. The threshold number could likely be much smaller.

Secondly, from glancing at the code, I didn't see where this index gets invalidated - for example in response to adds/deletes/renames. Maybe I missed it...

yebrahim:
You're right, I missed that. We'll have to think this one through, since the file tree can be modified in many ways. We'll probably have to rebuild the index periodically or use a filesystem watch package.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,143 @@
...
+  });
+}
+
+/**
+ * Recursively indexes all files under the given search path into the global fileIndex object
+ * @param searchpath the path prefix to start indexing
+ */
+function index(searchpath: string, callback: Function): void {
+  // stop the search if we've reached the file count limit
+  if (fileIndex.length >= fileCountLimit) {
```
jimmc: I don't think your limit logic is correct. This only checks at the beginning of a directory, but while scanning a directory it will continue to push files onto the list without checking its size.
Maybe test this code by setting fileCountLimit to a small number like 10 and see what happens.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,143 @@
...
+      logging.getLogger().info('Finished indexing ' + fileIndex.length + ' files in ' + indexTime[0] + ' seconds');
+    }
+  });
+}
+
+/**
+ * Recursively indexes all files under the given search path into the global fileIndex object
+ * @param searchpath the path prefix to start indexing
+ */
+function index(searchpath: string, callback: Function): void {
```
jimmc: Consider implementing a generic directory walking function, then calling it with a callback that processes one file and determines when to stop walking.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,143 @@
...
+    });
+  });
+}
+
+/**
+ * Filters the file index based on the provided pattern
+ * @param pattern the search pattern
+ * @returns a list of matches that are superstrings of pattern
+ */
+function filter(pattern: string): string[] {
```
yebrahim: I don't see an issue with the current naming or implementation. It's scoped to the fileSearch file, so the simple `filter` name should do.
Turning it to a pure function is a good idea though.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Builds an index of all files in the content directory to make search faster
+ */
+export function indexFiles(): void {
+  const startTime = process.hrtime();
+  index(appSettings.contentDir + '/', () => {
+    const indexTime = process.hrtime(startTime);
+    if (tooManyFiles) {
+      logging.getLogger().error('Hit file index maximum limit, searching files should be disabled');
```
yebrahim: It is a suggestion. Not to the user, but to the frontend, that it should disable searching.

There is no easy way for the nodejs server to affect what the frontend shows to the user, so it has to be merely a suggestion that the frontend acts upon.
In this vein, the limit number doesn't add much since it's just a log message.

sources/web/datalab/fileSearch.ts
yebrahim: I'm moving away from this implementation to using an external package that watches fs events.

sources/web/datalab/fileSearch.ts
yebrahim: You're right, we needed another check when adding files. I'm discarding this function though.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,143 @@
...
+      return callback();
+    }
+    let remaining = list.length;
+    if (remaining === 0) {
+      return callback();
+    }
+    list.forEach((file) => {
+      // ignore hidden files/dirs
+      if (file[0] === '.') {
+        if (--remaining === 0) {
```
yebrahim: Not really. The body of the `forEach` is asynchronous. If I just call the callback after the loop it'll be called immediately, not when the actual indexing is done.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,126 @@
...
+import url = require('url');
+import fs = require('fs');
+import path = require('path');
+import logging = require('./logging');
+import chokidar = require('chokidar');
+
+let appSettings: common.Settings;
+let fileIndex: string[] = [];
+let indexReady: boolean = false;
+let tooManyFiles: boolean = false;
```
jimmc: It looks like this is no longer being set anywhere.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,126 @@
...
+    })
+    .on('raw', () => {
+      if (!indexReady) {
+        indexReady = true;
+        const indexTime = process.hrtime(startTime);
+        logging.getLogger().info('Finished indexing ' + fileIndex.length + ' files in ' + indexTime[0] + ' seconds');
+        logging.getLogger().error('Indexing threw raw event');
+      }
+    })
+    .on('error', () => {
```
jimmc: It looks like there is no error flag for the client to tell that an error occurred. If the scan gets an error after scanning 10 files out of 1000, will that look the same as a successful scan of 10 total files?

sources/web/datalab/fileSearch.ts
yebrahim: I use it now to disable the finder on the frontend.

sources/web/datalab/fileSearch.ts
yebrahim: Yes. This is there to signal when indexing is done. We're not catching errors except for permission errors, which we're ignoring anyway.

yebrahim:
I rebased and pushed to get the test flakiness fixed. This is now ready for review.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,132 @@
...
+        fileIndex.splice(pos, 1);
+      }
+    })
+    .on('ready', () => {
+      indexReady = true;
+      const indexTime = process.hrtime(startTime);
+      logging.getLogger().info('Finished indexing ' + fileIndex.length + ' files in ' + indexTime[0] + ' seconds');
+    })
+    .on('raw', () => {
+      if (!indexReady) {
```
jimmc: Why do you test indexReady here, but not for on('ready')?
Based on the code, I can make guesses as to which of these message types might be received more than once, but it might be nice to add some comments to that effect. Then maybe I wouldn't have to ask the above question.

sources/web/datalab/fileSearch.ts
jimmc: If you get an error, you will be returning partial results, so I think you should handle this similar to how you handle "still indexing" and "too many files".
I don't understand your comment about ignoring permission errors. Does this code get executed on permission errors, or only on other kinds of errors?

sources/web/datalab/fileSearch.ts
yebrahim: Sure I can add a comment.

sources/web/datalab/fileSearch.ts
yebrahim: I'm not sure what kind of errors we can get with this, and I don't think we should disable the entire indexing if a file errored out for some reason. We're already ignoring permission errors, which can result from indexing inaccessible files.

sources/web/datalab/fileSearch.ts
```diff
@@ -0,0 +1,139 @@
...
+        indexReady = true;
+        const indexTime = process.hrtime(startTime);
+        logging.getLogger().info('Finished indexing ' + fileIndex.length + ' files in ' + indexTime[0] + ' seconds');
+        logging.getLogger().error('Indexing threw raw event');
+      }
+    })
+    // It's unclear why this event might get fired since permission errors are ignored,
+    // it's here to make sure indexing singals on complete, even if an error occurred
```
jimmc: signals

Issue #333, opened 2015-05-13
Remove some public APIs from pygcp


gramster:
Some of these are just to support unit tests and could be hidden. E.g.

job()
query_job()
dataset_name
table_name

PR #1284, opened 2017-03-21
Improve the user experience around accessing help information and notebook navigation

parthea:
It took a few seconds to locate the help documentation in the Datalab app bar (hover the icons until I found 'Help links').

This PR changes the icon for the link titled 'Help links'  to use the 'help_outline' icon instead of 'open in new' from https://material.io/icons/.

I've included a screenshot from my local environment.

Feel free to close if the 'open in new' icon is preferred.

datalabscreenshot

### DIFF
--- a/sources/web/datalab/static/appbar.html
+++ b/sources/web/datalab/static/appbar.html
@@ -36,16 +36,16 @@
         <i class="material-icons">help_outline</i>
       </button>
       <ul class="dropdown-menu dropdown-menu-right" style="width: 200px">
-        <li><a href="/tree/datalab/docs/Readme.ipynb" target="_blank"><span>Getting Started</span></a></li>
-        <li><a href="/tree/datalab/docs" target="_blank"><span>Samples and Tutorials</span></a></li>
-        <li><a href="http://googledatalab.github.io/pydatalab/" target="_blank"><span>Datalab API</span></a></li>
-        <li class="divider"></li>
+        <li><a id="showHelpLink" style="display:none" href="#">Show Help</a></li>
         <li><a id="keyboardHelpLink" style="display:none" href="#">Keyboard Shortcuts</a></li>
         <li><a id="markdownHelpLink" style="display:none" href="#">Markdown Syntax</a></li>
         <li id="notebookHelpDivider" class="divider" style="display:none"></li>
-        <li><a href="https://cloud.google.com/bigquery/what-is-bigquery" target="_blank"><span>BigQuery&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span></a></li>
-        <li><a href="https://cloud.google.com/storage/docs/overview" target="_blank"><span>Cloud Storage&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</span></a></li>
+        <li><a href="/tree/datalab/docs/Readme.ipynb" target="_blank"><span>Samples and Tutorials</span></a></li>
         <li class="divider"></li>
+        <li><a href="https://cloud.google.com/bigquery/what-is-bigquery" target="_blank"><span>BigQuery</span></a></li>
+        <li><a href="https://cloud.google.com/storage/docs/overview" target="_blank"><span>Cloud Storage</span></a></li>
+        <li class="divider"></li>
+        <li><a href="http://googledatalab.github.io/pydatalab/" target="_blank"><span>Datalab API</span></a></li>
         <li><a href="https://developers.google.com/chart/" target="_blank"><span>Google Charts</span></a></li>
         <li><a href="http://matplotlib.org/contents.html" target="_blank"><span>Matplotlib</span></a></li>
         <li><a href="http://docs.scipy.org/doc/numpy/reference/" target="_blank"><span>Numpy</span></a></li>
@@ -58,7 +58,7 @@
       <button id="accountDropdownButton" title="Account" class="toolbar-btn">
         <i class="material-icons">account_circle</i>
       </button>
-      <div class="dropdown-menu dropdown-menu-right" style="min-width: 250px;">
+      <div class="dropdown-menu dropdown-menu-right" style="min-width: 150px;">
         <div id="signInButton" style="display:none">
           <button title="Sign In" class="btn btn-success" style="width: 100%; float: none">Sign In</button>
         </div>
@@ -78,15 +78,9 @@
           <button id="stopVmButton" title="Stop VM" class="btn btn-danger" style="width: 100%; float: none">Stop VM</button>
         </div>
         <hr/>
-        <div id="settingsButton">
-          <button title="Settings" class="btn btn-default" style="width: 100%; float: none; border: 0">Settings</button>
-        </div>
-        <div id="aboutButton">
-          <button title="About Google Cloud Datalab" class="btn btn-default" style="width: 100%; float: none; border:0">About Datalab</button>
-        </div>
-        <div id="feedbackButton">
-          <button title="Provide feedback" class="btn btn-default" style="width: 100%; float: none; border:0">Feedback</button>
-        </div>
+        <li><a id="settingsButton" href="#">Settings</a></li>
+        <li><a id="aboutButton" href="#">About Datalab</a></li>
+        <li><a id="feedbackButton" href="#">Feedback</a></li>
       </div>
     </div>
   </div>
--- a/sources/web/datalab/static/appbar.html
+++ b/sources/web/datalab/static/appbar.html
@@ -13,11 +13,6 @@
   </span>
   <div class="btn-toolbar pull-right right-toolbar">
     <div class="btn-group">
-      <button id="settingsButton" class="toolbar-btn" title="Settings">
-        <i class="material-icons">settings</i>
-      </button>
-    </div>
-    <div class="btn-group">
       <button id="ungitButton" title="Open ungit on the notebooks repository" class="toolbar-btn">
         <span class="fa fa-code-fork"></span>
       </button>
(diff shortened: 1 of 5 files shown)


chmeyers:
I like the help_outline one.  I'm guessing we used 'open in new' because it pops open a menu and opens in a new tab?  I think that having the icon show what it does is more important than having it show how it does it, so I think we should switch.

+Yasser/Nikhil for opinion before merging.

yebrahim:
We chose that icon because we have a help icon in the notebook's toolbar that looks similar, and this one really is more like "help links" that open in separate tabs than direct help.

I think we should avoid having two help icons. Any ideas around that are welcome.

nikhilk:
We have a bit of icon proliferation going on...

Maybe we can combine the help button in the sidebar with this (and move the navigation button as well to the top) in order to have a single toolbar?

In that context, I think Settings and Info can be in the dropdown for the user icon... which should help making space for the navigation button, while also reducing overall icon clutter.

Thoughts? If you agree, it seems we could take this change, and then separately do that bit of icon/toolbar cleanup.

parthea:
@yebrahim My mistake. I didn't notice the 'help' icon was already used in the notebook toolbar. I agree it would be confusing to use the same icon twice. @nikhilk I do like the idea of merging the notebook help section with the help dropdown in the appbar.   My personal preference is to have a 'help' icon in the appbar as suggested.

In addition, I think it would be helpful to have _Getting Started_ as the link text in the help dropdown that points to https://github.com/googledatalab/notebooks/blob/master/Readme.ipynb, which is different from _Samples and Tutorials_ in the help dropdown which points to http://127.0.0.1:8081/tree/datalab/docs. 

The _Help for Python APIs_ content under the notebook help could potentially be removed altogether. The [Introduction to Python](https://github.com/googledatalab/notebooks/blob/master/intro/Introduction%20to%20Python.ipynb) notebook has sufficient content on  using `'?'` to invoke help on a class or a method.

If the task has not already been assigned and if it is ok with everyone, I would be interested in modifying this PR to incorporate the suggested changes.

notebookhelp

yebrahim:
I agree there's some work we can do to make toolbars better. However. merging the two icons and moving navigation to the top bar won't work for now, until we figure out a better redesign. The icons in the lower bar are specific to notebook functionality, navigation (bookmarks) for example does not make sense if you're not inside a notebook.

We could try to add the notebook-specific help content and bookmarks icon to the top bar conditionally (only if inside notebook), but right now the code is not structured well enough to make this easy as far as I can see. You're welcome to take a stab at it of course.

As far as moving info and settings icons under profile, I'm open to that.

parthea:
This is not ready for review yet. I'll comment once its ready.

parthea:
This is now ready for review. In addition to changing the icon for the help links , the following changes are also included in this PR:

- The help button and navigation buttons no longer appear in the notebook toolbar.
- Move the settings and info buttons to the dropdown for the user icon. See screenshot.
- Move the navigation button from the notebook toolbar to the appbar. Based on my testing the navigation button only appeared in the appbar when inside a notebook.
- Add a 'Getting Started' link under the help links drop down. See screenshot.
- Move the 'Getting Started', 'Samples and Tutorials' and 'Datalab API' links to the top of the help links drop down.
- Remove the 'toggle sidebar button' which would allow users to make the notebook navigation section larger. I did not find this toggle button useful in my testing with Datalab but I'm happy to revert this change if needed.


appbar
helplinksmenu

nikhilk:
Somewhat of a nit -- could you please make the items in the user dropdown left-aligned like the other menu?

I think the toggle on the sidebar is still useful, in order to close it. Its not easy to tell from screenshots, but I expect that menu entries (navigation and some from the help dropdown) cause the sidebar to expand when used, and the user can use the toggle to close the sidebar.

yebrahim:
I discussed with Nikhil, here's what we think the design should move to:
- "Show Help" at the top should retain the old "Help" button functionality, which opened the sidebar to initial help content, which changes if other help content is loaded. (This should be fixed ideally to always show that initial help content [here](https://github.com/googledatalab/datalab/blob/master/sources/web/datalab/templates/nb.html#L210), but this is a separate, bigger change)
- "Keyboard Shortcuts" and "Markdown Syntax" should show up directly afterwards in the same divider section
- "Samples and Tutorials" shows up next in a separate section, and it points to the "Readme.ipynb" file. This is in place of the Getting Started link in this PR, since that notebook really is a list of tutorials and samples.
- "Datalab API" should move to the beginning of the API section that has API docs for other components.

Here's a quick render of what we think it should look like:
image

Also please make sure the minimum body width ([here](https://github.com/googledatalab/datalab/blob/master/sources/web/datalab/static/datalab.css#L27)) and the maximum screen width for response design ([here](https://github.com/googledatalab/datalab/blob/master/sources/web/datalab/static/datalab.css#L529)) match the icon changes. I think at least the max width should be reduced if we're taking away icons.

I understand this is more work to do, so please let us know if you'd like us to pick it up, and I can help with that. :)

(btw, the PR title is a little off-topic now)

parthea:
@yebrahim Thanks for providing clarity on the work items. The screenshot is helpful! I will tackle this tonight and also incorporate the changes requested by @nikhilk .

yebrahim:
Awesome! Much appreciated @parthea.

parthea:
Thanks for providing detailed feedback. I think I've captured all of the comments. Please let me know if I missed anything. I'm happy to make further changes as required. 

> I think the toggle on the sidebar is still useful, in order to close it. Its not easy to tell from screenshots, but I expect that menu entries (navigation and some from the help dropdown) cause the sidebar to expand when used, and the user can use the toggle to close the sidebar.

I've replaced the toggle for the sidebar with a close button that sits in the sidebar. Please let me know if this is ok. Clicking on other the other help menu entries or the navigation button will automatically cause the sidebar to expand if closed (This functionality is unchanged). 
image


>"Show Help" at the top should retain the old "Help" button functionality, which opened the sidebar to initial help content, which changes if other help content is loaded. (This should be fixed ideally to always show that initial help content here, but this is a separate, bigger change)

I've updated the functionality so that clicking on "Show Help" will always show the old "Help" functionality rather than the last help content that was loaded.

Here are a couple screenshots to help with the review:

- Items in the user dropdown are left-aligned. 
image
- Resolve issue where the text in the user drop down was a different color (`'#000'`) compared to other links (`'#333'`) in the light scheme. Also fixed this for the modal window in the light scheme.
- Screenshot of the Help menu with the changes requested by @yebrahim .
image
- Clicking on 'Show Help' opens the sidebar with the old functionality. As previously described, I removed the toggle button and replaced it with a close button. I'm happy to revert this change if you prefer the toggle button. The reason I like the close button better is because users can quickly re-open the specific help using the links under the help menu. The toggle button will open the last used help which may not necessarily be what the user wants.
image
- Clicking on either 'Keyboard Shortcuts' , 'Markdown Syntax' , 'Navigation' or 'Show Help' will update the content in the side bar if the side bar is already open. The sidebar will automatically open if it is closed.
image

sources/web/datalab/static/appbar.html
```diff
@@ -30,37 +25,40 @@
       </a>
     </div>
     <div class="btn-group">
```
yebrahim: you need the `display:none` on the `btn-group` div, otherwise there will be some padding/margin showing for no reason.

sources/web/datalab/templates/nb.html
```diff
@@ -199,37 +180,16 @@
         </div>
         <div id="sidebarArea" style="display:none">
           <div id="sidebarContent">
-            <div class="btn-group">
-              <button id="toggleSidebarButton" type="button" class="toolbar-btn" title="Toggle the size of the sidebar">
-                <i class="material-icons">chevron_left</i>
-              </button>
+            <div>
+              <div class="btn-group pull-right">
+                <button id="hideSidebarButton" type="button" class="toolbar-btn" title="Close sidebar">
```
yebrahim: you'll need the `btn btn-default` classes here to style this correctly for both themes, instead of the `toolbar-btn`.
you might also need to adjust the padding for this element, the default button padding is too much for a small `X` button.

sources/web/datalab/static/appbar.html
```diff
@@ -30,37 +25,40 @@
...
-      <button id="aboutButton" title="About Google Cloud Datalab" class="toolbar-btn">
-        <i class="material-icons">info_outline</i>
-      </button>
-    </div>
     <div id="accountDropdown" class="btn-group">
       <button id="accountDropdownButton" title="Account" class="toolbar-btn">
         <i class="material-icons">account_circle</i>
       </button>
-      <div class="dropdown-menu dropdown-menu-right" style="min-width: 250px;">
+      <div class="dropdown-menu dropdown-menu-right" style="min-width: 150px;">
```
yebrahim: please keep this at 250px. I know it doesn't look ideal with left-aligned menu items, but 150 is too narrow.

sources/web/datalab/templates/nb.html
```diff
@@ -199,37 +180,16 @@
         </div>
         <div id="sidebarArea" style="display:none">
           <div id="sidebarContent">
-            <div class="btn-group">
-              <button id="toggleSidebarButton" type="button" class="toolbar-btn" title="Toggle the size of the sidebar">
-                <i class="material-icons">chevron_left</i>
-              </button>
+            <div>
+              <div class="btn-group pull-right">
```
yebrahim: this looks like it's superimposed on the sidebar content, this is likely because of the `pull-right` class, which floats it. we'll need to auto-overflow it's container div.

sources/web/datalab/static/datalab.css
```diff
@@ -526,7 +521,7 @@ div.modal-dialog pre, div.modal-dialog code, #help code, #help pre {
 }
 
 /* Responsive Design */
-@media screen and (max-width: 1270px) {
+@media screen and (max-width: 1230px) {
```
yebrahim: if we're removing all the buttons on the right side of the toolbar, i think we can afford less than this. i'd keep about 50px on top of the `maintoolbar`'s width and that should be enough.

parthea: Done.
1032px (`maintoolbar` width) + 48px = 1080px

sources/web/datalab/static/appbar.html
parthea: Done.

sources/web/datalab/static/appbar.html
parthea: Done.

sources/web/datalab/templates/nb.html
parthea: Done.

sources/web/datalab/templates/nb.html
parthea: Done but I'm not sure if I have `overflow: auto` in the right place. The 'X' button looks better now. It's possible that the issue is fixed by using `btn btn-default` instead of `toolbar-btn`.

parthea:
@yebrahim I have `'Allow edits from maintainers'` checked in GitHub for this pull request. Although I'm happy to make any changes as needed, you're able to make changes directly to my branch if that's easier.

Here is a description from GitHub:
> If checked, users with write access to googledatalab/datalab can add new commits to your update-help-links-icon branch.

yebrahim:
Made some fixes and brought back the chevrons for toggling sidebar and expanding it. Will merge this PR now, please open any issues you see with these changes, and we can take them separately.

image

PR #1287, opened 2017-03-25
Fix css cut-off issue on cell mini-toolbar

yebrahim:
Fixes https://github.com/googledatalab/datalab/issues/1063.

### DIFF
--- a/sources/web/datalab/static/datalab.css
+++ b/sources/web/datalab/static/datalab.css
@@ -115,12 +115,14 @@ body {
   justify-content: space-between;
 }
 #contentArea {
-  height: 100%;
   -webkit-order: 2;
   order: 2;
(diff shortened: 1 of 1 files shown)


sources/web/datalab/static/datalab.css
```diff
@@ -115,13 +115,15 @@ body {
   justify-content: space-between;
 }
 #contentArea {
-  height: 100%;
```
parthea: In my testing I found that the height of the notebook toolbar wasn't correct. The issue was resolved by restoring `height: 100%` under `#contentArea`.

notebooktoolbarissue

yebrahim: Interesting, it looks fine for me. I'm wondering what else am I missing when I test UI on my Chrome/Firefox then..

PR #1288, opened 2017-03-28
Remove support for kernel gateway backends.

### LINKED ISSUE #1146: Add endpoints with kernel gateway to Datalab
We are working on a project that relies on Jupyter notebooks using kernel gateway endpoints to make HTTP requests to it. Adding all the Datalab capabilities would tremendously improve our project, but we need to keep the endpoints live. It would be very helpful if you added kernel gateway compatibility to Datalab.

ojarjur:
This change removes the code that previously allowed Datalab to
run with a kernel gateway as its backend.

We were not able to provide a sufficiently reliable user experience
in that mode, so it was deprecated with our GA release.

Since it is deprecated, this change removes the code for supporting
it, and replaces it with an error message telling the user that it
is no longer supported. That error message also lists the last release
prior to GA in case the user does not wish to upgrade.

Since the kernel gateway image can no longer be used as the backend
for a Datalab instance, we would normally delete it. However, we
have had user requests to support running notebooks as HTTP servers
via kernel gateways, so there is another use case for them.

To that end, this change also adds the changes needed to fix #1146


containers/datalab/content/run.sh
```diff
@@ -14,107 +14,24 @@
...
-  fi
-  echo "Successfully established SSH tunnel to ${instance}"
+Please either switch to running all of Datalab in a VM via the `datalab` command line tool,
+or continue to use the unsupported image gcr.io/cloud-datalab/datalab:local-20170224
+'
 
-  watch_and_restart_tunnel "${project_id}" "${zone}" "${instance}" "${ssh_user}" &
-}
+ERR_UNSUPPORTED_GATEWAY_OPTION=1
+ERR_TMP_NOT_WRITABLE=2
```
jimmc: Is ERR_TMP_NOT_WRITABLE used anywhere? If so, is there any concern that changing this error code might break some script that is checking for specific error codes?

ojarjur: No, there is nothing that reads it.

This is just for debugging purposes if something exits unexpectedly.

containers/gateway/content/run.sh
```diff
@@ -25,21 +25,10 @@ then
...
-done
-
-# Start the proxy.
-FOREVER_CMD="forever --minUptime 1000 --spinSleepTime 1000"
-if [ -z "${DATALAB_DEBUG}" ]
-then
-  echo "Starting Datalab Kernel Gateway in silent mode, for debug output, rerun with an additional '-e DATALAB_DEBUG=true' argument"
-  FOREVER_CMD="${FOREVER_CMD} -s"
+# Start the Datalab kernel gateway.
+KG_COMMAND="jupyter kernelgateway --JupyterWebsocketPersonality.list_kernels=True --KernelGatewayApp.port=8080 --KernelGatewayApp.ip=0.0.0.0"
```
yebrahim: Shouldn't the port be 8081?

sources/web/build.sh
```diff
@@ -37,13 +37,3 @@ rsync -avp ./datalab/config/ $WEB_DIR/config
 rsync -avp ./datalab/static/ $WEB_DIR/static
 rsync -avp ./datalab/templates/ $WEB_DIR/templates
 rsync -avp ./datalab/package.json $WEB_DIR/package.json
-
-# Compile kernel proxy
-KERNELPROXY_DIR=$BUILD_DIR/web/kernelproxy
```
yebrahim: The kernelproxy directory is copied in the Dockerfile too

containers/datalab/content/run.sh
```diff
@@ -256,12 +96,6 @@ if [ -n "${vm_project}" ] && [ "${vm_project}" != "no-project-id" ]; then
    export VM_ZONE=$(curl -s "${compute_metadata_url}/instance/zone" -H "Metadata-Flavor: Google" | sed 's/.*zones\///')
 fi
 
-# Install the kernel gateway server extension, if a kernel gateway URL has been specified
-if [ -n "${KG_URL}" ]
```
yebrahim: We have more references in `jupyter.ts` that uses `KG_URL` to set up the server to connect to the gateway.

containers/datalab/content/run.sh
```diff
@@ -129,83 +46,6 @@ check_tmp_directory() {
...
-    export INSTANCE="${GATEWAY_PART_3}"
-  fi
-fi
-
-if [[ "${CLI_LOGIN}" == "true" ]]; then
-  run_login
-fi
-
-if [[ -n "${INSTANCE}" ]]; then
-  run_login "node /datalab/web/login.js 2>/dev/null"
```
yebrahim: Looks like this was the only reason we had `login.ts`, should we clean that up while at it?

containers/gateway/content/run.sh
ojarjur: No; the port we use inside the container is 8080

sources/web/build.sh
ojarjur: Done.

containers/datalab/content/run.sh
ojarjur: Done.

containers/datalab/content/run.sh
ojarjur: I've removed those now.

Issue #1049, opened 2016-11-15
The jupyter notebook have no button to stop a running cell.


soerendip:
Hi,
I cannot find a button to stop a cell in the jupyter notebooks in the datalab. Is that done on purpose?
I tried to install software with a %%bash magic function. But the script is interactive and needs input from stdin. The only way to stop the cell is to reset the whole notebook it seems. Usually ipython notebooks have a button to stop running cells without resetting the notebook.

dovy:
+1

PR #1289, opened 2017-03-28
Add "Interrupt Execution" (stop cell) button

yebrahim:
(Depends on https://github.com/googledatalab/pydatalab/pull/325)

This button interrupts the kernel execution, and can be accessed in a dropdown menu under the `Reset Session` button.

image

The `ipykernel` update is needed because the older `ipykernel` version had the undesired behavior of continuing queued cell execution after interrupting the first running cell.

Fixes https://github.com/googledatalab/datalab/issues/1049.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ containers/datalab/Dockerfile
 containers/datalab/content/license.txt
 containers/gateway/Dockerfile
 containers/gateway/content/license.txt
+containers/base/pydatalab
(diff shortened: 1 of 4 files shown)


chmeyers:
Okay, so at the risk of bikeshedding here: is "Interrupt Execution" the best name?  Also, is the button grayed out if nothing is running?

Worst case, we leave it as Interrupt Execution and change it if enough customers are confused.

yebrahim:
Oh, thanks for catching that!
We actually don't have pydatalab in the gitignore file, so everytime I build from my local pydatalab clone it gets copied, and then added by doing `git add -A`. I'll fix that.

yebrahim:
Well, `Interrupt Execution` is closest to what the button actually does, since it stops the kernel and not the running cell only. I'm open to a better name.

yebrahim:
Also, the button does not get grayed out, I don't know if this adds much to the user experience. You'll only go there if you want to stop something running, and pressing it when nothing is running doesn't achieve anything.
Also, the kernel may be busy for several reasons, executing user's code is just one of them.

Issue #1261, opened 2017-03-10
Safari side effect on action top bar (not able to Run anymore)


VelizarVESSELINOV:
Chrome view (OK):
![image](https://cloud.githubusercontent.com/assets/8209940/23781207/8662ecd2-0500-11e7-982b-de8ac6b81d38.png)

Safari view (not OK, the Run second bar is no more visible, if you scroll over large output window):
![image](https://cloud.githubusercontent.com/assets/8209940/23781218/994ef778-0500-11e7-8f09-a674714027c8.png)

chmeyers:
Another report of this issue here: http://stackoverflow.com/questions/42773452/menubar-in-datalab-keeps-vanishing

yebrahim:
I'll take a look at fixing this soon. As a workaround for now, if you scroll up till the end, then scroll up again a bit, the bar should come back.

PR #1291, opened 2017-03-28
Fix appContent height bug that shows on Safari

yebrahim:
Fixes https://github.com/googledatalab/datalab/issues/1261.


chmeyers:
This is https://github.com/googledatalab/datalab/issues/1261, right?  Please update http://stackoverflow.com/questions/42773452/menubar-in-datalab-keeps-vanishing after merge.

PR #1304, opened 2017-04-05
Only store directories as default tree locations.

### LINKED ISSUE #1303: notebooks accessed using /tree get set as default location
If you browse the tree to a folder containing a notebook, then edit the URL and add the notebook name to create a URL such as locahost:8081/tree/datalab/docs/Readme.ipynb, this location gets saved as the default location, so the next time you access localhost:8081, it opens that notebook.

jimmc:
This fixes #1303.

### DIFF
--- a/sources/web/datalab/server.ts
+++ b/sources/web/datalab/server.ts
@@ -155,10 +155,15 @@ function handleRequest(request: http.ServerRequest,
       (path.indexOf('/edit') == 0) ||
       (path.indexOf('/sessions') == 0)) {
 
-    if (path.indexOf('/tree') == 0) {
-        loadedSettings[startup_path_setting] = path
-        settings_.updateUserSetting(userId, startup_path_setting, path, true);
+    // When the user asks for something in the tree, we eventually get called
+    // here with a corresponding URL that starts with /api/contents/.
+    const apiPrefix = '/api/contents/';
+    if (path.indexOf(apiPrefix) == 0 && request.url.indexOf('type=directory') > 0) {
+        const treePath = '/tree/' + path.substr(apiPrefix.length);
+        loadedSettings[startup_path_setting] = treePath;
+        settings_.updateUserSetting(userId, startup_path_setting, treePath, true);
     }
+
     handleJupyterRequest(request, response);
     return;
   }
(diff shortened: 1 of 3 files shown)


sources/web/datalab/server.ts
```diff
@@ -155,10 +155,15 @@ function handleRequest(request: http.ServerRequest,
       (path.indexOf('/edit') == 0) ||
       (path.indexOf('/sessions') == 0)) {
 
-    if (path.indexOf('/tree') == 0) {
-        loadedSettings[startup_path_setting] = path
-        settings_.updateUserSetting(userId, startup_path_setting, path, true);
+    // When the user asks for something in the tree, we eventually get called
+    // here with a corresponding URL that starts with /api/contents/.
```
yebrahim: I think this changes the behavior of this feature. If you look through the debug logs, you'll see that `/api/contents` requests are made periodically by the frontend to poll for status, even if you don't navigate. This means if you have several tabs open with multiple directories (which is often the case for me at least), every tab you activate will at some point send a `/api/contents` request, saving it as startup path. I don't think this behavior is worse, but it's different, we might want to think through which is better for the user experience.

If we want to retain the original behavior, we should stick to triggering off of the `/tree` requests, and use the file system capabilities to check whether there's a directory with this name.

jimmc: Yeah, that makes more sense. I changed the code to check that the file exists and is a directory.

sources/web/datalab/server.ts
```diff
@@ -156,8 +156,15 @@ function handleRequest(request: http.ServerRequest,
       (path.indexOf('/sessions') == 0)) {
 
     if (path.indexOf('/tree') == 0) {
-        loadedSettings[startup_path_setting] = path
-        settings_.updateUserSetting(userId, startup_path_setting, path, true);
+        const filePath = '/content' + path.substr('/tree'.length);
```
yebrahim: nit: indentation is inconsistent in this file, it'd be nice to fix that while at it. Feel free to merge without.

jimmc: I'll do that in a separate PR.

PR #1313, opened 2017-04-09
Self-hosted material icons and web fonts

yebrahim:
Self-hosting the icons and fonts gives us the following:
- Better containment. Running Datalab now takes one less runtime dependency. If you're running Datalab while offline, for example, you'd get something like:
image
- Small improvement in load time locally, although loading these fonts on an SSH connection from GCE will take a little longer. However, browsers typically cache fonts after first visit.
- Better support for automated tests in headless browsers. We're looking to add this at some point, but the latest stable phantomjs does not support rendering web fonts in screenshots.


sources/web/datalab/static/fonts/fonts.css
```diff
@@ -0,0 +1,88 @@
...
+  line-height: 1;
+  text-transform: none;
+  letter-spacing: normal;
+  word-wrap: normal;
+  white-space: nowrap;
+  direction: ltr;
+
+  /* Support for all WebKit browsers. */
+  -webkit-font-smoothing: antialiased;
+  /* Support for Safari and Chrome. */
```
jimmc: Should there be a blank line here, as there is between the lines just below?

sources/web/datalab/static/fonts/fonts.css
```diff
@@ -0,0 +1,88 @@
...
+       url('open-sans-v13-latin-700.ttf') format('truetype'), /* Safari, Android, iOS */
+       url('open-sans-v13-latin-700.svg#OpenSans') format('svg'); /* Legacy iOS */
+}
+/* source-code-pro-regular - latin */
+@font-face {
+  font-family: 'Source Code Pro';
+  font-style: normal;
+  font-weight: 400;
+  src: url('source-code-pro-v6-latin-regular.eot'); /* IE9 Compat Modes */
+  src: local('Source Code Pro'), local('SourceCodePro-Regular'),
```
jimmc: Is the name 'SourceCodePro-Regular' required by some other use? On line 43 you use 'OpenSans', sans '-Regular'. Perhaps this should just be 'SourceCodePro' to match 'OpenSans'. Or maybe 'OpenSans' should be 'OpenSans-Regular'.

yebrahim: I'm not actually sure, these were [auto-generated](https://google-webfonts-helper.herokuapp.com/fonts), and I'm not comfortable changing things here since they're not very easy to testing because of the browser matrix.

sources/web/datalab/static/fonts/fonts.css
yebrahim: Fixed.

PR #1316, opened 2017-04-14
"Live" mode for faster development

jimmc:
A few changes to make development a little easier.

### DIFF
--- a/containers/datalab/build.sh
+++ b/containers/datalab/build.sh
@@ -26,9 +26,9 @@ COMMIT=`git log --pretty=format:'%H' -n 1`
 COMMIT_SUBSTITUTION="s/_commit_/$COMMIT/"
 
 if [ -z "$1" ]; then
-  pythonPath=''
+  pydatalabPath=''
 else
-  pythonPath=$(realpath "$1")
+  pydatalabPath=$(realpath "$1")
 fi
 
 cd $(dirname $0)
@@ -48,7 +48,7 @@ rsync -avp ../../build/ build
 cp ../../third_party/license.txt content/license.txt
 
 # Build the base docker image
-../base/build.sh "$pythonPath"
+../base/build.sh "$pydatalabPath"
 
 # Build the docker image
 docker build -t datalab .
--- a/containers/datalab/build.sh
+++ b/containers/datalab/build.sh
@@ -25,13 +25,21 @@ VERSION_SUBSTITUTION="s/_version_/0.5.$VERSION/"
 COMMIT=`git log --pretty=format:'%H' -n 1`
 COMMIT_SUBSTITUTION="s/_commit_/$COMMIT/"
 
+if [ -z "$1" ]; then
+  pythonPath=''
+else
+  pythonPath=$(realpath "$1")
+fi
+
+cd $(dirname $0)
+
 cat Dockerfile.in | sed $VERSION_SUBSTITUTION | sed $COMMIT_SUBSTITUTION > Dockerfile
 
-# Build the datalab frontend
+# Set up our required environment
 source ../../tools/initenv.sh
-cd ../../sources/web/
-./build.sh
-cd ../../containers/datalab
+
+# Build the datalab frontend
+../../sources/web/build.sh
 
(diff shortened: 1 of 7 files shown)


sources/web/datalab/config/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Running with --live, the container creates a metadata file here.
+metadata.json
```
yebrahim: I think this file is obsolete anyway, I can't see where we use it. This is separate from this change though.

containers/datalab/build.sh
```diff
@@ -25,13 +25,21 @@ VERSION_SUBSTITUTION="s/_version_/0.5.$VERSION/"
 COMMIT=`git log --pretty=format:'%H' -n 1`
 COMMIT_SUBSTITUTION="s/_commit_/$COMMIT/"
 
+if [ -z "$1" ]; then
+  pythonPath=''
```
jimmc: Done. Yeah, I didn't choose a very good name for that.

containers/datalab/run.sh
```diff
@@ -20,27 +20,68 @@
...
-    ENTRYPOINT="/bin/bash"
-  fi
-fi
+DEVROOT_DOCKER_OPTION=''
+while [ $# -gt 0 ]; do
+  case "$1" in
+    shell)
+      ENTRYPOINT="/bin/bash"
+      ;;
+    --live)
```
jimmc: Done.

containers/datalab/run.sh
```diff
@@ -20,27 +20,68 @@
...
+      # development directory so your changes are visible to the container
+      # without rebuilding.
+      DEVROOT=$(cd "$HERE" && realpath ../..)
+      DEVROOT_DOCKER_OPTION="-v $DEVROOT:/devroot"
+      echo "Updating build-dev"
+      rsync -a \
+          --exclude web/nb/config \
+          --exclude web/nb/static \
+          --exclude web/nb/templates \
+          $DEVROOT/build/ $DEVROOT/build-dev
```
jimmc: Done.

containers/datalab/run.sh
```diff
@@ -20,27 +20,68 @@
...
+      if [ ! -h $DEVROOT/build-dev/web/nb/node_modules ]; then
+        echo linking node_modules
+        ln -s /datalab/web/node_modules $DEVROOT/build-dev/web/nb/node_modules
```
jimmc: There are a whole bunch of symlinks in the build-dev directory. I don't think it's worth doing any partial clean up.

containers/datalab/run.sh
```diff
@@ -20,27 +20,68 @@
...
+      # The files in the subdirs of build/web/nb are just copied from the
+      # source dirs. In order to allow using updated versions of those files
+      # without having to run the build script, we remove them from build-dev
+      # and link them directly to the source dirs.
+      # You need to restart the notebook server in the container to make it
+      # pick up your changes.
+      for dir in config static templates; do
+        if [ ! -h $DEVROOT/build-dev/web/nb/$dir ]; then
+          echo linking $dir
+          ln -s /devroot/sources/web/datalab/$dir $DEVROOT/build-dev/web/nb/$dir
```
jimmc: No, it should be /devroot. Although this symlink is being created on the workstation, it is intended to be resolved from within the container, where the /devroot directory exists. I added a comment to the code.

jimmc:
Re using the build directory, that was the first thing I tried. I had trouble managing the symlinks, I was getting errors during the container build. I switched to build-dev so that the build directory could continue to be optimized for the build process, and the build-dev directory could be hacked up at will to make live-mode work.

jimmc:
I think we can get auto-reload functionality by adding the "--watch" option to the forever command. When I tried that, the node server would auto-restart when I changed a file, and I could just do a page reload and get the new code. Let me know if you think I should add that.

yebrahim:
Does running an `npm install` in the build directory inside the container resolve those references?

jimmc:
This PR has been superseded by the simpler and better implementation in #1321.

PR #1317, opened 2017-04-14
Add support for customizing `docker build` args.

craigcitro:
I have a collection of cases where I'm calling into
`containers/datalab/build.sh` from another script, and I'd like to customize
the args passed to `docker build`.

In my case, I'd like to add `--no-cache` to the build args. Given that this
calls docker from several places (there and in `containers/base/build.sh`),
setting an env var seems like the cleanest way to wire this in.

PTAL @chmeyers @yebrahim @jimmc (Not sure who watches the dockerfiles)

### DIFF
--- a/containers/base/build.sh
+++ b/containers/base/build.sh
@@ -34,4 +34,4 @@ else
 fi
 
 trap 'rm -rf pydatalab' exit
-docker build -t datalab-base .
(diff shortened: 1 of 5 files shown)


craigcitro:
(Just FYI, I can't merge.)

chmeyers:
Sorry, forgot you didn't have merge access. :)

PR #1318, opened 2017-04-17
Split monolithic datalab.js into smaller files

jimmc:
The code that has been pulled out of datalab.js into other files has not been changed in any way other than whitespace changes.


sources/web/datalab/static/page.js
```diff
@@ -0,0 +1,231 @@
...
+  if (event['metadata']) {
+    event['metadata'] = event['metadata'] + ',' + additionalMetadata;
+  } else {
+    event['metadata'] = additionalMetadata;
+  }
+  window.dataLayer = window.dataLayer || [];
+  window.dataLayer.push(event);
+}
+
+function toggleSidebar() {
```
yebrahim: I think `toggleSidebar`, `showHelp` (and the following markup snippet), and lines 197-227 should move into notebook-app.js.

Should this file be called app-bar.js? It consists mostly of the common app bar scripts, like the About, restarting and managing VM, sign in/out.. etc. Common code like the xhr function should perhaps stay in datalab.js.

sources/web/datalab/static/websocket.js
```diff
@@ -0,0 +1,93 @@
+function shouldShimWebsockets() {
```
yebrahim: Can we use two space tabs for this file for consistency?

sources/web/datalab/static/page.js
jimmc: I renamed page.js to appbar.js and moved some common functions back into datalab.js.

As discussed offline, I will leave toggleSidebar and showHelp here for now, until we can figure out a cleaner detangling of appbar and notebook-app.

sources/web/datalab/static/websocket.js
jimmc: Done.

sources/web/datalab/static/page.js
yebrahim: Sounds good. I'm actually surprised this seems to be the only place we got things tangled up, given we've been editing this massive file for a long time now.

PR #1319, opened 2017-04-18
Fix check for datalab version to use entire version from gcs file

yebrahim:
This PR fixes the Datalab version comparison to use the entire versions obtained from the GCS file (latest), and hard-coded in HTML doc (current). For versions that are not in semver format, they're converted to semver by prepending "0.5." to keep the current behavior.

### DIFF
--- a/sources/web/datalab/static/notebook-list.js
+++ b/sources/web/datalab/static/notebook-list.js
@@ -148,23 +148,62 @@ function initializeNotebookList(ipy, notebookList, newNotebook, events, dialog, 
     }
   });
 
-  function checkVersion(versionInfo) {
-    if (versionInfo == undefined) {
+  // Compare two arrays, each representing a semantic version in the form of [0,1,2]
+  // Returns -1, 0, 1. Returns null if they're malformed
+  function semverCompare(v1, v2) {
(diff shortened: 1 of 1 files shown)


sources/web/datalab/static/notebook-list.js
```diff
@@ -148,23 +148,62 @@ function initializeNotebookList(ipy, notebookList, newNotebook, events, dialog,
     }
   });
 
-  function checkVersion(versionInfo) {
-    if (versionInfo == undefined) {
-      return;
+  // Compare two arrays, each representing a semantic version in the form of [0,1,2]
+  // Returns -1, 0, 1. Returns null if they're malformed
+  function semverCompare(v1, v2) {
```
brandondutra: Is semver a common word? Why not call it SemanticVersionCompare or VersionCompare or etc.

yebrahim: I think `semver` is a common name for it. The go-to website is http://semver.org/.

yebrahim:
I'm working on a refactor to enable adding unit tests properly. Before that it won't be easy to fit tests here.

PR #1320, opened 2017-04-19
Allow running build scripts from anywhere

jimmc:
This also allows running just sources/web/build.sh to compile the typescript files, which I will be using in a later PR to implement live mode.

### DIFF
--- a/containers/base/build.sh
+++ b/containers/base/build.sh
@@ -20,16 +20,18 @@
 
 # Build the docker image
 if [ -n "$1" ]; then
-  src_pydatalab="$1"
+  src_pydatalab=$(realpath "$1")
 
   # Append trailing / if needed.
   if [ "${src_pydatalab: -1}" != "/" ]; then
   	src_pydatalab=$src_pydatalab"/"
   fi
 
+  cd $(dirname $0)
   rsync -avp "$src_pydatalab" pydatalab;
 else
   # Create empty dir to make docker build happy.
+  cd $(dirname $0)
   mkdir -p pydatalab;
 fi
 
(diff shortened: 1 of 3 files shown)


containers/base/build.sh
```diff
@@ -20,16 +20,18 @@
 
 # Build the docker image
 if [ -n "$1" ]; then
-  src_pydatalab="$1"
+  src_pydatalab=$(realpath "$1")
 
   # Append trailing / if needed.
   if [ "${src_pydatalab: -1}" != "/" ]; then
```
yebrahim: We don't need this if statement in this case, we can just append the slash.

yebrahim: A better fix is of course to make the code accept the path without the slash, but for the purposes of this PR, removing the if check is enough.

containers/datalab/build.sh
```diff
@@ -25,13 +25,21 @@ VERSION_SUBSTITUTION="s/_version_/0.5.$VERSION/"
 COMMIT=`git log --pretty=format:'%H' -n 1`
 COMMIT_SUBSTITUTION="s/_commit_/$COMMIT/"
 
+if [ -z "$1" ]; then
```
yebrahim: Why do we need this change? `pydatalabPath` isn't used anywhere in this script, other than being passed without modification to `base/build.sh`, which takes care of getting its `realpath`.

containers/base/build.sh
jimmc: The trailing slash is required by the rsync command, which behaves differently if it is not there. I moved adding the trailing slash to the rsync line.

containers/datalab/build.sh
jimmc: realpath (called in the else clause) fails if you pass it an empty string.

containers/base/build.sh
yebrahim: Right, thanks for fixing that.

PR #1321, opened 2017-04-19
Implement live mode for faster development.

jimmc:
This change takes advantage of existing code to implement a better and simpler live-mode than in #1316.
1. When running container/datalab/run.sh during development, by default it now starts the container with a mount point for /devroot that points to the datalab root directory.
2. Inside the container, when datalab/run.sh sees that a /devroot mount point exists, it copies in the latest js files for the notebook server, and sets a couple of environment variables to point to the source directories for the static and templates folders.
3. In the code, when those environment variables are set, it pulls static and template files from those locations, and files in those directories are not cached, but are reloaded when they are changed.

Usage:
1. By default, live-mode is enabled when starting the container using containers/datalab/run.sh. To disable live-mode, add the --no-live option.
2. When running in live mode, after making changes to files in the static and templates directories, just reload the page in the web browser.
3. If you will be making changes to typescript files, you should include the "shell" option to containers/datalab/run.sh. At the shell prompt in the container, start the notebook server by running datalab/run.sh. After making changes to typescript files on your workstation, run sources/web/build.sh, then stop the notebook server in the container and rerun datalab/run.sh. Alternatively, if you don't use the "shell" option, run "docker exec" to open a shell in the container in another window, then after making changes to typescript files, manually run "cp -p /devroot/build/web/nb/*.js /datalab/web" in the container, then restart the notebook server from the UI.

### DIFF
--- a/containers/datalab/content/run.sh
+++ b/containers/datalab/content/run.sh
@@ -114,5 +114,13 @@ then
   FOREVER_CMD="${FOREVER_CMD} -s"
 fi
 
+if [ -d /devroot ]; then
+  # For development purposes, if the user has mapped a /devroot dir, use it.
+  echo "Running notebook server in live mode"
+  export DATALAB_LIVE_STATIC_DIR=/devroot/sources/web/datalab/static
+  export DATALAB_LIVE_TEMPLATES_DIR=/devroot/sources/web/datalab/templates
+  # Make sure we have the latest compiled typescript output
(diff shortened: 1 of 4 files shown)


yebrahim:
Nice! This is a much better, and simpler, solution.

You don't need to stop/start the server manually in a shell session though, doesn't restarting the server from the About menu take care of this?

yebrahim:
I take that back, restarting won't take care of copying the compiled js files.

PR #1322, opened 2017-04-19
Make live mode work better for compiled typescript.

jimmc:
With this change, the developer no longer needs to restart the notebook server when making changes to typescript files. Just run sources/web/build.sh, and the notebook server will auto-restart.


yebrahim:
LGTM.
Would be nice to also add a comment in the run.sh header mentioning that sources can be edited outside the container, basically explaining live loading. New comers don't have the context of this and previous PRs.

PR #1331, opened 2017-04-21
Add --no-create-network flag when using `datalab create..`

parthea:
In my case I was unable to use `datalab create instance-name` at work because I didn't have permissions to run `gcloud compute networks create...`. The IT dept locked down the ability to create networks in gcloud because they want to have full control over the networks and firewall rules. Although access to Datalab is protected by SSH public key, in my case the network `'default'` has further restrictions which limit external access to port 22 so that only specific ip addresses can use SSH. This PR adds a `--no-create-network` which will use the `'default'` network as described [here](https://cloud.google.com/sdk/gcloud/reference/compute/instances/create) instead of the network `'datalab-network'`. I thought about removing `--network` altogether instead of specifying `'default'` according to the description below, but I thought `--network='default'` would improve readability.

```
--network=NETWORK
Specifies the network that the instances will be part of. If --subnet is also specified subnet must be a subnetwork of network specified by --network. If neither is specified, this defaults to the "default" network.
```

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -35,6 +35,8 @@ created instance. You can disable that behavior by passing in the
 _DATALAB_NETWORK = 'datalab-network'
 _DATALAB_NETWORK_DESCRIPTION = 'Network for Google Cloud Datalab instances'
 
+_GCE_DEFAULT_NETWORK = 'default'
+
 _DATALAB_FIREWALL_RULE = 'datalab-network-allow-ssh'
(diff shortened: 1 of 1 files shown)


parthea:
I think this issue may be specific to my unique environment (`gcloud networks create...` is locked down). Closing for now. I'll re-submit a PR if this issue is reported by more users.

nikhilk:
There are other reasons for picking a network besides the org policy ones you listed - anything that relies on being in a shared GCE network. We should reinstate this capability and get it into the product.

PR #1333, opened 2017-04-25
Add window level flag signaling proper load events

yebrahim:
This is fixing the previously arbitrary script load order, by waiting on events triggered by Jupyter for each of the four template pages.

The scripts we execute in any given page to customize it can have some code that must execute before the page is initialized by Jupyter (such as adding event listeners), but it also has some code that assumes elements have already been loaded (such as customizing functionality of buttons). This PR splits Datalab's customization scripts into preLoad and postLoad scripts, and calls them appropriately.

The big change in `nb.html` is just moving code around, no changes to that code were made.

The change also adds a `window.datalab.loaded` flag signaling that loading was finished. Later on I'll use this so that tests can wait on pages to load.

### DIFF
--- a/containers/test/content/startup.sh
+++ b/containers/test/content/startup.sh
@@ -15,8 +15,14 @@
 # limitations under the License.
 
 /datalab/run.sh &
-sleep 10
 Xvfb :10 -ac &
+
+echo -n "Polling on Datalab.."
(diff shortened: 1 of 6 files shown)


yebrahim:
Closing this one till I resolve the test issues.

PR #1336, opened 2017-04-27
Add mochajs framework for automated UI and notebook tests

yebrahim:
This PR moves tests to Javascript and adds UI regression tests.

It uses a standalone headless Chrome browser in a docker container provided by selenium, and nodejs bindings for selenium to write tests that launch a browser instance, take screenshots, and execute notebooks. Using the docker container reduces cross platform dependencies, provides a stable platform for tests, and reduces the overall test time, since no image needs to be built like before.

For UI tests, I use the resemblejs package to compare screenshots against saved (golden) images, which is able to tolerate small anti-aliasing differences sometimes caused by selenium webdrivers.

For notebook tests, I rewrote the python code that started the browser in Javascript. This way the code is smaller, and can be run using the same test runner.

Tests were written as a nodejs package, so in order to run them locally, you can do `npm install` then `npm test`. The one dependency is you have to be able to install Cairo, which is a resemblejs dependency, on your system. Take a look at the Travis config changes.


sources/web/datalab/static/appbar.html
```diff
@@ -12,12 +12,12 @@
     <span class="autosave_status"></span>
   </span>
   <div class="btn-toolbar pull-right right-toolbar">
-    <div class="btn-group">
-      <button id="ungitButton" title="Open ungit on the notebooks repository" class="toolbar-btn">
+    <div id="ungitButton" class="btn-group">
```
jimmc: I've seen people use divs as buttons, but I don't recall seeing a div used as a button, with a button element inside it. This  seems a bit confusing to me. Why do you need to move the click handler to the div? Perhaps that deserves a comment.

test/notebook/test.js
```diff
@@ -0,0 +1,116 @@
...
+      until = selenium.until;
+const test = require('selenium-webdriver/testing');
+const fs = require('fs');
+
+var driver = null;
+
+const notebooks_config = JSON.parse(fs.readFileSync('notebook/spec.json'))
+const timeOut = 60000;
+
+test.before(function() {
```
jimmc: Should we start using arrow syntax for anonymous functions?

test/ui/test.js
```diff
@@ -0,0 +1,109 @@
...
+      until = selenium.until;
+const test = require('selenium-webdriver/testing');
+const fs = require('fs');
+const resemble = require('node-resemble');
+
+var driver = null;
+
+const suiteTimeout = 60000;
+const misMatchThreshold = 0;
+const goldenPathPrefix = 'ui/goldens/';
```
jimmc: I think I would prefer no 's' on this name. That also makes it consistent with the 'broken' directory name.
Making them consistent by naming the other one 'brokens' seems a lot worse.

test/ui/test.js
```diff
@@ -0,0 +1,109 @@
...
+        fs.writeFileSync(brokenPathPrefix + testName + '.png', shot, 'base64');
+      }
+
+      assert(data.misMatchPercentage <= misMatchThreshold,
+             'Images for test ' + testName + ' are different');
+    });
+  });
+}
+
+test.describe('UI tests', function() {
```
jimmc: This string should be the name of the component being tested in this section, and the 'it' string should be a verb phrase that completes a sentence starting with the 'describe' string. I suggest you break this describe into two calls the describe: the first one would be test.descibe('Main window'), and the second one would be test.describe('Help window').

test/ui/test.js
```diff
@@ -0,0 +1,109 @@
...
+
+  test.it('Makes sure window looks fine at the beginning of tests', function() {
```
jimmc: Given my comment about 'describe' above, this string would be something like 'appears correctly before any actions have been taken'.

test/ui/test.js
```diff
@@ -0,0 +1,109 @@
...
+    return screenshotAndCompare('body.png', 'body');
+  });
+
+  test.it('Opens help menu and makes sure it looks fine', function() {
+    return driver.findElement(By.id('helpButton'))
+    .then(function(button) {
```
jimmc: I think this would look better as a one-line arrow function.

test/ui/test.js
```diff
@@ -0,0 +1,109 @@
...
+  });
+}
+
+test.describe('UI tests', function() {
+
+  test.it('Makes sure window looks fine at the beginning of tests', function() {
+    return screenshotAndCompare('body.png', 'body');
+  });
+
+  test.it('Opens help menu and makes sure it looks fine', function() {
```
jimmc: This string should be something like 'opens when Help button is clicked'.

test/ui/test.js
```diff
@@ -0,0 +1,109 @@
...
+    return driver.findElement(By.id('helpButton'))
+    .then(function(button) {
+      button.click();
+    })
+    .then(screenshotAndCompare('bodyWithHelp.png', 'bodyWithHelp'));
+  });
+
+  test.it('Close help menu and makes sure body looks fine', function() {
```
jimmc: This should be something like 'closes when body is clicked'.

test/README.md
```diff
@@ -0,0 +1,26 @@
+# Testing Google Cloud DataLab
+
+This directory contains tests Datalab. Tests are all written in Javascript, and they use a  to
```
jimmc: Looks like something is missing here: "they use a  to"

test/ui/test.js
jimmc: Looks good, thanks.

sources/web/datalab/static/appbar.html
jimmc: Here is a SO article with some discussion that mentions accessibility and screen readers as reasons not to us a div as a button: http://stackoverflow.com/questions/2777256/making-buttons-button-or-div

yebrahim: Synced with @jimmc on this. We might want to pick up some cleanup work later to make the code consistent.

yebrahim:
Oh I missed that. Fixed.

PR #1341, opened 2017-04-27
Add a terminal functionality

yebrahim:
This adds the Jupyter terminal functionality in Datalab. We will want to customize this for the Datalab experience, but we can do that in a later PR when we decide on a good usability design for it.

This PR adds the necessary plumbing for Jupyter's terminal, which is:
- A new HTML template with just the appbar and one big div for the terminal, and some css styling. A div (`dummy-screen`) is also needed to measure height and width of the terminal element, with some dummy data inside it.
- Changes to `server.ts` and `jupyter.ts` to pass that template page when requested
- An "Add Terminal" button in the tree page
- Changes to the sessions page to list active terminals under active notebooks

addterminal
terminal
sessions

### DIFF
--- a/sources/web/datalab/static/dark.css
+++ b/sources/web/datalab/static/dark.css
@@ -106,6 +106,15 @@ input {
   color: #333;
 }
 
+/* Terminal Page */
+#running .panel {
+  background-color: #444;
+  border-color: #333;
+}
+#running .panel-group .panel .panel-heading {
+  background-color: #333;
+}
+
 /* Inside Notebooks */
 #contentArea,#notebook-container {
   background-color: #444;
--- a/sources/web/datalab/templates/terminals.html
+++ b/sources/web/datalab/templates/terminals.html
@@ -10,7 +10,7 @@
   <link rel="stylesheet" href="/static/style/style.min.css" type="text/css" />
   <link rel="stylesheet" href="/static/style/datalab.css" type="text/css" />
   <link rel="stylesheet" id="themeStylesheet" href="/static/style/custom.css" type="text/css" />
-  <link rel="stylesheet" href="/static/terminal/css/override.css?v=b71f5060e3fa5cdd606894114e92b952" type="text/css">
+  <link rel="stylesheet" href="/static/terminal/css/override.css" type="text/css">
 
   <script src="/static/components/jquery/jquery.min.js"></script>
 </head>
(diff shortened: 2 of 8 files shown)


sources/web/datalab/static/dark.css
```diff
@@ -106,6 +106,15 @@ input {
   color: #333;
 }
 
+/* Terminal Page */
```
jimmc: Are there corresponding changes in light.css to keep the terminal color theme consistent with the rest of datalab?

sources/web/datalab/templates/terminals.html
```diff
@@ -0,0 +1,124 @@
...
+  <div id="app">
+    <div id="appBar">
+    </div>
+    <div id="header"></div>
+    <div id="site">
+      <div id="terminado-container">
+      </div>
+    </div>
+    <div style="position:absolute; left:-1000em">
+      <pre id="dummy-screen" style="border: solid 5px white;" class="terminal">
```
jimmc: The purpose of these 24 lines of numbers is not obvious. A comment would be useful.
Why 24 lines by 80 columns? Terminal windows can be any size now.

sources/web/datalab/templates/terminals.html
```diff
@@ -0,0 +1,124 @@
...
+        5
+        6
+        7
+        8
+        9
+        0
+        1
+        2
+        3
+      <span id="dummy-screen-rows" style="">01234567890123456789012345678901234567890123456789012345678901234567890123456789</span>
```
jimmc: Isn't this line counting out the columns?

sources/web/datalab/templates/terminals.html
yebrahim: It should actually be 25x80, I'll fix that.

The reason these exist (this `pre` and the following `span`) is they're used by the terminal script to measure the correct row height and column width, so that the number of rows and columns can be recalculated on element resize.

Take a look at https://github.com/jupyter/notebook/blob/master/notebook/static/terminal/js/main.js, lines 34 through 57.

sources/web/datalab/templates/terminals.html
yebrahim: Yes, see the comment above.

yebrahim:
I'd like to test in a little more first, then we'll see how we can document it. There's still a few things that are missing with it, so I'm hoping we'd do this incrementally.

A few things I could not get working:
- Tab completion. Pressing tab just sends the cursor to the next tab stop. This is probably related to the next point.
- Terminal does not understand arrow keys. Pressing Up for example prints `^[[A`. This looks like it's a limitation imposed by docker. Running stock Jupyter inside docker has the same problems.
- Coloring the prompt isn't working. Again, likely related to the point above.

I'll keep looking to see what's different about the docker environment that's causing this. For now, this is good enough for basic usage.

PR #1342, opened 2017-04-27
Convert most files in static to modules.

jimmc:
This PR is mostly whitespace changes Add the query parameter w=1 to the github URL to ignore whitespace and see the actual changes.
There is one intentional functional change: in appbar.js line 6 I removed a call to this.blur() that  no longer works.

### DIFF
--- a/sources/web/datalab/static/datalab.js
+++ b/sources/web/datalab/static/datalab.js
@@ -49,7 +49,7 @@ function xhr(url, callback, method) {
 
 function initializeDataLab(
     ipy, events, dialog, utils, security, appbar, editapp,
-    minitoolbar, notebookapp, notebooklist, websocket
+    notebookapp, notebooklist
   ) {
   var saveFn = function() {
     if (('notebook' in ipy) && ipy.notebook) {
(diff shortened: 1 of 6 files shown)


sources/web/datalab/static/datalab.js
```diff
@@ -47,13 +47,16 @@ function xhr(url, callback, method) {
   request.send();
 }
 
-function initializeDataLab(ipy, events, dialog, utils, security) {
+function initializeDataLab(
+    ipy, events, dialog, utils, security, appbar, editapp,
```
yebrahim: not all of these are needed though, why add the extra ones?

sources/web/datalab/static/datalab.js
```diff
@@ -102,7 +105,9 @@ define([
   'static/notebook-app',
   'static/notebook-list',
   'static/websocket'
-], function(ipy, events, dialog, utils, security, appbar, editapp,
+], function(ipy,     events, dialog, utils, security, appbar, editapp,
             minitoolbar, notebookapp, notebooklist, websocket) {
-  initializeDataLab(IPython, events, dialog, utils, security);
+     initializeDataLab(
+            IPython, events, dialog, utils, security, appbar, editapp,
```
yebrahim: while at it, can we change `IPython` to `ipy` for consistency? `IPython` is the global variable, we shouldn't be passing that.

sources/web/datalab/static/datalab.js
jimmc: I removed the unused args. I moved minibartool to the end with the other unused args to keep the matchup between the module names on lines 108 and 111.

sources/web/datalab/static/datalab.js
```diff
@@ -102,7 +105,9 @@ define([
   'static/notebook-app',
   'static/notebook-list',
   'static/websocket'
-], function(ipy, events, dialog, utils, security, appbar, editapp,
+], function(ipy,     events, dialog, utils, security, appbar, editapp,
```
jimmc: That was to make the names line up between this line and the call to initializeDatalab. Since I changed IPython to ipy, I removed the spaces here.

sources/web/datalab/static/datalab.js
jimmc: Sounds good to me. Done. I didn't know why IPython was used instead of ipy, so I didn't change it earlier.

jimmc:
I didn't export any private methods because I don't know which ones we want to add. I think we can probably just do that as we add unit tests that need that access.

PR #1344, opened 2017-04-28
Adding a Python 3 Kernel to Datalab.

### LINKED ISSUE #902: Python 3 Support
Hi, is it possible currently to upgrade the version of Python used by Datalab? My default install uses 2.7.9, but would love to use 3.xx. Thoughts?

Thank you.

chmeyers:
Installs Python 3 and py3 versions of our supported packages,
including the python3 jupyter kernel, and adds a dropdown to
the notebook page to allow users to switch kernels.

Docker Image size goes from 1.734GB -> 2.507GB, almost all of which is due to the size of the python3 packages.  I've added a "py2 only" container build in case someone needs the smaller image.

Fixes #902 

Somewhat based on yebrahim's py3 branch:
https://github.com/googledatalab/datalab/compare/master...yebrahim:yebrahim/py3


parthea:
> I've added a "py2 only" container build in case someone needs the smaller image.

My initial thought is that there is a cost associated with having 2 separate docker files for python 2 and python 2/3 (Specifically additional time would be required for contributors in terms of maintenance and testing 2 separate docker files). There is a risk that the python 2 version may not be maintained. One potential solution is to remove the python 2 only Dockerfile until there is demand for a smaller image.

sources/web/datalab/templates/nb.html
```diff
@@ -161,6 +161,15 @@
                 <li id="interruptKernelButton"><a href="#">Interrupt Execution</a></li>
               </ul>
             </div>
+            <div class="btn-group">
+              <button id="kernelSelectorButton" type="button" data-toggle="dropdown" class="toolbar-btn" title="Select kernel">
+                <i class="material-icons">track_changes</i>
+                <span class="toolbar-text">Kernel: </span>
+                <span id="currentKernelName"></span>
+              </button>
+              <ul class="dropdown-menu" id="kernelSelectorDropdown">
```
yebrahim: can you add a screenshot for this to start discussion on the design? I didn't actually put much thought into how usable this is at the time I made those changes, it was the simplest thing to do for switching kernels.

sources/web/datalab/templates/nb.html
```diff
@@ -161,6 +161,15 @@
                 <li id="interruptKernelButton"><a href="#">Interrupt Execution</a></li>
               </ul>
             </div>
+            <div class="btn-group">
```
yebrahim: Adding another button to the toolbar will likely need changes to the minimum width of screen, and the media size decorators.

containers/base/Dockerfile.py2
```diff
@@ -0,0 +1,175 @@
+# Copyright 2015 Google Inc. All rights reserved.
```
chmeyers: Yes, this is just a branch.

containers/datalab/build.sh
```diff
@@ -34,6 +34,7 @@ fi
 cd $(dirname $0)
 
 cat Dockerfile.in | sed $VERSION_SUBSTITUTION | sed $COMMIT_SUBSTITUTION > Dockerfile
+cat Dockerfile.in | sed $VERSION_SUBSTITUTION | sed $COMMIT_SUBSTITUTION | sed "s/datalab-base/datalab-base-py2/" > Dockerfile.py2
```
chmeyers: Done

containers/base/build.sh
```diff
@@ -32,3 +32,4 @@ fi
 
 trap 'rm -rf pydatalab' exit
 docker build ${DOCKER_BUILD_ARGS} -t datalab-base .
+docker build ${DOCKER_BUILD_ARGS} -t datalab-base-py2 -f Dockerfile.py2 .
```
chmeyers: Done

containers/base/Dockerfile
```diff
@@ -81,8 +77,78 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
...
+                   --upgrade-strategy only-if-needed --no-cache-dir requests==2.9.1 \
+                   --upgrade-strategy only-if-needed --no-cache-dir oauth2client==2.2.0 \
+                   --upgrade-strategy only-if-needed --no-cache-dir httplib2==0.9.2 \
+                   --upgrade-strategy only-if-needed --no-cache-dir futures==3.0.5 && \
+    pip3 install -U --upgrade-strategy only-if-needed --no-cache-dir matplotlib==1.5.3 && \
+    pip3 install -U --upgrade-strategy only-if-needed --no-cache-dir ggplot==0.6.8 && \
+    pip3 install -U --upgrade-strategy only-if-needed --no-cache-dir seaborn==0.7.0 && \
+    pip3 install -U --upgrade-strategy only-if-needed --no-cache-dir PyYAML==3.11 && \
+    pip3 install -U --upgrade-strategy only-if-needed --no-cache-dir six==1.10.0 && \
+    # Install notebook after ipywidgets, and keep it pinned to 4.2.3, higher versions may
```
chmeyers: Done

chmeyers:
Screenshot of UX:
image

sources/web/datalab/templates/nb.html
chmeyers: Done

sources/web/datalab/templates/nb.html
chmeyers: Done

chmeyers:
Did some investigation with the first official build that has this.

Docker image compressed on GCR is 790MB vs 566 MB for the previous version.  Expectation is that initial startup time for VMs is increased by ~15s.

RAM usage didn't seem to change significantly, this and the previous version appeared to use within tens of MB of each other.

nikhilk:
That's good to hear that there are no immediate significant issues.

PR #1348, opened 2017-04-30
Fix infinite recursion in Tree page init - add test

yebrahim:
This PR fixes an infinite recursion in the Tree page's post initialization script, which results from calling that script whenever the notebook list is redrawn, which happens periodically. One step in that init code is to override the `selection_changed` function with a wrapper. If this is done more than once, we get an infinite recursion since we lose our reference to the original function.

The best fix I could find is to raise a flag after calling the init code the first time. Jupyter does not offer a better event to listen to for page completion.

This PR also adds a regression UI test that makes sure selection still works after reloading the notebook list.
Another small fix included here is for a bug in the Travis file that wouldn't print broken images after testing.


.travis.yml
```diff
@@ -38,4 +38,4 @@ script:
 - cd $TRAVIS_BUILD_DIR/test
 - npm install
 - npm test
-- for i in `ls ui/broken/`; do echo $i; base64 \"$i\"; echo \"-----\"; done
+- for i in `ls ui/broken/`; do echo ui/broken/$i; base64 ui/broken/$i; echo \"-----\"; done
```
jimmc: What happens if ui/broken is empty?

sources/web/datalab/static/datalab.js
```diff
@@ -80,15 +80,29 @@ function initializeDataLab(
...
+    // the first occurrence
+    let notebookListLoaded = false;
     events.on('draw_notebook_list.NotebookList', function() {
-      notebooklist.postLoad(ipy, ipy.notebook_list, ipy.new_notebook_widget,
-                             events, dialog, utils);
-      window.datalab.loaded = true;
+      if (!notebookListLoaded) {
+        notebooklist.postLoad(ipy, ipy.notebook_list, ipy.new_notebook_widget,
+                              events, dialog, utils);
+        window.datalab.loaded = true;
```
jimmc: Since you are already setting window.datalab.loaded, is there a reason to also set notebookListLoaded? Will there ever be a time when one is set but not the other?
Likewise for sessionListLoaded below.

test/README.md
```diff
@@ -23,4 +23,7 @@ Cairo, which is a resemblejs dependency, on your system. Take a look at the Trav
 what is required on a vanilla VM to build resemblejs.
 
 [Mocha JS](https://github.com/mochajs/mocha) is the test runner. After you do `npm install`, you
-can run individual tests using the `mocha` cli, for example: `mocha notebook/test.js`.
+can run individual tests using the `mocha` cli, for example: `mocha notebook/test.js`. Please note
```
jimmc: Nice clarification, thanks for adding this note.

test/ui/test.js
```diff
@@ -26,6 +26,7 @@ const resemble = require('node-resemble');
 var driver = null;
 
 const suiteTimeout = 60000;
+const initTimeout = 10000;
```
jimmc: I assume these units are milliseconds, but it might be nice to make that explicit, either in code or a comment.

test/ui/test.js
```diff
@@ -57,11 +58,13 @@ test.describe('UI tests', function() {
 
   // build the selenium webdriver
   before(function() {
+    this.timeout(initTimeout);
     driver = new selenium.Builder()
       .forBrowser('chrome')
       .usingServer('http://localhost:4444/wd/hub')
       .build();
 
+    driver.manage().timeouts().setScriptTimeout(5000);
```
jimmc: Please add a variable for this constant, probably up with the other timeout variables.

test/ui/test.js
```diff
@@ -105,6 +108,33 @@ test.describe('UI tests', function() {
         .then(screenshotAndCompare('body.png', 'body'));
     });
 
+    test.it('shows(hides) extra buttons when a tree item is (un)selected', function() {
+      // simulate a list reload by calling the Jupyter function, and waiting
```
jimmc: I think block comments are a bit easier to read when they use normal sentence case and punctuation.

test/ui/test.js
```diff
@@ -105,6 +108,33 @@ test.describe('UI tests', function() {
...
+
+      // get an item in the file listing. the reason we're not using first
+      // is because the first item is not selectable (up dir)
+      let listItem = driver.findElement(
+        By.xpath('(//div[@id="notebook_list"]/div[@class="list_item row"])[last()]'));
+
+      // click the item, make sure the UI changes accordingly (extra buttons added)
+      listItem.click();
+      return screenshotAndCompare('listItemSelected.png', 'listItemSelected')
+      .then(function() {
```
jimmc: Please indent this block, as a continuation of the return statement.

test/ui/test.js
```diff
@@ -105,6 +108,33 @@ test.describe('UI tests', function() {
...
+        let callback = arguments[arguments.length - 1];
+        require(['base/js/events'], function(events) {
+          events.on('draw_notebook_list.NotebookList', callback);
+        });
+        Jupyter.notebook_list.load_list()
+      });
+
+      // get an item in the file listing. the reason we're not using first
+      // is because the first item is not selectable (up dir)
+      let listItem = driver.findElement(
```
jimmc: const instead of let

test/ui/test.js
```diff
@@ -105,6 +108,33 @@ test.describe('UI tests', function() {
         .then(screenshotAndCompare('body.png', 'body'));
     });
 
+    test.it('shows(hides) extra buttons when a tree item is (un)selected', function() {
+      // simulate a list reload by calling the Jupyter function, and waiting
+      // on the draw list event to make sure all elements have rendered
+      // this makes sure the tree initialization code isn't executed more than once
+      driver.executeAsyncScript(function() {
+        let callback = arguments[arguments.length - 1];
```
jimmc: const instead of let

.travis.yml
yebrahim: Nothing is printed.

sources/web/datalab/static/datalab.js
yebrahim: What does that add though? `window.datalab.loaded` should already indicate notebook list load for the Tree page, and so on.

test/ui/test.js
yebrahim: Sure, I'll put a comment.

test/ui/test.js
yebrahim: Done.

test/ui/test.js
yebrahim: Done.

test/ui/test.js
yebrahim: Done.

test/ui/test.js
yebrahim: Done.

test/ui/test.js
yebrahim: Done.

sources/web/datalab/static/datalab.js
jimmc: I am suggesting you remove notebookListLoaded and just use window.datalab.loaded to determine whether to call postLoad. That would be a couple fewer lines of code to maintain. Is there a reason not to do it that way?

yebrahim: Oh I see, you're absolutely right. Fixed.

PR #1355, opened 2017-05-04
Terminal fixes: use bash, UI improvements

yebrahim:
This fix adds a bashrc file inside the container to set up the enviornment for the user. Currently it changes directory to `/content`, colors the prompt, and adds a message to tell the user to keep sensitive work in `/content`. The `ENV` env var is set so that file gets loaded by all shells.

The rest of the PR is UI fixes to make things look better, and fixed the order of loading the Jupyter dialog module.

image

### DIFF
--- a/sources/web/datalab/templates/terminals.html
+++ b/sources/web/datalab/templates/terminals.html
@@ -116,7 +116,7 @@
     window.datalab = {};
 
     $("#appBar").load("/static/appbar.html", function() {
-      require(['tree/js/main.min', '/static/terminal/js/main.min.js']);
+      require(['/static/terminal/js/main.min.js', 'tree/js/main.min']);
     });
   </script>
   <script src="//www.gstatic.com/feedback/api.js" async="true" defer="true"></script>
--- a/sources/web/datalab/templates/terminals.html
+++ b/sources/web/datalab/templates/terminals.html
@@ -29,13 +29,12 @@
     </div>
     <div id="header"></div>
     <div id="site">
-      <div id="terminado-container">
-      </div>
+      <div id="terminado-container"></div>
     </div>
     <div style="position:absolute; left:-1000em">
       <!-- The following elements contain dummy content that is used to calculate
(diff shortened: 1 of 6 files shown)


sources/web/datalab/templates/terminals.html
```diff
@@ -60,9 +59,8 @@
         1
         2
         3
-        4
-      <span id="dummy-screen-rows" style="">01234567890123456789012345678901234567890123456789012345678901234567890123456789</span>
-      </pre>
+        </pre>
+      <span id="dummy-screen-rows" style="">01234567890123456789012345678901234567890123456789012345678901234567890123456789012</span>
```
jimmc: Why did you change from 80 to 83 characters here?

yebrahim: 80 characters wasn't well aligned with the size of the terminal. Unfortunately we have little control over this calculation because of the so many parameters it uses:
- Window dimensions
- Terminal container dimensions
- Line height
- Padding and margin

So I had to adjust it until it fit well with our usage.

sources/web/datalab/templates/terminals.html
```diff
@@ -29,13 +29,12 @@
     </div>
     <div id="header"></div>
     <div id="site">
-      <div id="terminado-container">
-      </div>
+      <div id="terminado-container"></div>
     </div>
     <div style="position:absolute; left:-1000em">
       <!-- The following elements contain dummy content that is used to calculate
       the size of the terminal characters. For more context, look at terminal/js/main.js -->
```
jimmc: Maybe "... that is used by hack code in Jupyter ..."

yebrahim: Sure.

jimmc: Was there a change here?

yebrahim: Oh, it seems I missed pushing the comment fix somehow!

PR #1358, opened 2017-05-05
Load optional env file in live mode.

jimmc:
This PR adds code that executes only in live (developer) mode. It looks for an optional env file to source from within the container before starting the datalab server. This makes it easy for the developer to add environment variables during development. Or to use apt-get to install personal favorite packages that are useful for development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ staging
 *.swo
 *.ipynb.yaml
 *.log
+container.env
 containers/datalab/Dockerfile
 containers/datalab/content/license.txt
 containers/gateway/Dockerfile
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.ipynb_checkpoints
 build
+dev
 staging
 *.gradle
 *.pyc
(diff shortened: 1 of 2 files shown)


yebrahim:
This change is definitely useful, but I think we should do it as part of the Datalab Dockerfile itself. Otherwise, every time you start the container, it'll take a few minutes to `apt-get update` and do whatever else you have in there (I'm guessing `apt-get update` is probably part of your env script). We should have a final step in the dockerfile that checks a parameter that gets passed by the `build.sh` script. What do you think?

ojarjur:
How does this compare to the startup script feature? In particular, why would a developer use this instead of having a startup.sh script mapped into their /content/datalab/.config directory?

jimmc:
OK, I figured out the startup.sh stuff, and it does what I need, so I will withdraw this PR. The one annoyance there is that my ~/datalab directory and everything under it is owned by root, so I had to do some "sudo chown" stuff to get my startup.sh created.

I do have apt-get in my script, but it is protected by a test for the program, so it is fast after the first time.

yebrahim:
I meant when you kill the container and start it again, nothing should be saved, so the apt get will be executed again from scratch.

PR #1368, opened 2017-05-17
Don't attempt to reconnect if the instance is no longer RUNNING.

### LINKED ISSUE #1111: Stop trying to reconnect if an instance switches state
Right now the `datalab connect` command continually tries to reconnect to an instance if the SSH tunnel dies.

However, if the VM is in a state other than 'RUNNING', then those attempts will always fail until the user kills the command and re-runs it (the 'connect' command will restart the instance before trying to connect).

Since the user has to re-run the command, we should just have the loop detect this condition and exit with a message to re-run it.

An alternative would be to have the loop restart the VM, but we don't want to do that. The reason is that restarting the VM should require explicit user action rather than happening automatically.

jimmc:
This fixes #1111. If the instance is no longer RUNNING when it is time to attempt to reconnect, it just prints a message and exits.

### DIFF
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -277,6 +277,13 @@ def connect(args, gcloud_compute, email, in_cloud_shell):
             return
         if remaining_reconnects == 0:
             return
+        # Before we try to reconnect, check to see if the VM is still running.
+        status, metadata_items = utils.describe_instance(
+            args, gcloud_compute, instance)
+        if status != 'RUNNING':
+            print('Instance {0} is no longer running ({1})'.format(
+                instance, status))
+            return
         print('Attempting to reconnect...')
         remaining_reconnects -= 1
         # Don't launch the browser on reconnect...
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -76,6 +76,10 @@ unsupported_browsers = [
 ]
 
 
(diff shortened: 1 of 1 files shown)


tools/cli/commands/connect.py
```diff
@@ -277,6 +277,13 @@ def connect_and_check():
             return
         if remaining_reconnects == 0:
             return
+        # Before we try to reconnect, check to see if the VM is still running.
+        status, metadata_items = utils.describe_instance(
+            args, gcloud_compute, instance)
+        if status != 'RUNNING':
```
ojarjur: Now that we are using 'RUNNING' in two different places, perhaps we should define a constant for it.

tools/cli/commands/connect.py
```diff
@@ -277,6 +277,13 @@ def connect_and_check():
             return
         if remaining_reconnects == 0:
             return
+        # Before we try to reconnect, check to see if the VM is still running.
+        status, metadata_items = utils.describe_instance(
```
ojarjur: I would rename `metadata_items` to `unused_metadata_items` to make it clear that we don't need that part of the return value.

jimmc: Done.

tools/cli/commands/connect.py
jimmc: Done.

PR #1369, opened 2017-05-18
Add starter unit tests for Datalab

yebrahim:
This PR adds the required plumbing for running unit tests, and two test to start with. The changes to `notebook-list.js` are refactoring to export some functions, and rely on a new `util` module for the debug logger, so that it can be mocked out.


test/unittests/frontend.js
```diff
@@ -0,0 +1,64 @@
+/*
```
jimmc: How about we adopt the convention that each source file has a corresponding test file? Since these tests are for notebook-list.js, this file should be named that, or test-notebook-list.js, or something like that.

Also, will these tests be for both the ts and the js files? I.e. for both the notebook server and the client code? Do we want to put those into separate directories?

sources/web/datalab/static/utilMock.js
```diff
@@ -0,0 +1,11 @@
+define(() => {
```
jimmc: Since this is a testing file, should it go in test/ somewhere?

test/unittests/frontend.js
```diff
@@ -0,0 +1,64 @@
...
+  baseUrl: '../sources/web/datalab/static/',
+  nodeRequire: require,
+  map: {
+    '*': {
+      util: 'utilMock'
+    }
+  }
+});
+
+describe('Unit tests', function() {
```
jimmc: Would there be any confusion caused by removing this level of nesting? Are there non-unit tests that produce output similar to the format in which these tests produce output?

yebrahim: No confusion, it's just organizational. We now have three suites of tests, each starts off with its category name, UI, notebook, or unit tests, then the name of the test.

test/unittests/frontend.js
yebrahim: Done.

.gitignore
```diff
@@ -16,5 +16,6 @@ containers/datalab/content/license.txt
 containers/gateway/Dockerfile
 containers/gateway/content/license.txt
 containers/base/pydatalab
+test/node_modules
 node_modules
-test/ui/broken
+test/ui/broken
```
jimmc: Could use a newline at the end of this line.

yebrahim: I'd like to keep in the same paragraph, because it's related to the `ui/golden` bit.

PR #1370, opened 2017-05-18
Fix test tree page path, clarify README

yebrahim:


test/README.md
```diff
@@ -17,13 +17,29 @@ a list of cells to ignore for each notebook, which is useful when specific cells
 intentionally erroneous. Each notebook is then loaded in a browser page, and all its cells are
 executed, and tests pass if no errors are produced.
 
-Tests were written as a nodejs package, so in order to run them locally, you can do `npm install`
-then `npm test` (or `npm install-test`). The one dependency is you have to be able to install
+## Running Tests
```
jimmc: Please make two sections, one for one-time setup, and one for running tests. If I do the setup, then come back two months later to run a test, I don't want to read anything about setup again.

test/README.md
```diff
@@ -17,13 +17,29 @@ a list of cells to ignore for each notebook, which is useful when specific cells
...
+
+### TL;DR
+
+Make sure you can install cairo on your system. Check the `.travis.yml` file for more info
+`cd test`
+`npm install-test`
```
jimmc: I think you should get rid of this npm target. It is only ever useful once, and it's not worth the mental overhead of worrying about whether it is any different than "npm install" followed by "npm test". If you have separate sections for one-time setup and for ongoing running of tests, then the setup section will say to run "npm install" and the test section will say to run "npm test", and the word "install-test" will never appear anywhere.

test/README.md
```diff
@@ -17,13 +17,29 @@ a list of cells to ignore for each notebook, which is useful when specific cells
...
+Or
+`npm test` if you have already installed the dependencies once
```
jimmc: Get rid of the comment here. If you have separate sections for one-time setup and for running tests, it will be clear that one-time setup has to be done before running tests.

test/README.md
```diff
@@ -17,13 +17,29 @@ a list of cells to ignore for each notebook, which is useful when specific cells
...
 executed, and tests pass if no errors are produced.
 
-Tests were written as a nodejs package, so in order to run them locally, you can do `npm install`
-then `npm test` (or `npm install-test`). The one dependency is you have to be able to install
+## Running Tests
+
+### TL;DR
+
+Make sure you can install cairo on your system. Check the `.travis.yml` file for more info
+`cd test`
```
jimmc: README files are typically at the root of a project, where it is unambiguous as to what the starting directory should be. In this case, I think the least ambiguity would be to state that commands are relative to this test directory, rather than putting in a 'cd' command which still could be ambiguous (is there supposed to be a test directory within this test directory?)

test/README.md
```diff
@@ -17,13 +17,29 @@ a list of cells to ignore for each notebook, which is useful when specific cells
...
+`npm install-test`
+Or
+`npm test` if you have already installed the dependencies once
+
```
jimmc: Might be nice to say something about looking in ui/broken for images of failed tests. Also, how to clean up that stuff when done - or maybe add ui/broken to the .gitignore file.

test/README.md
```diff
@@ -17,13 +17,29 @@ a list of cells to ignore for each notebook, which is useful when specific cells
 intentionally erroneous. Each notebook is then loaded in a browser page, and all its cells are
 executed, and tests pass if no errors are produced.
 
-Tests were written as a nodejs package, so in order to run them locally, you can do `npm install`
-then `npm test` (or `npm install-test`). The one dependency is you have to be able to install
+## Running Tests
+
+### TL;DR
+
+Make sure you can install cairo on your system. Check the `.travis.yml` file for more info
```
yebrahim: Well the installation is different according to your platform and set up, I can't list all steps here. If I list them for Ubuntu for example, it's different from one version to the other, and your specific machine might be missing some packages. I don't want to put a list of steps here that will only work for a subset of developers.

test/README.md
yebrahim: I'll detail it in the separate sections, but I think I should keep it here in the tl;dr, since this will always work.

yebrahim:
Merged a little too early, will send another PR to address @jimmc's comments.

PR #1374, opened 2017-05-20
Make the command line tool work well with redirected stdout.

ojarjur:
The `datalab` command line tool prints various messages to stdout,
but it was previously never calling `sys.stdout.flush()`.

That is fine when the program is run with stdout being the user's
console, because in that case the writes would not be buffered.

However, if a user runs the tool with stdout directed to a regular
file, then the writes to stdout are buffered until `sys.stdout.flush()`
is called. That, in turn, meant that the output of `print` calls
would not be written until the program exited (which results in a flush).

This becomes an issue when a user wants to issue a long-running
invocation of the tool (such as the `connect` command) in a background
process and watch the status using something like `tail -f`.

This change fixes that issue by adding a new `print_and_flush` helper
method to the `utils` package and then using that instead of calling
`print` directly.

### DIFF
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -160,7 +160,7 @@ def connect(args, gcloud_compute, email, in_cloud_shell):
     connect_msg = ('Connecting to {0}.\n'
                    'This will create an SSH tunnel '
                    'and may prompt you to create an rsa key pair.')
-    print(connect_msg.format(instance))
+    utils.print_and_flush(connect_msg.format(instance))
 
     datalab_port = args.port
     datalab_address = 'http://localhost:{0}/'.format(str(datalab_port))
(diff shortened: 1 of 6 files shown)


yebrahim:
What happens with exceptions? Should we also flush stderr when an exception is raised?

yebrahim:
I can't find a spec that says stderr is unbuffered for py2, for [py3 it is not](https://docs.python.org/3/library/sys.html#sys.stderr). There's also [this thread](https://mail.python.org/pipermail/tutor/2003-November/026645.html) about making python unbuffered, can we use the env var instead?

ojarjur:
@yebrahim The only time we write to stderr or print an exception is right before the tool terminates, so manually flushing the output buffer should be unnecessary.

ojarjur:
On second thought, adding the "-u" flag to our wrapper script (which invokes the datalab.py file) solves the same issue in a cleaner and more consistent way.

I'm dropping this change.

PR #1375, opened 2017-05-22
Update notebook library from 4.2.3 to 4.4.1

tylere:
This change upgrades  the notebook library to a version required by [IPyLeaflet](https://github.com/ellisonbg/ipyleaflet).

I attempted to run the Selenium test suite using the [Developer Environment](https://github.com/googledatalab/datalab/wiki/Development-Environment) instructions, but failed to install one of the dependencies (node-resemble).

As an alternative method of testing, I wrote a Python script ([https://gist.github.com/tylere/77eb0ac86f40006bf0016cacd276b93a](https://gist.github.com/tylere/77eb0ac86f40006bf0016cacd276b93a)) that uses nbconvert to execute each of the Datalab doc notebooks and to output HTML and Markdown representations using both notebook versions 4.2.3 and 4.4.1.  The outputs were compared by:

1. first comparing the Markdown outputs,
2. for those notebooks where the Markdown outputs differed, I opened up each of the executed notebooks in Datalab to compare them manually.

Given this method of testing, the upgrade from 4.2.3 to 4.4.1 appeared to have no detrimental effect.


yebrahim:
Looks like 4.4.1 breaks a few things. I think we'll need to spend some time looking at the upgrade, maybe to 5.0.0 and properly fix what breaks. As an example, terminals don't work, because Jupyter [switched to using xterm.js](https://github.com/jupyter/notebook/releases/tag/4.3.0) which is not working for us.

As a workaround for now, can your custom container restore notebook to 4.2.3 after installing ipyleaflet? I doubt that will break anything since their dependency is on the ipykernel, which can remain as is.

tylere:
Yes, it looks like I can install IPyLeaflet and downgrade using:

RUN pip install ipyleaflet \
  && jupyter nbextension enable --py --sys-prefix ipyleaflet \
  && pip install notebook==4.2.3

Is there a way to open a terminal in a Datalab browser window? I don't recall seeing that in the docs.

yebrahim:
We've already [added a](d7adeca889cfee65002be04be497e3332290d736) Terminal functionality in Datalab, it's coming in the new release. If you're building it locally anyway, then you should have it in the latest master.

For release Datalab instances, you'll have to do that from inside a notebook like you're already doing, until we get the Terminal out.

yebrahim:
I think we're good to close this one. Please reopen or comment if you have any concerns.

PR #1379, opened 2017-05-23
Move to version file backwards-compatible release.sh

yebrahim:
This change adds a template file for the config_local.js that gets uploaded to GCS on release. This template has placeholders for fields that are changed by the release process.

The release script is also changed to be backwards-compatible with this new format and the old one.

In addition, I added a `version.sh` file that contains the source of truth for the Datalab version that is used by the Datalab/Gateway build scripts as well as release script. *I also bumped the version to the next semver value `1.2.DATE`*, ideally this should be a standalone change before each release.

### DIFF
--- a/tools/release/config_local_template.js
+++ b/tools/release/config_local_template.js
@@ -1,8 +1,9 @@
 const vid = document.body.getAttribute("data-version-id");
 const GTM_ACCOUNT = {{GTM_ACCOUNT_PLACEHOLDER}};
 const LATEST_SEMVER = {{DATALAB_VERSION_PLACEHOLDER}};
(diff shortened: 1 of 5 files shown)


tools/release/config_local_template.js
```diff
@@ -0,0 +1,19 @@
+const vid = document.body.getAttribute("data-version-id");
+const LATEST_SEMVER = {{DATALAB_VERSION_PLACEHOLDER}};
+const latest = vid.split('.')[2] < "20170523" ? {{DATALAB_VERSION_PATCH_PLACEHOLDER}} : LATEST_SEMVER;
```
yebrahim: `vid` is always in semver format, but sure I can add a sanity test.

PR #1380, opened 2017-05-24
Switch to the GA version of the `gcloud source repos` commands.

ojarjur:
As of version 156.0.0 of the Cloud SDK, the `gcloud source repos`
command group has been moved to the GA release track, so we no
longer need to rely on the `alpha` commands.


ojarjur:
@jimmc Good catch; fixed.

PR #1382, opened 2017-05-24
Fix release script exiting with non-zero

yebrahim:
Release script exits if any command returns non-zero, I was using a grep command to check for the existence of the `LATEST_SEMVER` string, moved that into an if statement.

### DIFF
--- a/tools/release/publish.sh
+++ b/tools/release/publish.sh
@@ -69,8 +69,7 @@ gsutil cp gs://${PROJECT_ID}/deploy/config_local.js ./config_local.js
 # Get the latest and previous versions from the config_local. Note that older
 # config files don't have the full semantic version specified, so cannot extract using
 # the "LATEST_SEMVER = " pattern, and instead use "latest: "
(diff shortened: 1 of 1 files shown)


tools/release/publish.sh
```diff
@@ -69,8 +69,7 @@ gsutil cp gs://${PROJECT_ID}/deploy/config_local.js ./config_local.js
 # Get the latest and previous versions from the config_local. Note that older
 # config files don't have the full semantic version specified, so cannot extract using
 # the "LATEST_SEMVER = " pattern, and instead use "latest: "
-cat ./config_local.js | grep "LATEST_SEMVER = "
-if [[ $? -eq 0 ]]; then
+if [[ `cat ./config_local.js | grep "LATEST_SEMVER = "` -eq 0 ]]; then
```
ojarjur: This line should be:

    if [[ $(cat ./config_local.js | grep "LATEST_SEMVER = ") ]]; then

PR #1387, opened 2017-05-25
Update manual release process in release/README.md

yebrahim:


tools/release/README.md
```diff
@@ -35,30 +35,45 @@ the new image as described below:
...
-charts are displayed properly: 
-
-    samples/Conversion Analysis with Google AnalyticsData.ipynb 
-    tutorials/Data/Interactive Charts with Google Charting APIs.ipynb 
+1. Bump up the hard-coded Datalab version to the next semantic version in the
+[tools/release/version.sh](https://github.com/googledatalab/datalab/blob/master/tools/release/version.sh)
+file, send a PR with the changes and merge.
+
+2. Try the new image before releasing it. You can either try it locally:
+```
```
ojarjur: Remove the bit about trying it locally. We should always run `datalab create` with the new image before releasing it.

tools/release/README.md
```diff
@@ -35,30 +35,45 @@ the new image as described below:
...
-parcoords graph looks right.
+displayed properly. There is no need to run them, just verify that the
+graphs look right:
+- [samples/ML Toolbox/Image Classification/Flower/Local End to end.ipynb](http://localhost:8081/notebooks/datalab/docs/samples/ML%20Toolbox/Image%20Classification/Flower/Local%20End%20to%20End.ipynb)
+- [samples/ML Toolbox/Regression/Census/1 Local End to End.ipynb](http://localhost:8081/notebooks/datalab/docs/samples/ML%20Toolbox/Regression/Census/1%20Local%20End%20to%20End.ipynb)
+
+(*You can click these links directly if you're connected to Datalab at localhost://8081*)
 
-    tutorials/Machine Learning/Iris/7. HyperParameter Tuning.ipynb
+4. If everything looks fine, you can now run the
```
ojarjur: This should be the last step.

We need to update the releases first before publishing the latest build so that anyone seeing the update notification can see the update in the releases.

PR #1390, opened 2017-05-26
Only upload new config_local.js if the version changes

yebrahim:
*hint: use `w=1` querystring param for this PR*


tools/release/publish.sh
```diff
@@ -80,17 +80,23 @@ CURRENT_DIR=$(dirname "${BASH_SOURCE[0]}")
...
+  echo "Filling latest patch=${DATALAB_VERSION_PATCH}"
+  sed -i -e s/{{DATALAB_VERSION_PATCH_PLACEHOLDER}}/\"${DATALAB_VERSION_PATCH}\"/ $CONFIG_TEMPLATE
+  echo "Filling previous=${CURRENT_VERSION}"
+  sed -i -e s/{{PREV_SEMVER_PLACEHOLDER}}/\"${CURRENT_VERSION}\"/ $CONFIG_TEMPLATE
+  echo "Filling gtm account=${GTM_ACCOUNT}"
+  sed -i -e s/{{GTM_ACCOUNT_PLACEHOLDER}}/\"${GTM_ACCOUNT}\"/ $CONFIG_TEMPLATE
 
-gsutil cp ./config_local.js gs://${PROJECT_ID}/deploy/config_local_${BUILD}.js
-gsutil cp $CONFIG_TEMPLATE gs://${PROJECT_ID}/deploy/config_local.js
+  gsutil cp ./config_local.js gs://${PROJECT_ID}/deploy/config_local_${BUILD}.js
```
ojarjur: This should copy $CONFIG_TEMPLATE not ./config_local.js

yebrahim: ./config_local.js is the current file which is about to become old, it's copied with the BUILD suffix. Copying the template is the step below.

ojarjur: What I'm getting at is that the code as it is now will result in "gs://${PROJECT}/deploy/config_local_${BUILD}.js" and "gs://${PROJECT}/deploy/config_local.js" being different, whereas they should be the same.

I.E. "gs://${PROJECT}/deploy/config_local_${BUILD}.js" is supposed to reflect the version of that file for ${BUILD}, not the version of that file that preceded ${BUILD}

yebrahim: You're right, thanks for catching that.

PR #1391, opened 2017-05-26
Switch unit tests from mocha to jasmine.

jimmc:
The ui tests are failing, but they are failing for me on master as well. The notebook and unit tests pass.


test/spec/support/jasmine.json
```diff
@@ -0,0 +1,11 @@
+{
+  "spec_dir": "spec",
```
yebrahim: why not keep this as `unittests`?

test/run.sh
```diff
@@ -14,43 +14,103 @@
...
+  # Unit tests are fast, run them all first
+  runUnitTests
+
+  if [[ $RUN_INTEGRATION_TESTS -ne 0 ]]; then
+    makeTestsHome
+    startContainers
+  fi
+
+  if [[ $RUN_INTEGRATION_TESTS -ne 0 ]]; then
+    runIntegrationTests
```
yebrahim: combine these two `if`s?

test/spec/support/jasmine.json
jimmc: Done.
It was in "spec" because that was the default location, but I agree that using "unittests" is a better name given that we also have integration tests here.

test/run.sh
jimmc: Done. There used to be something between these two, and I didn't notice they were the same when I moved that piece elsewhere.

PR #1392, opened 2017-05-26
Use jasmine instead of mocha for remaining tests.

jimmc:


test/run.sh
```diff
@@ -101,13 +122,18 @@ function main() {
     trap cleanup INT EXIT SIGHUP SIGINT SIGTERM
   fi
 
-  # Unit tests are fast, run them all first
+  # Unit tests are fast, run them all first, and always run them
```
yebrahim: I think they should still respect the flag, if these grow to a large number the output will be noisy, making it hard to focus on another test suite.

test/notebook/test_spec.js
```diff
@@ -15,36 +15,35 @@
...
-var driver = null;
+require('jasmine2-custom-message');
 
 const notebooks_config = JSON.parse(fs.readFileSync('notebook/spec.json'))
 const timeOut = 60000;
 
-test.before(function() {
-  this.timeout(timeOut);
-  driver = new selenium.Builder()
+function createDriver() {
```
yebrahim: Did `beforeAll` not work here for building the driver once?

test/run.sh
jimmc: Done.

test/notebook/test_spec.js
jimmc: I had trouble with this before, but I tried it again and now it works. So, Done.

PR #1395, opened 2017-05-30
Implement idle-timeout shutdown.

jimmc:


sources/web/datalab/idleTimeout.ts
```diff
@@ -0,0 +1,256 @@
...
+  } else {
+    return Math.floor(secondsRemaining);
+  }
+}
+
+function shutdownIfTimedOut() {
+  if (hasTimedOut()) {
+    // TODO - check to see if the kernel is idle; if not, don't try to stop the VM
+    try {
+      logging.getLogger().debug('Idle timeout, shutting down with this command: ' + shutdownCommand);
```
yebrahim: these should probably be `info` or higher

sources/web/datalab/idleTimeout.ts
```diff
@@ -0,0 +1,256 @@
...
+    } else {
+      const parsedUrl = url.parse(request.url, true);
+      formData = parsedUrl.query;
+    }
+    logging.getLogger().debug('got form data, type=' + typeof(formData));
+    logging.getLogger().debug(formData);
+    if (formData.enabled) {
+      const newEnabled = formData.enabled == 'true' ? true : false;
+      idleTimeoutEnabled = newEnabled;
+      reset();  // Reset the timeout value when timeout is enabled or disabled
```
yebrahim: looks like this and the if statement below are redundant, since if a request is handled here it means it'll reset anyway.

sources/web/datalab/idleTimeout.ts
```diff
@@ -0,0 +1,256 @@
...
+  });
+  str = str.replace(/\s*/, '');
+  if (str) {
+    logging.getLogger().debug('garbage left over in interval string: ' + str);
+    return NaN;
+  }
+  return total;
+}
+
+export function resetBasedOnPath(path: string) {
```
yebrahim: looks like this is only called for unchecked requests, but it only does any work if the path is timeout, which means the path was checked. is this on purpose maybe to catch changes to `shouldResetOnPath`?

sources/web/datalab/static/appbar.html
```diff
@@ -75,6 +75,14 @@
           </div>
           <button id="stopVmButton" title="Stop VM" class="btn btn-danger" style="width: 100%; float: none">Stop VM</button>
         </div>
+        <li><a id="idleTimeoutEnabledButton" href="#" style="display:none;">
```
yebrahim: we should add `title` attributes for both these buttons, so the user has proper expectation before clicking

sources/web/datalab/static/idle-timeout.js
```diff
@@ -0,0 +1,264 @@
...
+  const updateTimeoutInfoNoDisplayInterval = 5 * 1000; // Keep tabs on timeout info even when not being displayed.
+  const queryTimeoutInfoInterval = 10 * 1000; // Only send a query every ten seconds.
+  let timeoutInfoTimeout;     // Result of setTimeout for calling updateTimeoutInfo
+  let timeoutInfo = {};
+  let lastUpdateTimeoutTime = 0;
+
+  let showDebugLog = false;   // Set to true in the dev console to get log messages during development.
+  function _debugLog(msg) {
+    if (showDebugLog) {
+      console.log(msg);
```
yebrahim: can we import and use `util.debug.log` instead?

sources/web/datalab/static/idle-timeout.js
```diff
@@ -0,0 +1,264 @@
...
+  function _alertIfTimedOut() {
+    if (timeoutInfo.enabled && timeoutInfo.idleTimeoutSeconds > 0) {
+      const secondsRemaining = Math.floor((timeoutInfo.expirationTime - Date.now()) / 1000);
+      if (secondsRemaining <=1 ) {
+        // If there are other handlers, let them run first.
+        const delayToLetOtherHandlersRun = 2 * 1000;
+        window.setTimeout(_alertIdleShutdown, delayToLetOtherHandlersRun);
+      }
+    }
+    timeoutInfo.enabled = false;  // Make sure we don't get a false indication later.
```
yebrahim: should this go inside the inner if statement?

sources/web/datalab/static/idle-timeout.js
```diff
@@ -0,0 +1,264 @@
...
+  // Assumes global timeoutInfo has been set.
+  function _updateTimeoutDisplay(dropdown) {
+    if (!dropdown) {
+      _debugLog('== updateTimeoutDisplay no dropdown');
+      return;
+    }
+    _debugLog('== updating timeout display');
+    _debugLog(dropdown);
+    if (timeoutInfo.enabled) {
+      let secondsRemaining = Math.floor((timeoutInfo.expirationTime - Date.now()) / 1000);
```
yebrahim: I really don't think this should be in seconds, because of the fact that the client code isn't always in sync with the server state. the user will likely see the timer go down second by second, then jump up when the server resets and the client polls the new state.
minutes makes a lot more sense to me.

sources/web/datalab/static/idle-timeout.js
```diff
@@ -0,0 +1,264 @@
...
+
+  // Sends a message to the server to enable or disable the timout.
+  function toggleIdleTimeout() {
+    const newValue = timeoutInfo.enabled ? "false" : "true";
+    const timeoutUrl = window.location.protocol + "//" + window.location.host + "/_timeout?enabled=" + newValue;
+    const dropdown = $(this).parent().parent().parent();  // Walk up to the account drop-down.
+    _debugLog('== Changing enabled to ' + newValue);
+    updateTimeoutInfo(dropdown);
+    xhr(timeoutUrl, () => {
+      timeoutInfo.enabled = newValue;   // Display the new value right away.
```
yebrahim: I'd also put the UI update here in the success callback

sources/web/datalab/static/notebook-app.js
```diff
@@ -95,6 +95,9 @@ define(['static/appbar', 'static/minitoolbar'], function(appbar, minitoolbar) {
       })
     });
 
+    idleTimeout.setupKernelBusyHeartbeat();
+    $('#mainArea').scroll(idleTimeout.notebookScrolled);
```
yebrahim: I think we'll need a different div to trigger this event in the non-notebook pages

containers/datalab/content/run.sh
```diff
@@ -104,6 +104,7 @@ if [ -n "${vm_project}" ] && [ "${vm_project}" != "no-project-id" ]; then
    export VM_PROJECT="${vm_project}"
    export VM_NAME=$(curl -s "${compute_metadata_url}/instance/hostname" -H "Metadata-Flavor: Google" | cut -d '.' -f 1)
    export VM_ZONE=$(curl -s "${compute_metadata_url}/instance/zone" -H "Metadata-Flavor: Google" | sed 's/.*zones\///')
+   export DATALAB_SHUTDOWN_COMMAND="gcloud compute instances stop ${VM_NAME} --project ${VM_PROJECT} --zone ${VM_ZONE}"
```
jimmc: You mean in the javascript code?
Putting the shutdown command here means it gets set when running in a VM but not otherwise.

sources/web/datalab/idleTimeout.ts
```diff
@@ -0,0 +1,256 @@
...
+import http = require('http');
+import logging = require('./logging');
+import net = require('net');
+import querystring = require('querystring');
+import url = require('url');
+import userManager = require('./userManager');
+
+const idleCheckIntervalSeconds = 5;    // Seconds to wait between idle checks
+
+let idleTimeoutEnabled = true;  // Allow for explicit disable without clearing other values
```
jimmc: I changed the comment on this line and the next. Let me know if this is any better.

sources/web/datalab/idleTimeout.ts
```diff
@@ -0,0 +1,256 @@
...
+let idleTimeoutSeconds = 0;   // Shutdown after being idle this long; disabled if 0 or NaN
+let lastUserActivity: number;  // The epoch, in seconds, of the last user activity
```
jimmc: I renamed this variable to be more reflective of reality. This module doesn't define what user activity is.

sources/web/datalab/idleTimeout.ts
```diff
@@ -0,0 +1,256 @@
...
+    if (!idleTimeoutSeconds) {
+      idleTimeoutSeconds = parseInterval(process.env.DATALAB_IDLE_TIMEOUT);
+    }
+    logging.getLogger().debug('Idle timeout set to ' + idleTimeoutSeconds + ' seconds');
+    if (! (idleTimeoutSeconds > 0)) {
+      logging.getLogger().debug('No idle timeout value, idle timeout is disabled');
+    }
+  } else {
+    idleTimeoutSeconds = NaN;   // No shutdown command available
+    logging.getLogger().debug('No shutdown command, idle timeout is disabled');
```
jimmc: Done.

sources/web/datalab/idleTimeout.ts
```diff
@@ -0,0 +1,256 @@
...
+  }
+}
+
+// Parse a string like '30s' or '1h 15m 5s' and return seconds.
+// If no input string, or unrecognized stuff, returns NaN.
+function parseInterval(str: string) {
```
jimmc: Sounds reasonable. What are the rules regarding adding new packages to what we ship? Licensing guidelines?

sources/web/datalab/idleTimeout.ts
jimmc: Not sure what you mean by unchecked requests. This calls reset for anything other than _timeout requests.

sources/web/datalab/idleTimeout.ts
```diff
@@ -0,0 +1,256 @@
...
+}
+
+export function resetBasedOnPath(path: string) {
+  if (shouldResetOnPath(path)) {
+    reset();
+  }
+}
+
+function shouldResetOnPath(path: string) {
+  logging.getLogger().debug('shouldReset sees path=' + path);
```
jimmc: Removed.

sources/web/datalab/idleTimeout.ts
```diff
@@ -0,0 +1,256 @@
...
+}
+
+function nowSeconds() {
+  return Date.now() / 1000;   // seconds since the epoch
+}
+
+function hasTimedOut() {
+  if (isTimeoutEnabled()) {
+    const idleSeconds = nowSeconds() - lastUserActivity;
+    logging.getLogger().debug('idleSeconds=' + Math.ceil(10*idleSeconds)/10 + ', idleTimeoutSeconds=' + idleTimeoutSeconds);
```
jimmc: Removed.

sources/web/datalab/idleTimeout.ts
jimmc: Done.

sources/web/datalab/idleTimeout.ts
jimmc: Not quite. If the user submits a request that does not include a change to the enabled flag or a reset, it will not reset.

sources/web/datalab/static/appbar.html
jimmc: Done.

sources/web/datalab/static/idle-timeout.js
jimmc: Done.

sources/web/datalab/static/idle-timeout.js
jimmc: No, I want it to be disabled any time this method is called. Changed the method name to be clear about that.

sources/web/datalab/static/idle-timeout.js
jimmc: A few lines below, the call to roundToApproximateTime will generally round to minutes or larger chunks. The user will only see seconds when the time gets down to the last couple of minutes. I find seeing the seconds to be useful for development work. I expect users will rarely see that.

sources/web/datalab/static/idle-timeout.js
jimmc: It is there a couple of lines below. Or did you mean something besides a call to updateTimeoutInfo?

sources/web/datalab/static/notebook-app.js
jimmc: I haven't played around as much with non-notebook pages. The list page is very reset-happy, it seems to refresh the list of notebooks pretty often, so I didn't bother adding anything else there.

sources/web/datalab/idleTimeout.ts
yebrahim: MIT license is generally fine.
@Di-Ku just in case.

containers/datalab/content/run.sh
yebrahim: but what's the value of this? we can already check if we're in a VM.

sources/web/datalab/idleTimeout.ts
yebrahim: Never mind, I missed this was actually called before handling requests.

sources/web/datalab/static/notebook-app.js
yebrahim: Yea I think you're right about the list page, the sessions page also uses the same code so it'll do the same refresh requests.
We have to think about other pages though, for now we only have the editor and terminal pages, which both sound like the user can spend some time on. Can we make sure we have those covered?

test/unittests/static/idle-timeout_spec.js
```diff
@@ -0,0 +1,97 @@
...
+
+  describe('secondsToString', function() {
+    it('converts zero and negative to empy string', function() {
+      expect(idleTimeout._secondsToString(0)).toEqual('');
+      expect(idleTimeout._secondsToString(-1)).toEqual('');
+    });
+    it('converts single-unit strings correctly', function() {
+      expect(idleTimeout._secondsToString(1)).toEqual('1s');
+      expect(idleTimeout._secondsToString(15)).toEqual('15s');
+      expect(idleTimeout._secondsToString(2*60)).toEqual('2m');
```
yebrahim: one more test with an unrounded number, e.g. 2*60+1 or 3*60*60+61?

containers/datalab/content/run.sh
jimmc: By using this environment variable, it is easy for me to set a different shutdown command in my startup.sh that shuts down my local notebook server during development. If you prefer, I can remove this line from this script, and instead add code to idleTimeout.ts that calls getVmInfo, checks to see if we are on a VM, and builds this shutdown command if so.

sources/web/datalab/static/notebook-app.js
jimmc: I will take a look at those pages.

test/unittests/static/idle-timeout_spec.js
jimmc: Multi-unit outputs are in the next test, a few lines lower. Is there something you want me to add there?

containers/datalab/content/run.sh
yebrahim: I see, no I think it's fine to stick with the env var if it's more convenient in code.

test/unittests/static/idle-timeout_spec.js
yebrahim: Nope, I think it's all covered.

PR #1401, opened 2017-06-02
Test fixes

yebrahim:
Fixes to make tests more reliable:
- Fix the promise chain
- Make tests clean up after themselves (especially ones that create new files)
- Increase default test timeout (`jasmine.DEFAULT_TIMEOUT_INTERVAL`)


test/ui/test_spec.js
```diff
@@ -96,29 +94,42 @@ describe('UI tests', function() {
...
+        .then(function() {
+          return done();
+        });
     });
 
     it('appears correctly after closing help menu by clicking the body element', function(done) {
       return driver.findElement(By.tagName('body'))
         .then(function(button) {
-          button.click();
+          return button.click();
```
jimmc: If you use the new syntax, I think you could write these lines like this:
return driver.findElement(By.tagName('body'))
  .then(button => button.click())
  .then(() => screenshotAndCompare('body.png', 'body'))
  .then(done);
Just a style issue - your choice.
The same comment applies for the other changes you made below, where there are now a bunch of one-line functions that could be rewritten as actual one-lines.

test/ui/test_spec.js
```diff
@@ -81,7 +79,7 @@ describe('UI tests', function() {
     beforeAll(function(done) {
       return driver.get('http://localhost:8081/tree/datalab')
         .then(function() {
-          driver.wait(until.titleIs('Google Cloud DataLab'), 5000);
+          return driver.wait(until.titleIs('Google Cloud DataLab'), 5000);
```
jimmc: Nice catch.

test/ui/test_spec.js
```diff
@@ -96,29 +94,42 @@ describe('UI tests', function() {
     });
 
     it('appears correctly before any actions have been taken', function(done) {
-      return screenshotAndCompare('body.png', 'body', done);
+      screenshotAndCompare('body.png', 'body')
```
jimmc: I find it interesting that in test_spec.js you move 'done' down into the function being called, whereas here you are doing just the opposite.

test/ui/test_spec.js
```diff
@@ -96,29 +94,42 @@ describe('UI tests', function() {
...
+        .then(function() {
```
jimmc: Even if you don't choose to collapse all the one-liners, I think you can still write all these calls to done as
  .then(done);

test/ui/test_spec.js
yebrahim: This done wasn't adding anything so I took it out completely. Its sole purpose was making sure we wait for the screenshot to be compared and written to disk (if incorrect) before continuing, but we don't need that for the correctness of the tests, we only need to make sure the `driver.takeScreenshot` call is done.

test/ui/test_spec.js
yebrahim: sure, switched to one-liners.

test/ui/test_spec.js
yebrahim: done.

test/ui/test_spec.js
```diff
@@ -80,14 +78,12 @@ describe('UI tests', function() {
...
-          driver.wait(until.titleIs('Google Cloud DataLab'), 5000);
-        })
-        .then(function() {
+        .then(() => driver.wait(until.titleIs('Google Cloud DataLab'), 5000))
+        .then(() => {
           // wait for the Datalab page to finish loading
           return driver.wait(function() {
             return driver.executeScript('return window.datalab.loaded')
-              .then(function(loaded) {
+              .then(loaded => {
```
jimmc: You can write this as
  .then(loaded => loaded === true);

yebrahim: I think this is one instance of this syntax where it's not very clear :)

PR #1403, opened 2017-06-02
Move xhr and reportEvent functions into utils.js.

jimmc:
This moves a couple more functions into utils.js as a cleanup.
It also makes settings.js a module. Unfortunately, I couldn't figure out how to late-load settings.js as a module. If you happen to know how to do that, please let me know.

PR #1404, opened 2017-06-02
Add ui tests for Settings.

jimmc:
After this is in, I'll go back to #1403.


test/ui/test_spec.js
```diff
@@ -88,7 +88,9 @@ describe('UI tests', function() {
               });
           }, 10000);
         })
-        .finally(() => done());
+        .then(driver.executeScript("window.datalab.vminfo.vm_name='fake_vm_name_for_testing';" +
+              "require(['static/appbar'],appbar=>appbar.setAccountMenuVminfo());"))
```
yebrahim: huh. guess that works too!

yebrahim: this should probably go in a `beforeAll` block though, we might have other tests that rely on it.

yebrahim: also, if we're on a VM, we shouldn't have the sign in button, so maybe we really should fool the container into thinking it's not running on a VM, just something to think about.

test/ui/test_spec.js
```diff
@@ -168,6 +168,37 @@ describe('UI tests', function() {
         .then(done);
     });
 
+    it('Settings page defaults to light theme', function(done) {
+      driver.findElement(By.id('accountDropdownButton')).click()
+        .then(() => screenshotAndCompare('settingsMenu.png', 'settingsMenu', null))
```
yebrahim: `screenshotAndCompare` doesn't take the done param anymore

test/ui/test_spec.js
```diff
@@ -168,6 +168,37 @@ describe('UI tests', function() {
...
+        .then(() => driver.findElement(By.id('settingsButton')).click())
+        // After clicking on the Settings menu item, wait for the dialog.
+        .then(() => driver.wait(until.elementLocated(By.id('lightThemeRadioOption'))))
+        // Wait for the fade-in to complete.
+        .then(() => driver.sleep(1000))
+        .then(() => screenshotAndCompare('settingsLightTheme.png', 'settingsLightTheme', null))
+        // Close the dialog.
+        .then(() => driver.findElement(By.xpath('//button[@data-dismiss="modal"]')).click())
```
yebrahim: let's also grab a screenshot here to make sure the dialog closes after this click

test/ui/test_spec.js
```diff
@@ -168,6 +168,37 @@ describe('UI tests', function() {
...
     });
 
+    it('Settings page defaults to light theme', function(done) {
+      driver.findElement(By.id('accountDropdownButton')).click()
+        .then(() => screenshotAndCompare('settingsMenu.png', 'settingsMenu', null))
+        .then(() => driver.findElement(By.id('settingsButton')).click())
+        // After clicking on the Settings menu item, wait for the dialog.
+        .then(() => driver.wait(until.elementLocated(By.id('lightThemeRadioOption'))))
+        // Wait for the fade-in to complete.
+        .then(() => driver.sleep(1000))
```
yebrahim: do we really need this? is the test flaky without it? if so, is there maybe a better way?

test/ui/test_spec.js
jimmc: Yeah, driver.execute returns a promise, so doesn't need to be wrapped in a function if that's the only call in the then arg.
This is already in the beforeAll block.
Yeah, I guess I could fix the sign-in button as well. It's a bit of a hack the way it is currently done, but at least it's the same on local and travis. :-)

test/ui/test_spec.js
jimmc: We need some kind of delay to wait for the fade-in to complete. There is probably a way to do it with a driver.wait and some condition, but after I spent ten minutes playing with it I decided this was good enough. If you happen to know of an event or something that we can catch when the fade-in completes, that could be a reasonable way to deal with it. Or maybe we could disable the fade-in.

PR #1409, opened 2017-06-07
Allow the base URL to be configurable.

### LINKED ISSUE #1408: Allow Datalab to run at base paths other than just "/"
Right now Datalab assumes that it is the only server running on a given host/port.

That is true when running Datalab with the supported command line tools, but the Docker image is public and should be able to be run in other environments/setups.

In particular, one compelling use case is to be able to run Datalab behind a reverse proxy that forwards to one of multiple servers (e.g. multiple Datalab instances) based on URL path.

Such a use case could be implemented by rewriting URLs in the proxy, but that won't make Datalab produce links to the correct, nested URLs when it outputs HTML and/or JS.

A better solution would be to add a setting to Datalab that specifies the base URL, and to have it produce the right URLs in any outputs it produces.

ojarjur:
This change adds a 'datalabBasePath' setting that allows setting a base for all Datalab URL paths.

This change is implemented in multiple parts:

1. Make the server support all incoming requests (both regular requests and websocket requests) that use the configured base path by stripping that prefix out of all requests before further processing.
2. Make the proxied servers (Jupyter and Ungit) also support this base path by passing it along to them when they are started.
3. Make the Datalab NodeJS server include this prefix in all redirect URLs that it generates.
4. Make the template files used to generate HTML include this base path as a variable, so that all of their embedded links can include the base path.
5. Make all URLs generated by the Javascript client code include the base path.
6. Make all of the client-side Javascript code that parses window.location strip the base path if it is present.
7. Remove absolute URLs from the contents of everything under the `static` directory; replacing them with URLs generated by our Javascript code if necessary.

This fixes #1408


containers/datalab/content/run.sh
```diff
@@ -115,7 +115,8 @@ then
 fi
 
 # Start the ungit server
-ungit --port=8083 --no-launchBrowser --forcedLaunchPath=/content/datalab --ungitVersionCheckOverride 1> /dev/null &
+DATALAB_BASE_PATH=$(echo ${DATALAB_SETTINGS_OVERRIDES} | python -c "import sys,json; print(json.load(sys.stdin).get('datalabBasePath',''))")
```
chmeyers: What does this do?

sources/web/datalab/sockets.ts
```diff
@@ -165,10 +166,11 @@ function socketHandler(socket: SocketIO.Socket) {
 
 export function init(settings: common.Settings): void {
   var io = socketio(String(settings.socketioPort), {
+    path: path_.join(settings.datalabBasePath, '/socket.io'),
```
jimmc: You already ensured that datalabBasePath ends with a slash (in settings.ts), so you don't need one here. Likewise in other places where you append to datalabBasePath.

sources/web/datalab/static/util.js
```diff
@@ -46,9 +46,30 @@ define(() => {
     request.send();
   }
 
+  // Generate an absolute link to a path on this same Datalab instance.
+  function datalabLink(fullPath) {
+    var basePath = document.body.getAttribute('data-base-url');
```
jimmc: Seems like this should be datalab-base-url rather than data-base-url.

sources/web/datalab/static/util.js
```diff
@@ -46,9 +46,30 @@ define(() => {
...
+      return '/' + fullPath.slice(basePath.length);
+    }
+    return fullPath;
+  }
+
   return {
     debug,
     reportEvent,
     xhr,
+    datalabLink,
```
jimmc: Maybe put these functions (and their defs) in alphabetical order?

sources/web/datalab/login.ts
```diff
@@ -1,109 +0,0 @@
-/*
```
ojarjur: You're absolutely right.

I've sent out https://github.com/googledatalab/datalab/pull/1414 to remove it separately.

sources/web/datalab/common.d.ts
```diff
@@ -62,6 +62,13 @@ declare module common {
     datalabRoot: string;
 
     /**
+     * If provided, use this as a prefix to all URL paths. This is useful
```
ojarjur: Done.

sources/web/datalab/common.d.ts
```diff
@@ -62,6 +62,13 @@ declare module common {
...
+     * for running a Datalab instance behind a shared proxy with other
```
ojarjur: Done.

sources/web/datalab/config/settings.json
```diff
@@ -12,6 +12,7 @@
   "nextJupyterPort": 9000,
   "socketioPort": 8084,
   "datalabRoot": "",
+  "datalabBasePath": "",
```
ojarjur: No need; the "/" is added by the server if it is missing.

sources/web/datalab/jupyter.ts
```diff
@@ -139,7 +139,8 @@ function createJupyterServerAtPort(port: number, userId: string, userDir: string
     '--port-retries=0',
     '--notebook-dir="' + server.notebooks + '"',
     '--NotebookNotary.algorithm=sha256',
-    '--NotebookNotary.secret_file=' + secretPath
+    '--NotebookNotary.secret_file=' + secretPath,
+    '--NotebookApp.base_url=' + appSettings.datalabBasePath
```

sources/web/datalab/server.ts
```diff
@@ -315,8 +332,9 @@ export function run(settings: common.Settings): void {
     new wsHttpProxy.WsHttpProxy(server, httpOverWebSocketPath, settings.allowOriginOverrides);
   }
 
-  logging.getLogger().info('Starting DataLab server at http://localhost:%d',
-                           settings.serverPort);
+  logging.getLogger().info('Starting DataLab server at http://localhost:%d/%s',
```
ojarjur: Done.

sources/web/datalab/static/util.js
ojarjur: `data-base-url` is the already-defined attribute used by the Jupyter JS code, so we want to reuse that.

sources/web/datalab/static/util.js
ojarjur: Done

sources/web/datalab/server.ts
```diff
@@ -267,19 +272,31 @@ function stopVmHandler(request: http.ServerRequest, response: http.ServerRespons
...
   // Avoid proxying websocket requests on this path, as it's handled locally rather than by Jupyter.
   if (request.url != httpOverWebSocketPath) {
     jupyter.handleSocket(request, socket, head);
   }
 }
 
+function trimBasePath(path: string): string {
+  let pathPrefix = appSettings.datalabBasePath;
+  if (path.indexOf(pathPrefix) == 0) {
+    let newPath = "/" + path.substring(pathPrefix.length);
```
jimmc: Or you could skip the "/"+ part and instead subtract 1 from the length you are passing to substring, since you know the pathPrefix ends with a slash. Your choice.

ojarjur: That's a fair point.

I think this current approach looks slightly more readable to me, but neither is ideal.

PR #1410, opened 2017-06-07
Rebuild web sources on container build

yebrahim:


sources/build.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
```
ojarjur: I believe the omission of the -e flag was intentional, so that we could build multiple directories under 'sources/' and not bail out on the first failure.

That decision may be worth revisiting, but I think it should be done in its own PR and will expand to include a lot of other cleanups (like removing the for loop that just iterates over a single sub directory).

yebrahim: done.

PR #1411, opened 2017-06-07
Disable debug logs by default

yebrahim:


sources/web/datalab/static/util.js
```diff
@@ -1,7 +1,14 @@
 define(() => {
+
+  window.datalab = window.datalab || {};
+  window.datalab.debug = false;
```
ojarjur: How does someone enable this when they want debug output?

yebrahim: You type `window.datalab.debug = true` in your browser console. It's not ideal, but a good fix until we figure out a build flag or something more proper.

ojarjur: That doesn't seem sufficient if you need to debug something that happens when the page first loads.

I would suggest something like a 'data-debug-enabled' attribute to the body of all of our templates and then set that based on an environment variable when the template gets turned into HTML

yebrahim: This seems reasonable. However for now we need something to hide the verbose logs we have on the client side.

jimmc: In general, I think it is OK to set the default false like this; in the case where a developer needs to see debug output during page load, he can edit the source code before reloading.
Another possibility would be to add a bit of code here to get location, grab the query parameters, and look for debug=true or something like that.

ojarjur: I think both things can be done in a single pull-request.

yebrahim: Done. Using a querystring `debug=true` now.

PR #1412, opened 2017-06-07
Map $HOME/datalab as /content/datalab

yebrahim:
We get a performance hit when starting the server on OSX if the home directory contains a lot of files, because of a limitation on the package we're using to index the file system (see https://github.com/paulmillr/chokidar/issues/597).

Since this script is used for the dev workflow, we can use a new directory `$HOME/datalab_home` that will only contain datalab files, just like the case on a VM.


yebrahim:
Done.

PR #1415, opened 2017-06-08
Tag the config_local file with a static tag as well.

chmeyers:
We don't tag the CLI in the same way as it is only ever picked up manually.


tools/release/cloudbuild.yaml
```diff
@@ -120,14 +111,31 @@ steps:
...
+  waitFor: ['pushConfigLocal', 'buildDatalab', 'buildDatalabGPU', 'buildGateway']
+
+# Cleanup the temporary files created to push build artifacts to GCS
+- name: 'debian'
+  args: ['rm', '-rf',
+         '/workspace/config_local.js',
+         '/workspace/config_local_${REVISION_ID}.js',
+         '/workspace/datalab-cli-${REVISION_ID}.tgz']
+  id:   'cleanupWorkspace'
+  waitFor: ['pushConfigLocal', 'pushTarball', 'pushConfigLocalStatic']
```
ojarjur: This was intentionally being done prior to the Docker builds so that we would not have any unwanted artifacts that could accidentally get copied into one of the Docker images.

Instead of moving it down here, just make the `gsutil` command copy directly from one location in the bucket to another (i.e. you make both the source and target remote files.

chmeyers: Done

PR #1416, opened 2017-06-09
Ground work for experimental UI with Polymer+TS

yebrahim:
This PR adds the infrastructure needed to build a Polymer 2.0 front end with Typescript. It adds an environment variable `DATALAB_EXPERIMENTAL_UI` that can be passed to the container to turn on this UI. The new frontend resources are kept under `sources/web/datalab/polymer/*`, and are built into `build/static/experimental/*` so the old static resources aren't changed.

Two new build dependencies are introduced: `bower`, which is currently the main package manager used by Polymer, and `polymer-cli`, which is a toolbox for creating, linting, building, and serving polymer apps. These can probably be factored into a single build system later, but for now, I'm using the cli to build the static files, and optionally bundling/minifying them.

This PR breaks off a piece of the UI I'm building to make code review easier, so only one element is added here: the datalab shell element, which will contain other elements introduced later.

### Typescript
As of now, there's officially no good solution for building Polymer 2.0 apps with Typescript, so the solution I have isn't perfect. I added a type declaration file (`third_party/externs/ts/polymer/polymer.d.ts`) that I adapted from Polymer 1.0 that seems to be covering all the main scenarios I tested, and referenced that file in the `tsconfig` file so those types are seen by all elements without explicitly referencing it. I also could not find a module loader that will work with this setup, since the javascript generated by the `tsc` transpiler using ES6 as target (which is needed by Polymer 2.0) uses es6 modules, which aren't natively supported by any browser yet, and will need a polyfill. So for now, **we'll have to reference the javascript files explicitly in the HTML of the element importing them**.

Finally, please keep in mind the `datalab-app` element is meant as a placeholder not a full component, it currently looks like this:

screen shot 2017-06-08 at 11 16 42 pm


sources/web/build.sh
```diff
@@ -29,12 +29,22 @@ WEB_DIR=$BUILD_DIR/web/nb
 
 mkdir -p $WEB_DIR
 
+# Experimental UI build step
+cd datalab/polymer
+bower install
+tsc
+polymer build
+rsync -avpq ./build/experimental/ ../static/experimental
```
ojarjur: This line is redundant (the copy is performed again below with the rest of the copy commands)

sources/web/datalab/static.ts
```diff
@@ -159,13 +160,50 @@ function sendUserCustomTheme(userId: string, response: http.ServerResponse): voi
...
  */
 function requestHandler(request: http.ServerRequest, response: http.ServerResponse): void {
   var path = url.parse(request.url).pathname;
 
+  // -------------------------------- start of experimental UI resources
+  if (isExperimentalResource(path)) {
+    console.log('Serving experimental UI resource: ' + path);
+    if (path === '/') {
+      response.statusCode = 302;
+      response.setHeader('Location', '/files');
```
ojarjur: Change `'/files'` to `util.datalabLink('/files')` so that the base URL path gets used if it is non-empty.

sources/web/build.sh
yebrahim: Actually this one is necessary, and the one below is the redundant one. I'll remove that.

sources/web/datalab/static.ts
yebrahim: Isn't that taken care of by the client-side code?

ojarjur: No, the path prefix needs to be added here.

This is a redirect response sent by the server. Client-side code does not affect it.

ojarjur: Sorry, my earlier suggestion was wrong...

This should be `path_.join(appSettings.datalabBasePath, '/files')`... since the util.datalabLink method is only in the client-side Javascript.

sources/web/datalab/server.ts
```diff
@@ -244,7 +244,9 @@ function uncheckedRequestHandler(request: http.ServerRequest, response: http.Ser
       urlpath.indexOf('/oauthcallback') == 0) {
     // Start or return from auth flow.
     auth.handleAuthFlow(request, response, parsed_url, appSettings);
-  } else if ((urlpath.indexOf('/static') == 0) || (urlpath.indexOf('/custom') == 0)) {
+  } else if (urlpath.indexOf('/static') == 0 ||
```
jimmc: Maybe you should have a method static_.isStaticResource(urlpath) that does this three-way check.

sources/web/datalab/static.ts
```diff
@@ -159,14 +160,51 @@ function sendUserCustomTheme(userId: string, response: http.ServerResponse): voi
...
  * @param request the incoming file request.
  * @param response the outgoing file response.
  */
 function requestHandler(request: http.ServerRequest, response: http.ServerResponse): void {
-  var path = url.parse(request.url).pathname;
+  var pathname = url.parse(request.url).pathname;
+
+  // -------------------------------- start of experimental UI resources
+  if (isExperimentalResource(pathname)) {
+    console.log('Serving experimental UI resource: ' + pathname);
```
jimmc: Consider using logging.getLogger().debug()

sources/web/datalab/static.ts
```diff
@@ -159,14 +160,51 @@ function sendUserCustomTheme(userId: string, response: http.ServerResponse): voi
...
+    if (pathname === '/') {
+      response.statusCode = 302;
+      response.setHeader('Location', path.join(appSettings.datalabBasePath, '/files'));
```
jimmc: No slash on files, datalabBasePath always begins and ends with a slash.

sources/web/datalab/static.ts
```diff
@@ -159,14 +160,51 @@ function sendUserCustomTheme(userId: string, response: http.ServerResponse): voi
 }
 
 /**
+ * Returns true if this path should return an experimental UI resource
+ * @param path the incoming request path
+ */
+export function isExperimentalResource(pathname: string) {
+  const experimentalUiEnabled = process.env.DATALAB_EXPERIMENTAL_UI;
```
jimmc: Is this the only place that DATALAB_EXPERIMENTAL_UI will be referenced in the ts code? If there might be other places where you have to make this check, I think I'd rather see a method experimentalUiIsEnabled() that returns true if that env var is set, and have all checks go through that.

sources/web/datalab/server.ts
yebrahim: done.

sources/web/datalab/static.ts
yebrahim: this is the only place, I think I'd like to keep it this way since it's meant as temporary code to be removed, I don't want to flesh it out too much.

sources/web/datalab/static.ts
yebrahim: done.

sources/web/datalab/static.ts
yebrahim: done.

sources/web/datalab/server.ts
```diff
@@ -228,6 +228,16 @@ function handleRequest(request: http.ServerRequest,
 }
 
 /**
+ * Returns true iff the supplied path should be handled by the static handler
+ */
+function isStaticResource(urlpath: string) {
```
jimmc: It would make a bit more sense to put this in static.ts. Perhaps in the next PR.

PR #1418, opened 2017-06-09
Add live mode for pydatalab.

jimmc:
This PR speeds up the development cycle in cases where pydatalab code must be tested in Datalab. See the comments in the diff of containers/datalab/run.sh for how to use it. If you have an idea for a better approach, please let me know.


containers/datalab/content/run.sh
```diff
@@ -54,6 +54,21 @@ check_tmp_directory() {
     echo "The /tmp directory is writable"
 }
 
+# Reinstall the parts of pydatalab from /content/pydatalab, which is where it gets live-mounted.
+function reinstall_pydatalab() {
+  PYDATALAB=/content/pydatalab
+  echo "Reinstalling pydatalab from ${PYDATALAB}"
+  pip uninstall -y datalab
```
yebrahim: I don't think you need the uninstall, `--force-reinstall` or `--upgrade` will overwrite.

containers/datalab/run.sh
```diff
@@ -65,6 +82,7 @@ fi
 docker run -it --entrypoint=$ENTRYPOINT \
   -p 127.0.0.1:8081:8080 \
   -v "$CONTENT/datalab:/content/datalab" \
+  $PYDATALAB_MOUNT_OPT \
```
yebrahim: there was [another change that attempted this](https://github.com/googledatalab/datalab/blob/master/containers/base/build.sh#L18) during the build process, I think keeping both is redundant and might lead to a misunderstanding, should we remove that one, since this one doesn't require rebuilding the image?

containers/datalab/content/run.sh
jimmc: I tried just using --force-install and it didn't seem to be working. I can take another look.

yebrahim: I always got it working with `--upgrade/-U`

containers/datalab/run.sh
jimmc: That other line is not quite the same: it takes a snapshot of pydatalab and uses it when building the container. Once that snapshot has been taken, changes to the original pydatalab dir have no affect on the container. This new --pydatalab option essentially overwrites whatever pydatalab is in the container with the specified pydatalab when the container starts, but if you then rerun the container without the option, you will be back to whatever pydatalab the container was built with.

Having said that, I don't do enough pydatalab development to know if that old mode might still be useful, so I am OK either way. Given this explanation, do you think I should remove the old approach?

yebrahim: I think the build-step option becomes unnecessary once this is in, since this change sort of covers that other workflow. I'll let @qimingj chime in on this since he's the one who added it IIRC.

containers/datalab/content/run.sh
```diff
@@ -54,6 +54,21 @@ check_tmp_directory() {
...
+  pip3 uninstall -y datalab
+  pip install --force-reinstall --no-cache-dir ${PYDATALAB}
+  pip install --force-reinstall ${PYDATALAB}/solutionbox/image_classification/.
+  pip install --force-reinstall ${PYDATALAB}/solutionbox/structured_data/.
+  pip3 install --force-reinstall --no-cache-dir ${PYDATALAB}
+  pip3 install --force-reinstall ${PYDATALAB}/solutionbox/image_classification/.
```
qimingj: No need to pip3 install MLToolbox packages because they depend on DataFlow, which doesn't support py3 yet.

containers/datalab/run.sh
qimingj: The old way is useful in cases where you want to build an image with custom pydatalab. For example, I want to add a new experimental feature in pydatalab (not in master yet) and I want to build a container with that feature and give it to a few users to try. Is the scenario still valid?

containers/datalab/content/run.sh
jimmc: This was directly copied from the Dockerfile, where it does the same thing. chmeyers suggested keeping them in in both places, and it will start working once DataFlow supports py3.

containers/datalab/run.sh
jimmc: That seems like a valid use case to me. I'm happy to leave it in.

containers/datalab/content/run.sh
qimingj: Sounds good.

containers/datalab/content/run.sh
jimmc: OK, I took out the "pip uninstall", it still logs that it is uninstalling before it installs, so that looks like it is working.

jimmc:
I believe I have responded to all the comments, and am ready for another round of review.

PR #1419, opened 2017-06-12
Add datalab-toolbar and datalab-sidebar elements to Polymer UI

yebrahim:
This PR fleshes out the toolbar and sidebar elements, and adds placeholder elements for the files and sessions pages. It also completes the routing to lazy load pages as they're clicked.

image

The settings and info buttons currently open empty dialogs, and the profile button opens an empty dropdown as well.


sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -77,6 +80,20 @@ class DatalabAppElement extends Polymer.Element {
     this.page = page || 'files';
   }
 
+  /**
+   * on changes to the page property, resolve the new page's uri, and
```
jimmc: Please use sentence case (initial caps in these two lines) and punctuation (period on next sentence) in full-sentence comments. Also, function comments should use verb phrases as would appear after the implied "This function...", so in this case, "resolves" rather than "resolve", "tells" rather than "tell".

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
```diff
@@ -0,0 +1,72 @@
+<link rel="import" href="../../components/shared-styles/shared-styles.html">
+
+<link rel="import" href="../../bower_components/iron-selector/iron-selector.html">
```
jimmc: order?

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
```diff
@@ -0,0 +1,72 @@
...
+<link rel="import" href="../../bower_components/iron-icons/editor-icons.html">
+<link rel="import" href="../../bower_components/iron-icons/hardware-icons.html">
+<link rel="import" href="../../bower_components/iron-icons/av-icons.html">
+<link rel="import" href="../../bower_components/paper-button/paper-button.html">
+
+<dom-module id="datalab-sidebar">
+  <template>
+    <style include="datalab-shared-styles">
+      a, u {
+        text-decoration: none;
```
jimmc: Weren't you just complaining about how annoying it is to have invisible anchors? What's the intent here?

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
```diff
@@ -0,0 +1,119 @@
...
+<link rel="import" href="../../bower_components/app-layout/app-toolbar/app-toolbar.html">
+<link rel="import" href="../../bower_components/iron-icons/iron-icons.html">
+<link rel="import" href="../../bower_components/paper-icon-button/paper-icon-button.html">
+<link rel="import" href="../../bower_components/paper-dialog/paper-dialog.html">
+<link rel="import" href="../../bower_components/paper-button/paper-button.html">
+
+<dom-module id="datalab-toolbar">
+  <template>
+    <style include="datalab-shared-styles">
+      :host {
```
jimmc: How is the order of these stanzas determined?

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
```diff
@@ -0,0 +1,119 @@
...
+      }
+      .dialog-content {
+        flex-grow: 1;
+      }
+      .account-dropdown {
+        position: absolute;
+        right: 0px;
+        top: var(--toolbar-height);
+        margin: 0px;
+        min-width: 160px;
```
jimmc: Do these pixel-based sizes work properly when the user zooms the page?

sources/web/datalab/polymer/components/datalab-app/datalab-app.html
```diff
@@ -1,31 +1,32 @@
 <link rel="import" href="../../bower_components/polymer/polymer.html">
```
ojarjur: All of the html files here need a proper copyright and license header.

sources/web/datalab/polymer/bower.json
```diff
@@ -3,7 +3,13 @@
   "main": "index.html",
   "dependencies": {
     "polymer": "Polymer/polymer#^2.0.0",
+    "paper-icon-button": "PolymerElements/paper-icon-button#^2.0.0",
     "iron-pages": "PolymerElements/iron-pages#^2.0.0",
-    "app-route": "PolymerElements/app-route#^2.0.0"
+    "app-route": "PolymerElements/app-route#^2.0.0",
```
yebrahim: done.

sources/web/datalab/polymer/components/datalab-app/datalab-app.html
yebrahim: done.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -16,9 +16,11 @@
  * Shell element for Datalab.
  * It contains a <datalab-toolbar> element at the top, a <datalab-sidebar>
  * element beneath that to the left, and a paged view to switch between
- * different pages. It will hold references to the different page elements,
- * and uses a local router element to switch between these according to the
- * current page location
+ * different pages. It holds references to <datalab-files> and
+ * <datalab-sessions>, and uses a local router element to switch between
+ * these according to the current page location
```
yebrahim: done.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -53,7 +55,8 @@ class DatalabAppElement extends Polymer.Element {
     return {
       page: {
         type: String,
-        value: 'files'
+        value: 'files',
+        observer: '_pageChanged'
```
yebrahim: done.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: done, and done.
So formal. :)

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
yebrahim: done.

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
yebrahim: I was complaining about hard to find links, this is different. If we ever want to use raw anchors we should change this, but I've never seen that in material design, everything is done using buttons, or otherwise styled links.

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
yebrahim: I try to keep them roughly in the same order as the elements in the dom, but sometimes I miss something. How important do you think ordering is here?

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
yebrahim: Yes, using explicit sizes is the best I know to keep the design tight, and it works fine with zooming.

sources/web/datalab/polymer/components/datalab-files/datalab-files.html
```diff
@@ -0,0 +1,25 @@
...
+<dom-module id="datalab-files">
+  <template>
+
+    <style>
+    </style>
+
+    <div>Files</div>
+
+  </template>
+< /dom-module>
```
ojarjur: It looks like there is an extra space here.

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
jimmc: OK, makes sense.

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
jimmc: It's minor. Could become more important later if more are added. Using a sort order of "sorted by implementation date" is generally not useful. If they are sorted according to where they appear on the screen, that's OK. If that's the case, a comment might be useful.

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
jimmc: Cool, thanks.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -43,7 +45,7 @@ class DatalabAppElement extends Polymer.Element {
   constructor() {
     super();
 
-    // set the pattern once to be the current document pathname
+    // Sets the pattern once to be the current document pathname
```
jimmc: No need to use the third-person version form ("Sets") here, as it is not a function comment. One of the reasons to do so on functions is to avoid any abiguity that the comment might be a command to the caller of the function.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -62,21 +65,35 @@ class DatalabAppElement extends Polymer.Element {
 
   static get observers() {
     return [
-      // need a complex observer for changes to the routeData
+      // We need a complex observer for changes to the routeData
```
jimmc: Please add trailing periods on block comments that use full sentences.

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/**
+ * Sidebar element for Datalab.
+ * This element puts a side bar on the left side that contains links to
+ * different pages, and exposes a two-way bound 'page' property
```
jimmc: add period

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
yebrahim: I generally haven't seen much benefit of following a specific order here, so I'm not sure we want to maintain that anyway, I also think it's not easy to always pay attention to that when editing the file.

sources/web/datalab/polymer/components/datalab-files/datalab-files.html
yebrahim: done.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: agreed.

PR #1425, opened 2017-06-13
Make default timeout 90m and add UI settings field for it.

jimmc:
This PR sets the auto-shutdown idle timeout interval to 90 minutes, making it on by default for all users.

It also adds a text box in the Settings dialog to allow users to change the value or to disable it by setting it to zero. As the user enters text into the box, it adds a message if the format is not valid or if the value is too small. In these cases, the Update button is disabled. The Update button starts out disabled. Once the value has been changed and is valid, the Update button is enabled. When the user clicks Update, the timeout value is updated in the server and the dialog closes.

The new timeout field as initially displayed, with the current timeout value:
timeoutfield

Displaying an error message for an invalid format:
timeoutfieldinvalid

Displaying an error message for a value that is too small:
timeoutfieldtoosmall

After a valid new value has been entered, with the Update button enabled:
timeoutfieldenabled

New screenshot after shortening the prompt and adding a tooltip icon: 
autoshutdownupdated

... and with the tooltip displaying (the cursor, invisible in this screenshot, is over the question mark icon):
autoshutdowntooltip

Updated again, this time with a slightly smaller Update button that matches the height of the text box:
idletimeoutsmallerbutton


sources/web/datalab/idleTimeout.ts
```diff
@@ -95,9 +104,13 @@ function parseInterval(str: string) {
   str = str.replace(/\s*/, '');
   if (str) {
     logging.getLogger().debug('garbage left over in interval string: ' + str);
-    return NaN;
+    return { seconds: NaN, errorMessage: 'Invalid format' };
   }
-  return total;
+  if (total > 0 && total < idleTimeoutMinSeconds) {
+    const message = total + ' is smaller than minimum of ' + idleTimeoutMinSeconds;
```
chmeyers: Units?  Nothing in the UI seems to say "seconds", and it is probably non-obvious given that the default is "90m"

sources/web/datalab/static/settings.html
```diff
@@ -4,15 +4,23 @@
...
+    Theme:
     <div class="radio">
       <label><input type="radio" id="lightThemeRadioOption" checked="">Light Theme</label>
     </div>
     <div class="radio">
       <label><input type="radio" id="darkThemeRadioOption" checked="">Dark Theme</label>
     </div>
   </div>
+  <div>
+    Auto-shutdown idle timeout interval (0 to disable):
```
jimmc: Ah, UI design!
Anyone else have an opinion here?

jimmc: yebrahim, how about this: I change the label to "Auto Shutdown" and add an info/question icon just to the right of the label that, when clicked (or hovered) pops up a tooltip with additional text.

yebrahim: That works too I think. We have a lot of room here though, this is why I proposed this design. Up to you. :)

jimmc: I made this change, and added two new screenshots to the PR description.

yebrahim:
One small UI comment, if we can make the input box's height the same as the button it would better. :) Other than this LGTM.

PR #1426, opened 2017-06-13
Add tests for Polymer UI datalab-toolbar

yebrahim:
This starts adding tests for Polymer elements, ideally we want to add these as we write the elements, so I'll try to get tests to catch up with the few we already have.

Polymer conveniently offers a test framework that uses Web Component Tester, which automatically installs selenium webdrivers for all of the running system's browsers, and starts these browsers to run our tests.

The tests are currently not run by any of our test runners, since I'm still getting a sense of how reliable they are, and the plan is to have them run side by side with the other UI and unit tests we have. At some point we also want to add UI tests in the UI test runner, but for now we only run the new unit tests for the Polymre UI.


sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
```diff
@@ -83,15 +83,15 @@
 
     <app-header slot="header" fixed shadow>
       <app-toolbar>
-        <div main-title class="title">
+        <div id="title" main-title class="toolbar-title">
           <!--TODO: @yebrahim replace this with SVG version of the logo-->
-          <img src="images/datalab-logo.png" class="title-image">
+          <img src="/images/datalab-logo.png" class="title-image">
```
ojarjur: Why are these absolute links now? That will cause issues if the base path is not empty.

sources/web/datalab/polymer/index.html
```diff
@@ -20,7 +20,7 @@
           content="width=device-width, minimum-scale=1, initial-scale=1, user-scalable=yes">
 
     <title>Google Cloud Datalab</title>
-    <link rel="shortcut icon" href="images/favicon.ico">
+    <link rel="shortcut icon" href="/images/favicon.ico">
```
ojarjur: Can this remain "images/favicon.ico" or change to "./images/favicon.ico"?

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
yebrahim: The reason for this change is because the tests run from a different path than the actual elements, so if the elements include relative paths, those paths won't be resolved in the test.
Maybe we should have a discussion on the base path change, I want to understand more to see if there's a different way to do it with the new UI.

sources/web/datalab/polymer/index.html
yebrahim: Done. I'll figure out some other way to test the elements with proper paths. Maybe we'll need to put the tests along with elements.

PR #1429, opened 2017-06-15
Add preview images that are built on every push to master

ojarjur:
This change also fixed some hard-coded bucket names so that it is possible to test the Container Builder config by pushing to a test project.

Similarly, this also changed the bucket name from where we pull the production `config_local.js` file so that it *is* hard-coded. The difference is that one is independent of the project ID, whereas the others are not.

### DIFF
--- a/tools/release/cloudbuild.yaml
+++ b/tools/release/cloudbuild.yaml
@@ -107,7 +107,7 @@ steps:
   id:   'cleanupDatalab'
   waitFor: ['buildDatalab']
 
-## And finally, build the GPU Datalab image...
+## Next, build the GPU Datalab image...
 - name: 'debian'
   args: ['/workspace/containers/datalab/prepare.sh', 'datalab-base-gpu', '${REVISION_ID}']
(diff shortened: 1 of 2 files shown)


tools/release/cloudbuild.yaml
```diff
@@ -3,7 +3,7 @@ steps:
 # Write (and publish) a config_local.js file for this build. This
 # can be done in parallel with the rest of the build.
 - name: 'gcr.io/cloud-builders/gsutil'
-  args: ['cp', 'gs://${PROJECT_ID}/deploy/config_local.js', '/workspace/config_local.js']
+  args: ['cp', 'gs://cloud-datalab/deploy/config_local.js', '/workspace/config_local.js']
```
chmeyers: Just to be clear, you are changing this one so it IS hardcoded, and the rest so that they aren't?

ojarjur: Yes, you are right.

I've updated the PR description to reflect that subtlety.

PR #1430, opened 2017-06-15
Add setting to force shimming of websockets.

chmeyers:
This setting can be set with the DATALAB_SETTINGS_OVERRIDES env
variable.

### DIFF
--- a/sources/web/datalab/static/websocket.js
+++ b/sources/web/datalab/static/websocket.js
@@ -1,6 +1,9 @@
 define(['util'], (util) => {
 
   function shouldShimWebsockets() {
+    if (appSettings.shimWebsockets) {
+      return true;
+    }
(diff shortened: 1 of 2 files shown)


sources/web/datalab/static/websocket.js
```diff
@@ -1,6 +1,9 @@
 define(['util'], (util) => {
 
   function shouldShimWebsockets() {
+    if (appSettings.shimWebsockets) {
```
ojarjur: This variable isn't available in the UI.

I would change this to something like `if (document.body.getAttribute('data-proxy-web-sockets') == 'true') {`, and then add `data-proxy-web-sockets` to the HTML template and fill it in based on the `appSettings.shimWebsockets` setting.

PR #1434, opened 2017-06-16
+Notebook, +Folder, Rename, and Delete functionality for Polymer UI

yebrahim:
Initial functionality for the toolbar buttons on the files page. I still didn't hook the enabled/disabled state for these buttons, so they will only work if items are selected.

The UI here is still a proposal, this might change later.

New notebook/folder:
image

Rename:
image

Delete:
image


nikhilk:
Some comments on the UI

- Can you get rid of the Polymer purple? (the bottom border of the textbox)
- Remove the spell checking happening in the textbox?
- In the delete confirmation prompt, what happens if there are way too many items to fit in the bullet list? Does the prompt need to list each file?

sources/web/datalab/polymer/modules/Utils.ts
```diff
@@ -0,0 +1,76 @@
...
+      createModal.inputValue = dialogOptions.inputValue;
+    if (dialogOptions.okTitle)
+      createModal.okTitle = dialogOptions.okTitle;
+    if (dialogOptions.cancelTitle)
+      createModal.cancelTitle = dialogOptions.cancelTitle;
+
+    // Open the dialog
+    return new Promise(resolve => {
+      createModal.openAndWaitAsync((closeResult: DialogCloseResult) => {
+        document.body.removeChild(createModal);
```
jimmc: Since as caller I am not putting the element into the document, it seems like I should not have to take it out. Can this be moved into InputDialogElement?
Or, would it make sense to define a function such as InputDialogElement.openAndReturnPromise that has all of this code, then just make this function "return createModal.openAndReturnPromise()".

yebrahim:
@nikhilk I updated the screenshots, PTAL.

sources/web/datalab/polymer/modules/Utils.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Class provides helper methods for various operations.
+ */
+class Utils {
+
+  /**
+   * Opens a dialog with the specified options. It uses the Datalab custom element
+   * <input-dialog>, attaches a new instance to the current document, opens it
+   * and waits until it closes. It returns a promise that can be waited on for
```
yebrahim: done.

sources/web/datalab/polymer/modules/Utils.ts
```diff
@@ -0,0 +1,76 @@
...
+   * dialog close.
+   * @param dialogOptions options for configuring the dialog
+   */
+  static getUserInputAsync(dialogOptions: DialogOptions) {
+    let createModal = <InputDialogElement>document.createElement('input-dialog');
```
yebrahim: done.

sources/web/datalab/polymer/modules/Utils.ts
```diff
@@ -0,0 +1,76 @@
...
+    document.body.appendChild(createModal);
+
+    createModal.title = dialogOptions.title;
+    if (dialogOptions.bodyHtml)
+      createModal.bodyHtml = dialogOptions.bodyHtml;
+    if (dialogOptions.withInput !== undefined)
```
yebrahim: This is the only boolean one, I want to make sure it's passed before setting it here. I'm not making any assumptions that this should be true if an inputValue is passed.

sources/web/datalab/polymer/modules/Utils.ts
yebrahim: I'm not sure what you mean. The utility function takes care of inserting the element, and it's the one that does the cleanup after. The dialog class shouldn't instantiate/cleanup itself.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -258,7 +262,163 @@ class FilesElement extends Polymer.Element {
...
+          if (type === 'notebook' && !newName.endsWith('.ipynb')) {
+            newName += '.ipynb';
+          }
+          let notebookPlaceholderName = '';
+          return ApiManager.createNewItem(type)
+            .then((notebook: JupyterFile) => {
+              notebookPlaceholderName = notebook.path;
+              return ApiManager.renameItem(notebookPlaceholderName, newName);
+            })
+            .then(this._fetchFileList.bind(this), (e: number) => {
```
jimmc: I think it is a bit easier to read error blocks if you use a .catch statement following the .then rather than passing two functions to .then. Or, in this case, put the catch before this then if you only want it to catch errors from renameItem. In either case, the error block will also get executed if createNewItem resolves with an error, so that might change how you interpret the e arg.

jimmc: Rather than bind, might be better to use an arrow function, as you do on line 351.

sources/web/datalab/polymer/components/input-dialog/input-dialog.ts
```diff
@@ -0,0 +1,171 @@
...
+   */
+  public bodyHtml: string;
+
+  /**
+   * Whether an input field should be included in this dialog
+   */
+  public withInput: boolean;
+
+  /**
+   * If an input field is included, a label that shows up inside it input is given
```
jimmc: I think there is a typo somewhere in this comment.

sources/web/datalab/polymer/components/input-dialog/input-dialog.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * If an input field is included, this will be its initial value
+   */
+  public inputValue: string;
+
+  /**
+   * String for confirm button
+   */
+  public okTitle: string;
```
jimmc: Maybe okLabel and cancelLabel instead.

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -19,7 +19,7 @@ interface ItemListRow {
   firstCol: string,
   secondCol: string,
   icon: string,
-  selected: boolean
+  selected: boolean,
```
jimmc: 👍

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -62,39 +65,103 @@ class ItemListElement extends Polymer.Element {
...
       },
       columns: {
         type: Array,
         value: function(): Array<string> {
           return [];
         }
+      },
+      _selectedElements: {
+        type: Array,
+        value: function(): Array<Object> {
```
jimmc: Can this be written like this?
  value: () => [];

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -62,39 +65,103 @@ class ItemListElement extends Polymer.Element {
...
+   * Returns list of indices for the currently selected elements.
+   */
+  getSelectedIndices() {
+    return this._selectedElements.map(element => {
+      return this.$.list.indexForElement(element);
+    });
+  }
+
+  /**
+   * When the list of rows is refreshed, no items should be selected.
```
jimmc: Perhaps you can figure out a way to reword this comment so it says what this function does ("clears the list of selected items" or some such) as opposed to the desired state after it runs.

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -62,39 +65,103 @@ class ItemListElement extends Polymer.Element {
...
+   */
+  _rowsChanged() {
+    this._selectedElements = [];
+  }
+
+  /**
+   * On row click, checks the click target, if it's the checkbox, adds it to
+   * the selected rows, otherwise selects it only.
+   * This method also maintains the _selectedElements list
```
jimmc: add period

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -62,39 +65,103 @@ class ItemListElement extends Polymer.Element {
...
+    // toggles selection.
+    // Otherwise, select this element, unselect all others.
     if (target.tagName !== 'PAPER-CHECKBOX') {
       for (let i = 0; i < this.rows.length; ++i) {
         this.set('rows.' + i + '.selected', false);
       }
       this.set('rows.' + index + '.selected', true);
+
+      // This is now the only selected element.
+      if (rowElement)
```
jimmc: I had to look at this for a while before figuring it out. rowElement is null or undefined when the clicked element is the checkbox, right? If so, please add a comment somewhere explaining that. If I'm wrong, please add whatever other comment is necessary to explain.

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -149,29 +148,101 @@ class ApiManager {
   }
 
   /**
+   * Create a new notebook or directory.
+   * @param type string type of the created item, can be 'notebook' or 'directory'
+   */
+  static createNewItem(type: string) {
+    return new Promise((resolve, reject) => {
+      ApiManager._xhr(ApiManager.contentApiUrl,
+          (request: XMLHttpRequest) => {
```
jimmc: Now that you are using promises, consider making _xhr return a Promise, then you can use then and catch instead of the main callback and errorCallback.

sources/web/datalab/polymer/modules/Utils.ts
jimmc: Ah, OK, I missed the appendChild call up on line 51.

sources/web/datalab/polymer/modules/ApiManager.ts
yebrahim: good point.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -258,7 +256,154 @@ class FilesElement extends Polymer.Element {
...
+        inputValue: selectedObject.name,
+        okLabel: 'Rename',
+      };
+
+      // Only if the dialog has been confirmed with some user input, rename the
+      // selected item. Then if that is successful, and reload the file list.
+      return Utils.getUserInputAsync(inputOptions)
+        .then((closeResult: DialogCloseResult) => {
+          if (closeResult.confirmed && closeResult.userInput) {
+            let newName = this.currentPath + '/' + closeResult.userInput;
```
jimmc: Should this include the same check for .ipynb extension as in the create method on line 295? If not, then newName can be const.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -258,7 +256,154 @@ class FilesElement extends Polymer.Element {
...
+      // of the selected items, and wait for all promises to finish. Then if that
+      // is successful, reload the file list.
+      return Utils.getUserInputAsync(inputOptions)
+        .then((closeResult: DialogCloseResult) => {
+          if (closeResult.confirmed) {
+            let deletePromises = selectedIndices.map((i: number) => {
+              return ApiManager.deleteItem(this._fileList[i].path);
+            });
+            return Promise.all(deletePromises)
+              .then(() => this._fetchFileList());
```
jimmc: What happens if one of the renames fails? Does an error message get displayed, and does fetchFileList get called?

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -258,7 +256,154 @@ class FilesElement extends Polymer.Element {
...
+          } else {
+            return Promise.resolve(null);
+          }
+        });
+    } else {
+      // TODO: Handle rename errors properly by showing some message to the user
```
jimmc: I don't think this comment is right for this block of code. I think this block only gets executed when this function is called and no items are selected.

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -141,37 +144,108 @@ class ApiManager {
...
+
+    return ApiManager._xhrAsync(path, xhrOptions);
+  }
+
+  /**
+   * Sends an XMLHttpRequest to the specified URL, and parses the
+   * the response text. This method returns immediately with a promise
+   * that resolves with the parsed object when the request completes.
+   */
+  static _xhrAsync(url: string, options?: XhrOptions) {
```
jimmc: Thanks, this updated method and the calling code both seem much cleaner now.

yebrahim: Agreed, good call.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: You're right, same is true for rename block above. Fixed.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: You mean the deletes?
No, no message is displayed for errors yet, I have a few TODO lines in code to add those later.
You're right, if at least one delete gets through and then a failure happens, fetching the file list won't happen. I'll add a comment to think about this when adding error handling.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: Not for rename. The rename dialog already opens with the previous name selected up to the extension, so if the user deletes the extension deliberately, we shouldn't add it back.
I'll change it to const.

PR #1444, opened 2017-06-20
[Polymer UI] Adding a details pane to file picker

yebrahim:
This change adds a toggle-able pane on the right side to the file picker. If exactly one file or notebook is selected in the picker, the pane on the right shows a preview for it. For notebooks, it fetches the first two cells, and renders any markdown in them. For any files with `text/*` mime types, the first 30 rows are fetched and printed. This can be expanded to show other file types as well, such as images.

image
image


sources/web/datalab/polymer/components/datalab-files/datalab-files.html
```diff
@@ -188,6 +195,9 @@
             </div>
           </template>
         </div>
+        <paper-button id="toggleDetails" class="navbar-button" on-click="_toggleDetailsPane">
+          <iron-icon icon="visibility"></iron-icon>
```
jimmc: "visibility" doesn't seem quite right to me here. I admit I don't see a better standard material icon, but maybe I missed something, or maybe we can make a custom "preview" icon that is more appropriate.

As an aside, how the heck did the "rowing" icon get into the material icon set, and why is it in the Action section? Maybe it should have been in the Maps section as "directions row", next to "directions walk" and "directions run".

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -38,12 +38,18 @@ class FilesElement extends Polymer.Element {
    */
   public currentPath: string;
 
+  /**
+   * The currently selected file if exactly one is selected, or null if none is.
```
jimmc: What if multiple files are selected?

sources/web/datalab/polymer/components/details-pane/details-pane.html
```diff
@@ -0,0 +1,87 @@
...
+        margin: 10px 0px;
+        overflow-y: auto;
+        overflow-x: hidden;
+        box-shadow: 5px 5px 10px -3px rgba(0, 0, 0, 0.1);
+      }
+    </style>
+    <div class="placeholder" hidden$="{{file}}">
+      Select an item to view its details.
+    </div>
+    <div id="container" hidden$="{{!file.name}}">
```
jimmc: Does this work without raising an error when file is null?

sources/web/datalab/polymer/components/details-pane/details-pane.html
```diff
@@ -0,0 +1,87 @@
...
+        background-color: #fff;
+        padding: 0px 10px;
+        margin: 10px 0px;
+        overflow-y: auto;
+        overflow-x: hidden;
+        box-shadow: 5px 5px 10px -3px rgba(0, 0, 0, 0.1);
+      }
+    </style>
+    <div class="placeholder" hidden$="{{file}}">
+      Select an item to view its details.
```
jimmc: If multiple items are selected, this message might be confusing.

sources/web/datalab/polymer/components/details-pane/details-pane.html
```diff
@@ -0,0 +1,87 @@
...
+      <div id="details-pane-title">
+        <iron-icon class="content icon" icon="{{_icon}}"></iron-icon>
+        <h1 class="content">{{file.name}}</h1>
+      </div>
+      <hr/>
+      <div class="content">
+        <table>
+          <tr>
+            <td>
+              <strong>Created:</strong>
```
jimmc: Rather than using &lt;strong&gt; here, I think it would be better to do that styling in css. You could either put a class on this &lt;td&gt;, or use &lt;th&gt;.

sources/web/datalab/polymer/components/details-pane/details-pane.html
```diff
@@ -0,0 +1,87 @@
...
+    <div class="placeholder" hidden$="{{file}}">
+      Select an item to view its details.
+    </div>
+    <div id="container" hidden$="{{!file.name}}">
+      <div id="details-pane-title">
+        <iron-icon class="content icon" icon="{{_icon}}"></iron-icon>
+        <h1 class="content">{{file.name}}</h1>
+      </div>
+      <hr/>
+      <div class="content">
```
jimmc: Is this "content" class really the same as the "content" class on line 58? They don't seem like they are necessarily the same.

sources/web/datalab/polymer/components/details-pane/details-pane.ts
```diff
@@ -0,0 +1,147 @@
...
+   * 
+   * TODO: However, consider adding a spinning animation while this data loads.
+   */
+  _reloadDetails() {
+    if (!this.file || !this.active)
+      return;
+
+    this.$.previewHtml.innerHTML = '';
+
+    // If this is a notebook, get the first two cells and render any markdown in them.
```
jimmc: This comment doesn't appear to quite match the condition on the next line.

sources/web/datalab/polymer/components/details-pane/details-pane.html
jimmc: "Select one item to view its details."

yebrahim:
@nikhilk PTAL at the new screenshot, it only has the item preview. I couldn't find a good design just having the contents, they look kind of forced onto the file list, we can discuss this in person. Same for how to indicate this is a partial view of the file. For text files I show a `...` at the end, not sure how to do this in the notebook case.
The eye icon is the toggle for the details pane.

sources/web/datalab/polymer/components/details-pane/details-pane.ts
```diff
@@ -0,0 +1,147 @@
...
+      return;
+
+    if (this.file.type === 'notebook' || this._isPlainTextFile(this.file)) {
+      ApiManager.getJupyterFile(this.file.path)
+        .then((file: JupyterFile) => {
+
+          // If this is a notebook, get the first two cells and render any markdown in them.
+          if (file.type === 'notebook') {
+            const cells = (<JupyterNotebookModel>file.content).cells;
+            const firstTwoCells = cells.slice(0, 2);
```
jimmc: If there are cells you are not showing in the preview, you could add a line at the end saying something like "This notebook includes 3 additional cells that are not shown in this preview", or maybe a shorter version of that.

yebrahim: done.

PR #1445, opened 2017-06-20
Tell the client that it can cache static content

ojarjur:
From my experimenting with it, adding the `Cache-Control` header to responses for static files cuts ~ 1/2 second off of the total time required for each directory page load.

We add that header for every static resource except for `custom.css`, since that one changes every time a user updates their theme setting.

### DIFF
--- a/sources/web/datalab/static.ts
+++ b/sources/web/datalab/static.ts
@@ -100,6 +100,10 @@ function sendFile(filePath: string, response: http.ServerResponse,
       }
     }
     else {
+      if (isDynamic) {
+        response.removeHeader('Cache-Control');
+        response.setHeader('Cache-Control', 'no-cache');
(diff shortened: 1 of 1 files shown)


sources/web/datalab/static.ts
```diff
@@ -228,6 +228,7 @@ function requestHandler(request: http.ServerRequest, response: http.ServerRespon
     'logo.png': 'datalab.png',
   };
 
+  response.setHeader('Cache-Control', 'public, max-age=3600');
```
jimmc: I think perhaps we want not to set this flag if we are running in LIVE mode, or maybe clear it in sendFile if isDynamic is set, around line 91.

ojarjur: Done.

PR #1446, opened 2017-06-21
Reduce page load latency

ojarjur:
This change fixes two server-side performance issues that affected every page load:

1. Loading a user setting that was removed prior to the GA release (this was
   effectively dead code, but still slowed things down).
2. Calling `gcloud auth list` twice; once to determine if the user was signed
   in and a second time to find out the email address of their account. Even
   calling `gcloud auth list` once on every page load is overkill, but at least
   this cuts the number of calls in half. A TODO has been added to get this down
   even further.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -366,9 +366,13 @@ export function handleRequest(request: http.ServerRequest, response: http.Server
 }
 
 function getBaseTemplateData(request: http.ServerRequest): common.Map<string> {
-  var userId: string = userManager.getUserId(request);
-  var reportingEnabled: string = process.env.ENABLE_USAGE_REPORTING;
-  var templateData: common.Map<string> = {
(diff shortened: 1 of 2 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -384,10 +378,13 @@ function getBaseTemplateData(request: http.ServerRequest): common.Map<string> {
     reportingEnabled: reportingEnabled,
     proxyWebSockets: appSettings.proxyWebSockets,
   };
-  var signedIn = auth.isSignedIn();
+  // TODO: Cache the gcloudAccount value so that we are not
+  // calling `gcloud` on every page load.
+  var gcloudAccount : string = auth.getGcloudAccount();
```
jimmc: We should be using const and let rather than var in new and updated code.

ojarjur: Done

PR #1447, opened 2017-06-21
[Polymer UI] Implement input dialog as a subclass of base dialog

yebrahim:
As we're adding more types of dialogs, using inheritance is a cleaner and easier approach.

In order to use inheritance with Polymer, this change adds a template stamp mechanism, which takes the subclass and baseclass, and stamps the former into an element inside the latter. This way we can reuse the Javascript across the elements, and we can extend the template at the same time.

This change adds a BaseDialog type that shows a dialog with room for a title, messag, and ok/cancel buttons. The InputDialog then extends this to add an input element with some more input behavior.

### DIFF
--- a/sources/web/datalab/polymer/modules/Utils.ts
+++ b/sources/web/datalab/polymer/modules/Utils.ts
@@ -15,23 +15,19 @@
 /**
  * Options for opening a dialog.
  */
+enum DialogType {
+  input,
+  confirm,
+}
 interface DialogOptions {
   title: string,
+  messageHtml?: string,
   bodyHtml?: string,
-  withInput?: boolean,
   inputLabel?: string,
   inputValue?: string,
   okLabel?: string,
   cancelLabel?: string,
-}
-
-/**
- * Dialog close context, includes whether the dialog was confirmed, and any
- * user input given.
- */
-interface DialogCloseResult {
-  confirmed: boolean
-  userInput: string,
+  big?: boolean,
 }
 
 /**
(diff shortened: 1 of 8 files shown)


sources/web/datalab/polymer/modules/Utils.ts
```diff
@@ -41,32 +37,40 @@ class Utils {
...
+   * document, opens it, and returns a promise that resolves when the dialog is closed.
+   * @param type specifies which type of dialog to use
+   * @param dialogOptions specifies different options for opening the dialog
    */
-  static getUserInputAsync(dialogOptions: DialogOptions) {
-    const createModal = <InputDialogElement>document.createElement('input-dialog');
-    document.body.appendChild(createModal);
+  static showDialog(type: DialogType, dialogOptions: DialogOptions) {
+    let dialogElement = '';
+    if (type === DialogType.input) {
```
jimmc: Seems like you might be able to use a "rich enum" here, basically have the dialog types be subclasses of a DialogType class, then you could just grab the dialogElement from a field of that class. But this is OK as is for just two dialog types.

yebrahim: I'm not sure I see how this is better, I think a simple enum is all we need here.
I'd love to see your take on this, let's discuss this face to face.

yebrahim:
Looks like a merge issue removed my `stampInBaseTemplate` method, which is a critical part of this PR, this is ready for another review pass.

sources/web/datalab/polymer/modules/Utils.ts
jimmc: When using an OO system, case statements and long if/else chains are usually a sign to look for an OO solution that allows new choices or types to be added just by adding another subclass. In this case, since your if/else chain is only two items long, and I think this may be the only place where you are checking the value of the DialogType enum, it's not a problem.

sources/web/datalab/polymer/modules/Utils.ts
```diff
@@ -41,35 +37,77 @@ class Utils {
...
+   * @param baseRootElementSelector a selector for an element that will be root
+   *                                for the stamped template
+   */
+  static stampInBaseTemplate(subType: string, baseType: string,
+                             baseRootElementSelector: string) {
+    // Start with the base class's template
+    const basetypeTemplate = Polymer.DomModule.import(baseType, 'template');
+    const subtypeTemplate = Polymer.DomModule.import(subType, 'template');
+    // Clone the base template; we don't want to change it in-place
+    let stampedTemplate = <PolymerTemplate>basetypeTemplate.cloneNode(true);
```
jimmc: const

sources/web/datalab/polymer/modules/Utils.ts
yebrahim: I see. Makes sense.

PR #1448, opened 2017-06-21
[Polymer UI] +File, Copy and Move functionality

yebrahim:
This change adds the ability to copy and move around files and directories. It shows a flyout directory picker for the destination, then copies/moves the selected item.

In order to achieve this, I added a `small` attribute to the `datalab-files` element, which hides its chrome and keeps only the navigation buttons, see details in the comments.

image

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
+++ b/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
@@ -296,7 +296,7 @@ class FilesElement extends Polymer.Element {
       okLabel: 'Create',
     };
 
-    return Utils.showDialog(DialogType.input, inputOptions)
+    return Utils.showDialog(InputDialogElement, inputOptions)
       .then((closeResult: InputDialogCloseResult) => {
         // Only if the dialog has been confirmed with some user input, rename the
         // newly created file. Then if that is successful, reload the file list
@@ -336,7 +336,7 @@ class FilesElement extends Polymer.Element {
 
       // Only if the dialog has been confirmed with some user input, rename the
       // selected item. Then if that is successful, and reload the file list.
-      return Utils.showDialog(DialogType.input, inputOptions)
+      return Utils.showDialog(InputDialogElement, inputOptions)
         .then((closeResult: InputDialogCloseResult) => {
           if (closeResult.confirmed && closeResult.userInput) {
             const newName = this.currentPath + '/' + closeResult.userInput;
@@ -397,7 +397,7 @@ class FilesElement extends Polymer.Element {
       // Only if the dialog has been confirmed, call the ApiManager to delete each
       // of the selected items, and wait for all promises to finish. Then if that
       // is successful, reload the file list.
-      return Utils.showDialog(DialogType.confirm, inputOptions)
+      return Utils.showDialog(BaseDialogElement, inputOptions)
         .then((closeResult: BaseDialogCloseResult) => {
           if (closeResult.confirmed) {
             let deletePromises = selectedIndices.map((i: number) => {
@@ -429,7 +429,7 @@ class FilesElement extends Polymer.Element {
         okLabel: 'Copy Here',
         big: true,
       };
-      Utils.showDialog(DialogType.dirPicker, options)
+      Utils.showDialog(DirectoryPickerDialogElement, options)
         .then((closeResult: DirectoryPickerDialogCloseResult) => {
           if (closeResult.confirmed) {
             const newPath = closeResult.directoryPath;
@@ -454,7 +454,7 @@ class FilesElement extends Polymer.Element {
         okLabel: 'Move Here',
         big: true,
       };
-      Utils.showDialog(DialogType.dirPicker, options)
+      Utils.showDialog(DirectoryPickerDialogElement, options)
         .then((closeResult: DirectoryPickerDialogCloseResult) => {
           if (closeResult.confirmed) {
             const newPath = closeResult.directoryPath;
--- a/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
+++ b/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
@@ -287,6 +287,9 @@ class FilesElement extends Polymer.Element {
   _createNewNotebook() {
     return this._createNewItem('notebook');
   }
+  _createNewFile() {
+    return this._createNewItem('file');
+  }
 
   _createNewDirectory() {
     return this._createNewItem('directory');
--- a/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
+++ b/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
@@ -125,8 +125,8 @@ class FilesElement extends Polymer.Element {
     // Refresh the file list periodically.
     // TODO: [yebrahim] Start periodic refresh when the window is in focus, and
     // the files page is open, and stop it on blur to minimize unnecessary traffic
-    this._fileListRefreshIntervalHandle = setInterval(this._fetchFileList.bind(this),
-                                                      this._fileListRefreshInterval);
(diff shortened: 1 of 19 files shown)


sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -404,6 +435,70 @@ class FilesElement extends Polymer.Element {
...
+        okLabel: 'Move Here',
+        big: true,
+      };
+      Utils.showDialog(DirectoryPickerDialogElement, options)
+        .then((closeResult: DirectoryPickerDialogCloseResult) => {
+          if (closeResult.confirmed) {
+            const newPath = closeResult.directoryPath;
+            // Moving is renaming.
+            return ApiManager.renameItem(selectedObject.path, newPath + '/' + selectedObject.name)
+              .then(() => this._fetchFileList());
```
jimmc: What happens if the rename fails? Should this "then" be a "finally" instead so that it always updates the list?

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -404,6 +435,70 @@ class FilesElement extends Polymer.Element {
...
+    if (selectedIndices.length === 1) {
+      const i = selectedIndices[0];
+      const selectedObject = this._fileList[i];
+
+      const options: DialogOptions = {
+        title: 'Copy Item',
+        okLabel: 'Copy Here',
+        big: true,
+      };
+      Utils.showDialog(DirectoryPickerDialogElement, options)
```
jimmc: Should this function and the rename function return a Promise to make testing easier?

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -259,6 +287,9 @@ class FilesElement extends Polymer.Element {
   _createNewNotebook() {
     return this._createNewItem('notebook');
   }
+  _createNewFile() {
```
jimmc: Please add a blank line before this function.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -104,12 +119,20 @@ class FilesElement extends Polymer.Element {
...
 
-    this.$.files.columns = ['Name', 'Status'];
+    // For a small file/directory picker, we don't need to show the status.
+    this.$.files.columns = this.small ? ['Name'] : ['Name', 'Status'];
 
     // Refresh the file list periodically.
     // TODO: [yebrahim] Start periodic refresh when the window is in focus, and
     // the files page is open, and stop it on blur to minimize unnecessary traffic
-    setInterval(this._fetchFileList.bind(this), this._fileListRefreshInterval);
+    this._fileListRefreshIntervalHandle = setInterval(this._fetchFileList.bind(this),
```
jimmc: Prefer breaking lines at a higher semantic level; in this case, just after the "=".

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: No I think we want this to fail if the rename fails, and the file list refresh isn't going to do anything in that case, the list hasn't changed.
I'm going to add error handling UX later for all these operations.

sources/web/datalab/polymer/components/item-list/item-list.html
```diff
@@ -29,8 +29,13 @@
       }
       #list-container {
         overflow: auto;
+      }
+      .list-container-no-header--false {
```
yebrahim: Well this follows the name of the added attribute: `no-header` that you can add to the `item-list` element to disable the header. I think this makes more sense that defining a `with-header` attribute, which should be the default behavior.

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -54,6 +57,16 @@ class ItemListElement extends Polymer.Element {
    */
   public columns: Array<string>;
 
+  /**
+   * Whether to hide the header row
+   */
+  public noHeader: boolean;
```
yebrahim: This feels like it should be the default behavior though. If you're adding an `item-list` element, you should explicitly tell it if you want to disable that feature, rather than add it.

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -54,6 +57,16 @@ class ItemListElement extends Polymer.Element {
...
+
+  /**
+   * Whether to disable item selection
+   */
+  public noSelection: boolean;
```
yebrahim: Same discussion as above, I'll rename to `disable-selection` for now, and we can discuss this in person if you have strong feelings.

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -216,6 +216,25 @@ class ApiManager {
...
   /**
+   * Copy an item from source to destination. Item name collisions at the destination
+   * are handled by Jupyter.
+   * @param fromDir source path to copy from
+   * @param toPath destination path to copy to
+   */
+  static copyItem(fromPath: string, toPath: string) {
+    toPath = ApiManager.contentApiUrl + '/' + toPath;
+    const xhrOptions: XhrOptions = {
+      method: 'POST',
```
yebrahim: The Jupyter backend expects `POST` to be used for copy. See [here](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/handlers.py#L182) and [here](https://github.com/jupyter/notebook/blob/master/notebook/services/contents/handlers.py#L224).

sources/web/datalab/polymer/components/item-list/item-list.ts
jimmc: It's easy to set the default value to true, right? Down on line 87, where the default is currently explicitly set to false.

sources/web/datalab/polymer/modules/ApiManager.ts
jimmc: I had assumed the toPath argument was the full path to the destination file. If it is actually the containing directory, as it appears to be in your first link, then please change the variable name or param description to make that clear.

sources/web/datalab/polymer/components/item-list/item-list.ts
yebrahim: Yes, but the usage will then have to be `<item-list show-header="false" enable-selection="false>` rather than simply `<item-list hide-header disable-selection>`, which is a much smoother design.
I think using the negative name is fine since it's not the main experience.

sources/web/datalab/polymer/modules/ApiManager.ts
yebrahim: done.

sources/web/datalab/polymer/components/item-list/item-list.ts
jimmc: Yes, hide-header would be better than no-header.

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -216,6 +216,25 @@ class ApiManager {
   }
 
   /**
+   * Copies an item from source to destination. Item name collisions at the destination
+   * are handled by Jupyter.
+   * @param fromDir source path to copy from
```
jimmc: Please update comments to match your new parameter names.

sources/web/datalab/polymer/components/item-list/item-list.ts
yebrahim: sure, done.

sources/web/datalab/polymer/modules/ApiManager.ts
yebrahim: fixed.

PR #1452, opened 2017-06-22
[Polymer UI] Save last user path to startuppath

yebrahim:
This PR makes the new UI remember the user's last visited path, like the existing UI does. It does this in a backward-compatible way, so both UIs use the same server setting.

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
+++ b/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
@@ -93,6 +93,23 @@ class FilesElement extends Polymer.Element {
    * to initialize element state.
    */
   ready() {
+    // Get the last startup path. Must be called before super.ready() to beat the
+    // currentPath changed event.
+    ApiManager.getUserSettings()
+      .then((settings: UserSettings) => {
+        if (settings.startuppath) {
+          let path = settings.startuppath;
+          // For backward compatibility with the current path format.
+          if (path.startsWith('/tree/')) {
+            path = path.substr('/tree/'.length);
+          }
+          if (path.startsWith(this.basePath)) {
+            path = path.substr(this.basePath.length);
+          }
+          this.currentPath = path;
+        }
+      });
+
     super.ready();
 
     // TODO: [yebrahim] The current path should be fetched from the server
--- a/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
+++ b/sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
@@ -115,10 +115,7 @@ class FilesElement extends Polymer.Element {
           this.currentPath = path;
         }
       })
-      .catch(() => {
-        this._fetching = false;
-        console.log('Failed to get the user settings.');
-      })
+      .catch(() => console.log('Failed to get the user settings.'))
       .then(() => {
         this._fetching = false;
         // Refresh the file list periodically.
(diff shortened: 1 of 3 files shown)


sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -93,6 +93,23 @@ class FilesElement extends Polymer.Element {
    * to initialize element state.
    */
   ready() {
+    // Get the last startup path. Must be called before super.ready() to beat the
+    // currentPath changed event.
+    ApiManager.getUserSettings()
```
jimmc: I think it would be useful to define ApiManager.getUserSetting(name:string, defaultValue:any) where defaultValue is optional.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -93,6 +93,23 @@ class FilesElement extends Polymer.Element {
...
+      .then((settings: UserSettings) => {
+        if (settings.startuppath) {
+          let path = settings.startuppath;
+          // For backward compatibility with the current path format.
+          if (path.startsWith('/tree/')) {
+            path = path.substr('/tree/'.length);
+          }
+          if (path.startsWith(this.basePath)) {
```
jimmc: Shouldn't the check for basePath go before the check for /tree/? If I understand basePath correctly, you would get something like /base-path/tree/ in the path coming in. Or is /tree/ no longer used when there is a basePath?

sources/web/datalab/server.ts
```diff
@@ -164,12 +164,13 @@ function handleRequest(request: http.ServerRequest,
       (path.indexOf('/terminals') == 0) ||
       (path.indexOf('/sessions') == 0)) {
 
-    if (path.indexOf('/tree') == 0) {
-      const filePath = '/content' + path.substr('/tree'.length);
+    if (path.indexOf('/api/contents') == 0) {
+      const subPath = path.substr('/api/contents'.length);
+      const filePath = path_.join('/content', subPath);
```
jimmc: Using both path and path_ seems like asking for trouble. How about renaming path to requestPath?

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -93,10 +93,41 @@ class FilesElement extends Polymer.Element {
...
+          }
+          // For backward compatibility with the current path format.
+          if (path.startsWith('/tree/')) {
+            path = path.substr('/tree/'.length);
+          }
+          this.currentPath = path;
+        }
+      })
+      .catch(() => {
+        this._fetching = false;
```
jimmc: You don't need to set _fetching to false here, because you are doing that in the following then block, which will always be executed after this catch block. If your intent is not to execute the following then block, you should either throw from this catch block, or move this catch block to the end of the chain.

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -216,6 +227,13 @@ class ApiManager {
   }
 
   /**
+   * Gets the user settings JSON from the server.
+   */
+  static getUserSettings() {
+    return ApiManager._xhrAsync('/_settings');
```
jimmc: The point of cacheing here would be to avoid having multiple places that cache. I don't care if it is outside ApiManager, but I think it should be done in one place, and I should be able to call that one place from anywhere.

yebrahim: What I'm trying to say is, it depends on the consumption of the settings. For datalab-files, it always needs a fresh copy of the settings, so it doesn't make sense for it to use a cached version if one exists.
When the need rises for caching settings for reuse, we can add that.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: You're right, fixed.

PR #1457, opened 2017-06-23
Use separate settings file for basepath

yebrahim:
This change separates out the datalab base path into its own JSON file. This makes the deployment to a different backend that requires a different setting value easier by swapping out this file.
It also adds a `/api/basepath` handler that returns this basepath setting.


sources/web/datalab/server.ts
```diff
@@ -153,6 +153,12 @@ function handleRequest(request: http.ServerRequest,
     return;
   }
 
+  if (requestPath.indexOf('/api/basepath') === 0) {
```
chmeyers: Just to make sure I'm reading this correctly, nothing is yet actually making a request for /api/basepath, right?

sources/web/datalab/server.ts
```diff
@@ -153,6 +153,12 @@ function handleRequest(request: http.ServerRequest,
...
+    response.statusCode = 200;
+    response.end(appSettings.datalabBasePath);
```
chmeyers: Since I'm not sure, what exact format is this returning?  is it pure text or is it a JSON dict?

sources/web/datalab/server.ts
yebrahim: Correct.

sources/web/datalab/server.ts
yebrahim: It's a string.

PR #1460, opened 2017-06-26
Make old ui backward compatible with the new startuppath format

yebrahim:
I recently changed the `startuppath` format to not contain the `/tree` prefix, but the old UI still expects the prefix. This broke clients that already have the setting saved in the old format, and were not using the experimental UI. This change makes the old UI work with or without it so it can talk to either format.

This change also lets the `sources/build.sh` output to stdout directly, and fixes a few things that broke with the new `tsc`.

### DIFF
--- a/sources/web/datalab/server.ts
+++ b/sources/web/datalab/server.ts
@@ -127,7 +127,7 @@ function handleRequest(request: http.ServerRequest,
     if (startup_path_setting in loadedSettings) {
       let startuppath = loadedSettings[startup_path_setting];
 
-      // For backward compatibility with the old path format, trim /tree prefix.
+      // For backward compatibility with the old path format, prepend /tree prefix.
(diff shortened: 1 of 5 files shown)


sources/web/datalab/server.ts
```diff
@@ -125,7 +125,16 @@ function handleRequest(request: http.ServerRequest,
     response.statusCode = 302;
     var redirectUrl : string;
     if (startup_path_setting in loadedSettings) {
-      redirectUrl = '/tree' + loadedSettings[startup_path_setting];
+      let startuppath = loadedSettings[startup_path_setting];
+
+      // For backward compatibility with the old path format, trim /tree prefix.
```
ojarjur: s/trim/add/

yebrahim: fixed.

PR #1463, opened 2017-06-27
[Polymer UI] Add terminal functionality

yebrahim:
Terminal experience for the new UI. This uses `xterm.js` for the frontend element, and connects to one Jupyter backend terminal.

image

### DIFF
--- a/containers/datalab/content/.bashrc
+++ b/containers/datalab/content/.bashrc
@@ -1,8 +1,8 @@
 # ~/.bashrc: executed by bash(1) for non-login shells.
 
 cd /content
-STARTCOLOR='\e[1;32m';
-ENDCOLOR="\e[0m"
+STARTCOLOR='\[\e[1;32m\]';
+ENDCOLOR="\[\e[0m\]"
 export PS1="$STARTCOLOR\u# \w> $ENDCOLOR"
 echo 'Welcome to Google Cloud Datalab.
 Please note that only changes inside /content will be preserved if
--- a/containers/datalab/content/.bashrc
+++ b/containers/datalab/content/.bashrc
@@ -1,9 +1,9 @@
 # ~/.bashrc: executed by bash(1) for non-login shells.
 
 cd /content
-STARTCOLOR='\[\e[1;32m\]';
-ENDCOLOR="\[\e[0m\]"
-export PS1="$STARTCOLOR\u# \w> $ENDCOLOR"
+STARTCOLOR='\[\e[1;32m\]'
+ENDCOLOR='\[\e[0m\]'
+export PS1="$STARTCOLOR\w> $ENDCOLOR"
 echo 'Welcome to Google Cloud Datalab.
 Please note that only changes inside /content will be preserved if
 the Datalab VM is shutdown or restarted. Package installations
--- a/sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
+++ b/sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
@@ -91,8 +91,16 @@ the License.
         <a href="/sessions" id="sessions">
           <paper-button class="sidebar-button">
             <div class="sidebar-button-container">
-              <iron-icon class="sidebar-button-icon" icon="av:featured-play-list"></iron-icon>
+              <iron-icon class="sidebar-button-icon" icon="list"></iron-icon>
               <div class="sidebar-button-title">Sessions</div>
+            </div>
+          </paper-button>
+        </a>
+        <a href="/terminal" id="terminal">
+          <paper-button class="sidebar-button">
+            <div class="sidebar-button-container">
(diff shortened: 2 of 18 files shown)


sources/web/datalab/polymer/components/datalab-terminal/datalab-terminal.ts
```diff
@@ -0,0 +1,140 @@
...
+   * kept in sync, otherwise line wrapping issues will happen.
+   */
+  _resizeHandler() {
+    if (this._xterm) {
+      const rows = this.$.theTerminal.clientHeight / this._charHeight;
+      const cols = this.$.theTerminal.clientWidth / this._charWidth;
+
+      this._xterm.resize(Math.floor(cols), Math.floor(rows));
+
+      if (this._wsConnection.readyState === 1) {
```
jimmc: Under what conditions would a resize happen and the connection not be ready? If that happens, would the Jupyter terminal size then not match the front-end terminal size?

containers/datalab/content/.bashrc
```diff
@@ -1,8 +1,8 @@
 # ~/.bashrc: executed by bash(1) for non-login shells.
 
 cd /content
-STARTCOLOR='\e[1;32m';
-ENDCOLOR="\e[0m"
+STARTCOLOR='\[\e[1;32m\]';
+ENDCOLOR="\[\e[0m\]"
 export PS1="$STARTCOLOR\u# \w> $ENDCOLOR"
```
yebrahim: Just cosmetic. I think I'll take it out along with the username, since our instances are meant to support a single user anyway, and just keep it simple with the CWD only.

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
```diff
@@ -91,11 +91,19 @@
         <a href="/sessions" id="sessions">
           <paper-button class="sidebar-button">
             <div class="sidebar-button-container">
-              <iron-icon class="sidebar-button-icon" icon="av:featured-play-list"></iron-icon>
+              <iron-icon class="sidebar-button-icon" icon="list"></iron-icon>
               <div class="sidebar-button-title">Sessions</div>
             </div>
           </paper-button>
         </a>
+        <a href="/terminal" id="terminal">
```
yebrahim: The new UI doesn't support basepath yet, I'll still do a pass over all these to fix them when I get the basepath setting from the server. That'll be a separate change.

sources/web/datalab/polymer/components/datalab-terminal/datalab-terminal.ts
```diff
@@ -0,0 +1,140 @@
...
+  clear(): void;
+  reset(): void;
+  resize(cols: number, rows: number): void;
+  on(event: string, handler: Function): void;
+}
+
+/**
+ * Terminal element for Datalab.
+ * This element uses xterm.js for the front end terminal element, and opens a websocket
+ * connection to a Jupyter terminal on the backend. Although Jupyter supports multiple
```
yebrahim: Well the user can still open the dev tools and send an XHR request to create a new terminal, so that case *could* theoretically happen. We don't have to prevent it, but we won't support it.

sources/web/datalab/polymer/components/datalab-terminal/datalab-terminal.ts
```diff
@@ -0,0 +1,140 @@
...
+  private _charWidth: number;
+
+  static get is() { return "datalab-terminal"; }
+
+  ready() {
+    super.ready();
+    const self = this;
+
+    // Will be called after the custom element is done rendering.
+    window.addEventListener('WebComponentsReady', function() {
```
yebrahim: done.

sources/web/datalab/polymer/components/datalab-terminal/datalab-terminal.ts
```diff
@@ -0,0 +1,140 @@
...
+      // Use the size helper element to get the height and width of a character. This
+      // makes changing the style simpler, instead of hard-coding these values.
+      self._charHeight = self.$.sizeHelper.clientHeight;
+      self._charWidth = self.$.sizeHelper.clientWidth / 10; // The element has 10 characters.
+
+      self._boundResizeHandler = self._resizeHandler.bind(self);
```
yebrahim: This is here so I can call `removeEventListener`. The binding, regardless whether it's done explicitly or implicitly using an arrow function, generates a new function reference, which is what needs to be passed in the remove call.

sources/web/datalab/polymer/components/datalab-terminal/datalab-terminal.ts
yebrahim: This is here from an earlier form of the change, where the handler was called right at the start before opening the connection. No longer needed, and removed.

PR #1464, opened 2017-06-27
Expose app settings as an API

yebrahim:
This PR exposes a new `/api/settings` to return the app settings JSON. We do this to make it possible to query those settings without modifying the static front-end files before serving them, so as to avoid using document data attribute.

It also adds strong typing for user settings, to match how app settings are typed, instead of using a string map.


sources/web/datalab/common.d.ts
```diff
@@ -14,6 +14,14 @@
 
 declare module common {
 
+  interface UserSettings {
```
jimmc: One of the things I have found a little confusing is the use of the term "settings" to refer to both application configuration settings and user preferences. Do you suppose we could start calling those "app config" and "user preferences"? I feel like those terms are the more commonly used terms for these two types of information.

sources/web/datalab/settings.ts
```diff
@@ -28,6 +28,7 @@ import idleTimeout = require('./idleTimeout');
 import userManager = require('./userManager');
 
 var SETTINGS_FILE = 'settings.json';
```
jimmc: Does this file contain user settings, app settings, or both?

sources/web/datalab/settings.ts
```diff
@@ -186,15 +205,30 @@ export function updateUserSetting(userId: string, key: string, value: string, as
  * @param response the outgoing http response.
  */
 function requestHandler(request: http.ServerRequest, response: http.ServerResponse): void {
-  var userId = userManager.getUserId(request);
-  if ('POST' == request.method) {
-    postSettingsHandler(userId, request, response);
+  if (request.url.indexOf('/api/settings') === 0) {
```
jimmc: Is this call supposed to return app settings, user settings, or both?

sources/web/datalab/common.d.ts
yebrahim: It used to be a little confusing before your rename, I think it's clear now.
Also from the user's perspective, there's only one set of settings they know about, which is user settings, unless they dig really deep to find what we call app settings.

sources/web/datalab/settings.ts
yebrahim: Both. There are functions to load and handle each.

sources/web/datalab/settings.ts
yebrahim: Both as of this update. It handles calls to `/api/settings` (app settings) as well as `/_settings` (user settings).

PR #1466, opened 2017-06-28
Allow alternate access to experimental UI via /exp/ path.

jimmc:
This change allows me to more easily compare details of the old and new UIs. I can run without the experiment flag to view and use the old UI, then open another tab at localhost:80801/exp/ to view and use the new UI at the same time.


jimmc:
I can directly open /sessions in one tab and /exp/sessions in another tab, and reload each as needed. So this PR allows me to view both old and new in those cases where there is a conflict. And I don't need to worry about knowing all the places where there might be a conflict.

yebrahim:
Ah I see, this is useful then.

PR #1467, opened 2017-06-28
Avoid failing if empty settings file found

yebrahim:
If a user settings or app settings file is found but empty, the `JSON.parse` will fail, since empty string isn't valid JSON.

### DIFF
--- a/sources/web/datalab/settings.ts
+++ b/sources/web/datalab/settings.ts
@@ -62,7 +62,7 @@ export function loadAppSettings(): common.AppSettings {
       metadata = <Metadata>JSON.parse(fs.readFileSync(metadataPath, 'utf8'));
     }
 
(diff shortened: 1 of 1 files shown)


sources/web/datalab/settings.ts
```diff
@@ -62,7 +62,7 @@ export function loadAppSettings(): common.AppSettings {
       metadata = <Metadata>JSON.parse(fs.readFileSync(metadataPath, 'utf8'));
     }
 
-    var settings = <common.AppSettings>JSON.parse(fs.readFileSync(settingsPath, 'utf8'));
+    var settings = <common.AppSettings>JSON.parse(fs.readFileSync(settingsPath, 'utf8') || '{}');
```
jimmc: Since your're updating, please use let instead of var.

PR #1471, opened 2017-06-28
[Polymer UI] Add support for dark theme

yebrahim:
Most of this change is styling to use CSS3 variables instead of hard coded colors, then there's a `index.light.css` and a `index.dark.css` that specify values for these colors, then there's a little bit of logic on the server side to return the right file.

The theme can currently be switched manually by making a `POST` request to the settings API. Next I'll add the UI setting.

image


sources/web/datalab/polymer/index.dark.css
```diff
@@ -12,16 +12,13 @@
...
-  right: 0;
-  min-width: 470px;
+:root {
+  --primary-accent-color: #222;
+  --primary-bg-color: #222;
+  --secondary-bg-color: #2b2b2b;
+  --primary-fg-color: #aaa;
+  --selection-bg-color: #444;
+  --selection-fg-color: #d68600;
+  --disabled-bg-color: var(--primary-bg-color);
```
jimmc: Why is there a disabled background color if it is always set to the same as the primary background color? If the expectation is that it is used with a different background color, such as secondary-bg-color, perhaps the name should be disabled-secondary-bg-color.

sources/web/datalab/polymer/index.dark.css
```diff
@@ -12,16 +12,13 @@
...
-  top: 0;
-  bottom: 0;
-  left: 0;
-  right: 0;
-  min-width: 470px;
+:root {
+  --primary-accent-color: #222;
+  --primary-bg-color: #222;
+  --secondary-bg-color: #2b2b2b;
+  --primary-fg-color: #aaa;
```
jimmc: The order seems a bit random here. Maybe you could put all the primary colors together, or put all the backgrounds together, or use alpha order.

sources/web/datalab/polymer/index.dark.css
yebrahim: It's just that it's a different state, so I think it's good to separate its variable out, even if it has the same value.
I'm ok removing this if you still don't think it should be there.

sources/web/datalab/polymer/index.dark.css
yebrahim: Sure, alpha seems fine.

sources/web/datalab/polymer/index.dark.css
jimmc: Not high priority for me, so it's OK as is.

PR #1472, opened 2017-06-28
[Polymer UI] Add shutdown sessions functionality

yebrahim:
This allows (multiple) selected sessions to be shutdown in the sessions page.

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
+++ b/sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
@@ -113,6 +113,26 @@ class SessionsElement extends Polymer.Element {
       });
   }
 
+  /**
+   * Calls the ApiManager to terminate the selected sessions.
+   */
+  _shutdownSelectedSessions() {
+    const selectedIndices = this.$.sessions.getSelectedIndices();
+    if (selectedIndices.length) {
+      let shutdownPromises = selectedIndices.map((i: number) => {
+        return ApiManager.shutdownSessionAsync(this._sessionList[i].id);
+      });
+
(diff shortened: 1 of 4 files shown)


sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
```diff
@@ -113,6 +113,26 @@ class SessionsElement extends Polymer.Element {
       });
   }
 
+  /**
+   * Calls the ApiManager to terminate the selected sessions.
+   */
+  _shutdownSelectedSessions() {
+    const selectedIndices = this.$.sessions.getSelectedIndices();
+    if (selectedIndices.length) {
+      let shutdownPromises = selectedIndices.map((i: number) => {
```
jimmc: const

sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
```diff
@@ -113,6 +113,26 @@ class SessionsElement extends Polymer.Element {
...
+        return ApiManager.shutdownSessionAsync(this._sessionList[i].id);
+      });
+
+      // TODO: [yebrahim] If at least one delete completes then a failure happens with
+      // any of the rest, _fetchFileList will never be called.
```
jimmc: Do you mean _fetchSessionList will never be called?
The 'then' is only called after all promises succeed; if any of them fail, the all() fails immediately. "At least one ... completes" is not part of the calculation.

yebrahim: You're right, I confused it with the reverse case. Fixed.

PR #1473, opened 2017-06-28
Fix color of toolbar icons in experimental UI.

jimmc:
The recent commit 0698e8 caused the icon buttons in the polymer UI to become invisible. This fixes that.


jimmc:
It looks like all the other colors are currently defined in the two theme css files, so it makes more sense to me to leave it there for now.

yebrahim:
SGTM.

PR #1474, opened 2017-06-29
Add polymer UI elements for sign-in and sign-out.

jimmc:
This PR adds sign-in and sign-out buttons to the account menu in the Polymer UI. At the moment, they do nothing except toggle the signed-in state, they don't actually sign in.

When signed out, the sign-in button is visible:
accountmenusignin

When signed in, some info about user and project appears along with the sign-out button:
accountmenusignout

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
+++ b/sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
@@ -24,6 +24,9 @@ the License.
 <dom-module id="datalab-toolbar">
   <template>
     <style include="datalab-shared-styles">
+      [hidden] {
+        display: none !important;
+      }
       :host {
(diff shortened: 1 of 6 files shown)


sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
```diff
@@ -24,6 +24,9 @@
 <dom-module id="datalab-toolbar">
   <template>
     <style include="datalab-shared-styles">
+      [hidden] {
```
yebrahim: I think we should move this (and its other occurrences) into the `shared-styles` element.

jimmc: Done.

jimmc:
Created auth-panel component.

PR #1475, opened 2017-06-30
Adding basepath support in the experimental UI ApiManager.

chmeyers:
This includes doing an XSRF token dance if the server supports it.


sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -289,14 +309,68 @@ class ApiManager {
...
+          } else {
+            ApiManager.xsrfToken = j['token'];
+            var formData = new FormData();
+            formData.append("token", ApiManager.xsrfToken);
+            const xhrOptions: XhrOptions = {
+              noCache: true,
+              method: 'POST',
+              parameters: formData,
+            };
+            return ApiManager._xhrAsyncRaw(ApiManager.basepathApiUrl, xhrOptions).then((response:string) => {
```
yebrahim: Convention is to put the `.then` part on the next line, indented by two spaces, makes it a lot more readable.

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -124,13 +124,33 @@ class ApiManager {
   static readonly terminalApiUrl = '/api/terminals';
```
yebrahim: We should remove all `/` prefix from the api urls here, except for `/api/basepath` right?

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -180,12 +180,12 @@ class FilesElement extends Polymer.Element {
   }
 
   _getNotebookUrlPrefix() {
-    let prefix = location.protocol + '//' + location.host + '/';
+    let prefix = location.protocol + '//' + location.host + ApiManager.basepath + '/';
```
chmeyers: Done.  As discussed offline: only datalab-files needs the basepath because the notebooks/files are stored on the VM.

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -289,14 +309,68 @@ class ApiManager {
...
+    }
+  }
+
+  /**
+   * Sends an XMLHttpRequest to the specified URL, adding the required
+   * base path. This method returns immediately with a promise
+   * that resolves with the parsed object when the request completes.
+   */
+  static _xhrAsyncBase(url: string, options?: XhrOptions) {
+    return ApiManager.getBasePath().then((base:string) => {return ApiManager._xhrAsync(base + url, options)});
```
yebrahim: Woops.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -179,13 +179,17 @@ class FilesElement extends Polymer.Element {
     clearInterval(this._fileListRefreshIntervalHandle);
   }
 
-  _getNotebookUrlPrefix() {
-    let prefix = location.protocol + '//' + location.host + '/';
+  async _getNotebookUrlPrefix() {
+    // Notebooks that are stored on the VM requires the basepath.
+    let basepath = await ApiManager.getBasePath();
```
yebrahim: I would've preferred to keep the base path hidden in `ApiManager` and instead expose a `getNotebookUrlPrefix` there, but up to you. :)

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -124,13 +124,33 @@ class ApiManager {
...
 
   /**
+   * URL for retrieving the base path
+   */
+  static readonly basepathApiUrl = '/api/basepath';
+
+  /**
+   * URL for starting terminals, undefined until we call basepathApiUrl
+   */
+  static _basepath = '';
```
yebrahim: This is unnecessary now, everyone should use the promise, which will return immediately if it already had the value resolved.

chmeyers: Done

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
chmeyers: I think this would make sense if we were willing for location.protocol + '//' + location.host + basepath+ '/' to be there, but 'edit' and 'notebook' are specific to this file.  Currently ApiManager doesn't do anything with location.* either, so I'm going to leave it all here.

chmeyers:
PTAL.  The merge brought in another use of _xhrAsync that I had to replace.

PR #1476, opened 2017-06-30
[Polymer UI] Enable/Disable toolbar buttons 

yebrahim:
This PR enables/disables the toolbar buttons that rely on items being selected in both `datalab-files` and `datalab-sessions` elements.

image

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-files/datalab-files.html
+++ b/sources/web/datalab/polymer/components/datalab-files/datalab-files.html
@@ -51,7 +51,7 @@ the License.
       }
       .toolbar-button[disabled] {
         color: var(--disabled-fg-color);
-        background-color: var(--disabled-bg-color);
+        background-color: var(--secondary-bg-color);
       }
       .file-picker-small--false {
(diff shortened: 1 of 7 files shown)


sources/web/datalab/polymer/components/datalab-files/datalab-files.html
```diff
@@ -49,6 +49,10 @@
         width: 20px;
         padding-right: 5px;
       }
+      .toolbar-button[disabled] {
+        color: var(--disabled-fg-color);
+        background-color: var(--secondary-bg-color);
```
jimmc: Are you really changing the background color when you disable a button? It seems like that should not be the case, and therefore this line should not be here.

yebrahim: That's because by default a button's disabled state applies a different background color, I'm overriding that to keep it consistent with the toolbar's background color.

jimmc: Ack.

PR #1477, opened 2017-06-30
Make sign-in and sign-out work by adding gapi.

jimmc:
This PR implements the sign-in and sign-out buttons in the Account drop-down menu. It uses gapi to ask the user to select an account and consent to our use of a couple of scopes. The clientId is stored in the user settings, until we figure out how we actually want to handle that. We will likely also change the way the scopes are handled so as to incrementally request scopes as we need them. But for now, this provides enough functionality to start implementing and testing code that calls APIs that require sign-in.


sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -60,6 +60,7 @@ interface JupyterFile {
 interface UserSettings {
```
yebrahim: Looks like I forgot to remove this, this is defined and maintained in `common.d.ts` now along with `AppSettings`. I think we want to add the new field there instead, and remove this interface definition altogether. :)

sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -0,0 +1,87 @@
...
+    // Loads the gapi client library and the auth2 library together for efficiency.
+    // Loading the auth2 library is optional here since `gapi.client.init` function will load
+    // it if not already loaded. Loading it upfront can save one network request.
+    GapiManager.loadClientId()
+      .then(() => {
+        gapi.load('client:auth2', this._initClient.bind(this, signInChangedCallback));
+      });
+  }
+
+  static signIn(doPrompt: boolean) {
```
yebrahim: Block comments for public functions.

sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  static getSignedInEmail() {
+    return gapi.auth2.getAuthInstance().currentUser.get().getBasicProfile().getEmail();
+  }
+
+  static loadClientId() {
+    return ApiManager.getUserSettings()
+      .then((settings: common.UserSettings) => {
+        if (settings.oauth2ClientId) {
+          GapiManager.clientId =  settings.oauth2ClientId;
```
yebrahim: nit: extra space.

sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
```
yebrahim: `console.log` here if we don't find the client id in settings?

sources/web/datalab/polymer/modules/gapi.d.ts
```diff
@@ -0,0 +1,57 @@
+/** Definition of the gapi module we load from https://apis.google.com/js/api.js
```
yebrahim: I think this should live under `third_party/externs/ts/gapi`.

sources/web/datalab/polymer/modules/ApiManager.ts
jimmc: Done.

sources/web/datalab/polymer/modules/GapiManager.ts
jimmc: Done.

sources/web/datalab/polymer/modules/GapiManager.ts
jimmc: Done.

sources/web/datalab/polymer/modules/GapiManager.ts
jimmc: Done.

sources/web/datalab/polymer/modules/gapi.d.ts
jimmc: Done.

PR #1480, opened 2017-06-30
[Polymer UI] Add theme switcher setting

yebrahim:
This change adds a settings panel element, and adds the first setting to it for switching themes.

Note that settings are now loaded twice from the backend, so next we'll need a centralized cache for user and app settings. I'll do that as a separate change.

image

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
+++ b/sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
@@ -1,0 +1,35 @@
+<!--
+Copyright 2017 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+in compliance with the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License
+is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+or implied. See the License for the specific language governing permissions and limitations under
+the License.
+-->
+
+<link rel="import" href="../../components/shared-styles/shared-styles.html">
+
+<link rel="import" href="../../bower_components/paper-radio-group/paper-radio-group.html">
+<link rel="import" href="../../bower_components/paper-input/paper-input.html">
+
+<dom-module id="datalab-settings">
+  <template>
+    <style include="datalab-shared-styles">
+    </style>
+
+    <!--Theme-->
+    <div>Theme:</div>
+    <paper-radio-group selected="{{theme}}" on-paper-radio-group-changed="_themeChanged">
+      <paper-radio-button name="light">Light</paper-radio-button>
+      <paper-radio-button name="dark">Dark</paper-radio-button>
+    </paper-radio-group>
+
+  </template>
+</dom-module>
+
+<script src="datalab-settings.js"></script>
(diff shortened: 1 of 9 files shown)


sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
```diff
@@ -0,0 +1,44 @@
...
+      .setting {
+        padding: 15px;
+        border-bottom: 1px solid var(--border-color);
+      }
+    </style>
+
+    <!--Theme-->
+    <div class="setting">
+      <div>Theme:</div>
+      <paper-radio-group selected="{{theme}}" on-paper-radio-group-changed="_themeChanged">
```
blois: Minor nit- radio buttons seems like an odd choice here. It adds more visual clutter than a checkbox and doesn't scale as well as a dropdown menu.

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Called when element is attached to DOM. Gets the user settings and sets
+   * the visible ones to element properties.
+   */
+  ready() {
+    super.ready();
+
+    // TODO: add a cache for user/app settings and call it here instead.
+    ApiManager.getUserSettings()
```
blois: Should this be done while the dialog is opening, so the values don't change after it's been opened?

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
```diff
@@ -115,7 +116,9 @@
     <!--Settings dialog. This shows centered in the middle of the window-->
     <paper-dialog id="settingsDialog" class="dialog" with-backdrop>
```
blois: This is a bit odd to have the settingsDialog here rather than datalab-settings be the entire dialog.

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
```diff
@@ -115,7 +116,9 @@
...
       <div class="dialog-title">Settings</div>
-      <div class="dialog-content"></div>
+      <div class="dialog-content">
+        <datalab-settings></datalab-settings>
+      </div>
       <hr>
       <div>
         <paper-button dialog-dismiss raised>Close</paper-button>
```
blois: The button is left aligned, but material design normally has it right aligned.

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
yebrahim: Like several other pieces here, this design isn't final, and I do agree a dropdown makes more sense here, but we'll have to review this along with other parts before we're ready to make the switch to the new UI.

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
yebrahim: Why? The `datalab-settings` element can theoretically be embedded in any container. For example, we might want to switch to have the settings be its own page instead of a dialog.

sources/web/datalab/polymer/components/datalab-toolbar/datalab-toolbar.html
yebrahim: Fixed.

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.ts
yebrahim: Yes that's a good point. Fixed.

PR #1481, opened 2017-07-01
Fix small bug with /api/basepath

yebrahim:
This was driving me crazy, the errors we got were not remotely indicative of the problem!


yebrahim:
Thanks for the quick attention. :)

PR #1484, opened 2017-07-05
[Polymer UI] Make design more responsive with screen width

yebrahim:
This change adds some responsiveness to the `datalab-files` element, by splitting the toolbar into two sets of buttons: Add buttons, and File Operations buttons. Each set collapses separately into a dropdown to save space when the screen is too narrow.

To do this, the parent `datalab-app` element now listens for window size changes, and calls resize handlers on the currently active page. Currently `datalab-files` and `datalab-terminal` have handlers that make use of this event.
The actual collapsing is done in javascript because it moves the buttons dynamically to an alternative dropdown element. I could not get a similar effect just using CSS, let me know if you have any ideas here.

This is what it looks like with the Add toolbar collapsed:
image

This is when both toolbars are collapsed:
image

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -42,11 +42,21 @@ class DatalabAppElement extends Polymer.Element {
    */
   public routeData: Object;
 
+  private _boundResizeHandler: EventListenerObject;
+
   constructor() {
     super();
 
     // Set the pattern once to be the current document pathname.
     this.rootPattern = (new URL(this.rootPath)).pathname;
+
+    this._boundResizeHandler = this._resizeHandler.bind(this);
+    window.addEventListener('resize', this._boundResizeHandler, true);
+
+    // Will be called after the custom element is done rendering.
+    window.addEventListener('WebComponentsReady', () => {
+      this._resizeHandler();
+    });
   }
 
   static get is() { return 'datalab-app'; }
@@ -72,6 +82,15 @@ class DatalabAppElement extends Polymer.Element {
   }
 
   /**
+   * Called when the element is detached from the DOM. Cleans up event listeners.
+   */
+  disconnectedCallback() {
+    if (this._boundResizeHandler) {
+      window.removeEventListener('resize', this._boundResizeHandler);
+    }
+  }
+
+  /**
    * On changes to the current route, explicitly sets the page property
    * so it can be used by other elements.
    */
@@ -94,6 +113,13 @@ class DatalabAppElement extends Polymer.Element {
     Polymer.importHref(resolvedPageUrl, undefined, undefined, true);
 
     // If the new page has a resize handler, call it.
+    this._resizeHandler();
+  }
+
+  /**
+   * If the selected page has a resize handler, calls it.
+   */
+  _resizeHandler() {
     const selectedPage = this.$.pages.selectedItem;
     if (selectedPage && selectedPage._resizeHandler) {
       selectedPage._resizeHandler();
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -54,9 +54,7 @@ class DatalabAppElement extends Polymer.Element {
     window.addEventListener('resize', this._boundResizeHandler, true);
 
     // Will be called after the custom element is done rendering.
-    window.addEventListener('WebComponentsReady', () => {
-      this._resizeHandler();
-    });
+    window.addEventListener('WebComponentsReady', this._boundResizeHandler, true);
   }
 
   static get is() { return 'datalab-app'; }
@@ -87,6 +85,7 @@ class DatalabAppElement extends Polymer.Element {
   disconnectedCallback() {
     if (this._boundResizeHandler) {
       window.removeEventListener('resize', this._boundResizeHandler);
+      window.removeEventListener('WebComponentsReady', this._boundResizeHandler);
     }
   }
 
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -92,6 +92,12 @@ class DatalabAppElement extends Polymer.Element {
     const subpath = 'datalab-' + page
     const resolvedPageUrl = this.resolveUrl('../' + subpath + '/' + subpath + '.html');
     Polymer.importHref(resolvedPageUrl, undefined, undefined, true);
+
+    // If the new page has a resize handler, call it.
+    const selectedPage = this.$.pages.selectedItem;
+    if (selectedPage && selectedPage._resizeHandler) {
+      selectedPage._resizeHandler();
+    }
   }
 
 }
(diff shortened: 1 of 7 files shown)


sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -42,11 +42,21 @@ class DatalabAppElement extends Polymer.Element {
...
     super();
 
     // Set the pattern once to be the current document pathname.
     this.rootPattern = (new URL(this.rootPath)).pathname;
+
+    this._boundResizeHandler = this._resizeHandler.bind(this);
+    window.addEventListener('resize', this._boundResizeHandler, true);
+
+    // Will be called after the custom element is done rendering.
+    window.addEventListener('WebComponentsReady', () => {
```
jimmc: Is there a reason this call and the one on line 54 use different approaches to calling the same function?
Alternatively, can you call _resizeHandler from ready() instead of using a listener?

sources/web/datalab/polymer/components/datalab-files/datalab-files.html
```diff
@@ -159,48 +194,77 @@
...
-      only enable the button when one file is selected for now, until we get
-      the group moving/copying functionality too-->
-      <paper-button class="toolbar-button" on-click="_deleteSelectedItems"
-                    disabled$={{!selectedFile}}>
-        <iron-icon icon="delete"></iron-icon>
-        <span>Delete</span>
-      </paper-button>
+
+      <!--Default Add toolbar, contains Add buttons for notebook, file, and folder-->
+      <span id="addToolbar">
```
jimmc: For consistency with "FileOps", maybe call this "AddOps" instead of just "Add"?
Or maybe rename "FileOps" to something like "UpdateOps", since these Add ops are also file ops.
Or don't use "Ops" in either.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -598,6 +598,46 @@ class FilesElement extends Polymer.Element {
     }
   }
 
+  _toggleAltAddToolbar() {
+    this.$.altAddToolbar.toggle();
+  }
+
+  _toggleAltFileOpsToolbar() {
+      this.$.altFileOpsToolbar.toggle();
```
jimmc: dedent

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -598,6 +598,46 @@ class FilesElement extends Polymer.Element {
...
+  }
+
+  /**
+   * Called on window.resize, collapses elements to keep the element usable
+   * on small screens.
+   */
+  _resizeHandler() {
+    const width = this.$.toolbar.clientWidth;
+    // Collapse the add buttons on the toolbar
+    if (width < 800) {
```
jimmc: Please make these magic numbers constants.

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
```diff
@@ -68,7 +68,7 @@
       .iron-selected > .sidebar-button > * {
         color: var(--selection-fg-color);
       }
-      @media only screen and (max-width: 970px) {
+      @media only screen and (max-width: 800px) {
```
jimmc: Assuming this is supposed to happen at the same width as the javascript code reference to the same number, is there a way this number can be in one place and referenced by both html and javascript?

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: Not sure what you mean. Line 54 only attaches the event listener, it doesn't call it. Line 58 calls the resize handler explicitly to set the initial size on element load.

sources/web/datalab/polymer/components/datalab-files/datalab-files.html
yebrahim: Done. Renamed to `UpdateToolbar`.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: done.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: done.

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.html
yebrahim: These are actually logically different, and they aren't the same numbers either. These are screen width numbers, the numbers in the javascript are based on the element width, which is smaller.

As far as I know, there isn't an easy way to make this one a constant without using a CSS preprocessor.

jimmc: I saw 800 for the value that is now _addToolbarCollapseThreshold, which I assumed meant 800px. But if there is not an easy way to share a constant, then even if it were the same value, it would not be worth it for this one line. So OK as is.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -598,6 +601,46 @@ class FilesElement extends Polymer.Element {
...
+      Utils.moveElementChildren(this.$.addToolbar, this.$.altAddToolbar);
+      this.$.altAddToolbarToggle.style.display = "inline-flex";
+    } else {
+      Utils.moveElementChildren(this.$.altAddToolbar, this.$.addToolbar);
+      this.$.altAddToolbarToggle.style.display = "none";
+      this.$.altAddToolbar.close();
+    }
+
+    // Collapse the file operations buttons on the toolbar
+    if (width < this._updateToolbarCollapseThreshold) {
```
jimmc: This code looks a lot like the code at line 619. Is it worth writing a method for these 10 lines of code?

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: Oh I see, no you're right, there's no good reason for calling it differently.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: I'm not very comfortable writing very specialized functions, such as what's needed to fix this. The function will be unreadable and won't be a lot of sense as a standalone routine. I'd rather repeat the code if it's more readable.
Thoughts?

jimmc: I think creating a function could be clear enough if you could come up with a good descriptive name for the it, something like collapseOrUncollapseToolbarSegment. As it is, I think some of your comments are perhaps a bit confusing, they talk about collapsing elements, but this code also uncollapses elements if the size is large. But refactoring would not reduce the code by that much, it's more for preventing code drift in the future, where someone might add something to one and not the other, so I'm OK with your leaving it as is.

yebrahim: I agree about the general concept of not repeating code of course, but this case looks too specialized for a function, even just thinking about that function's parameters makes me want to redesign the whole thing.
I think I'll keep this for now, we can rethink it if we're touching this code again.

PR #1488, opened 2017-07-05
[Polymer UI] Add timeout interval setting

yebrahim:
This change adds the timeout interval setting to the settings element, with a tooltip that explains its usage. (The tooltip shows on mouseover, the screen shot is a little misleading)

It also triggers loading settings every time the dialog is opened instead of only once.

screenshot from 2017-07-05 15 31 59

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
+++ b/sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
@@ -14,8 +14,10 @@ the License.
 
 <link rel="import" href="../../components/shared-styles/shared-styles.html">
 
-<link rel="import" href="../../bower_components/paper-radio-group/paper-radio-group.html">
+<link rel="import" href="../../bower_components/paper-icon-button/paper-icon-button.html">
 <link rel="import" href="../../bower_components/paper-input/paper-input.html">
+<link rel="import" href="../../bower_components/paper-radio-group/paper-radio-group.html">
+<link rel="import" href="../../bower_components/paper-tooltip/paper-tooltip.html">
 
 <dom-module id="datalab-settings">
   <template>
@@ -27,6 +29,28 @@ the License.
         padding: 15px;
         border-bottom: 1px solid var(--border-color);
       }
+      .input-container {
+        display: flex;
+        height: 35px;
+      }
+      .input-container > * {
+        margin-right: 10px;
+      }
+      .input-box {
+        width: 150px;
+        --primary-text-color: var(--primary-fg-color);
+      }
+      .super-info {
+        width: 18px;
+        height: 18px;
+        color: #888;
+        padding: 0px;
+      }
+      #idleTimeoutTooltip {
+        --paper-tooltip: {
+          font-size: 13px;
+        }
+      }
     </style>
 
     <!--Theme-->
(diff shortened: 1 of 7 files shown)


sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
```diff
@@ -14,21 +14,55 @@
 
 <link rel="import" href="../../components/shared-styles/shared-styles.html">
 
-<link rel="import" href="../../bower_components/paper-radio-group/paper-radio-group.html">
+<link rel="import" href="../../bower_components/neon-animation/web-animations.html">
 <link rel="import" href="../../bower_components/paper-input/paper-input.html">
+<link rel="import" href="../../bower_components/paper-radio-group/paper-radio-group.html">
+<link rel="import" href="../../bower_components/paper-tooltip/paper-tooltip.html">
+<link rel="import" href="../../bower_components/paper-progress/paper-progress.html">
```
jimmc: alpha ordering?

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
```diff
@@ -14,21 +14,55 @@
...
       .setting {
         padding: 15px;
         border-bottom: 1px solid var(--border-color);
       }
+      .input-container {
+        display: flex;
+        height: 35px;
+      }
+      .input-container > * {
+        margin-right: 10px;
```
jimmc: Would setting padding-right on input-container do the same as this? That seems simpler.

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
```diff
@@ -14,21 +14,55 @@
...
+      }
+      #idleTimeoutTooltip {
+        --paper-tooltip: {
+          font-size: 13px;
+        }
+      }
     </style>
 
+    <!--Thin progress bar that shows until files are loaded-->
+    <paper-progress id="progressBar" indeterminate disabled={{!_busy}}></paper-progress>
```
jimmc: Is this progress bar supposed to show progress, or is it only a "busy" indicator? If the latter, is this the best element to use?

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.ts
```diff
@@ -42,19 +56,41 @@ class SettingsElement extends Polymer.Element {
...
     return SettingsManager.setUserSettingAsync('theme', this.theme)
       .then(() => document.dispatchEvent(new Event('ThemeChanged')));
   }
 
+  _idleTimoutIntervalChanged() {
+    // TODO: Show success/error status to user
+    this._busy = true;
+    return SettingsManager.setUserSettingAsync('idleTimeoutInterval', this.idleTimeoutInterval)
+      .catch((e: Error) => console.log('Error updating idle timeout interval: ', e))
+      .then(() => this._busy = false);
```
jimmc: This should probably be .finally rather than .then, I think it would read better (and just in case there is an error in the catch block).

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
yebrahim: done.

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
yebrahim: Not really, this here is setting some spacing between the `input-container`'s elements, not for the `input-container` itself, and the elements are in row-layout.

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.html
yebrahim: It's only a busy indicator, but I'm using this element to match the style used in other elements. I think it looks better than a spinning circle, which is the other popular element.

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.ts
yebrahim: ES6 doesn't support `.finally` though, so you have to resort to `.catch` followed by `.then`.

jimmc: Sigh ... too many languages.

PR #1489, opened 2017-07-06
[Polymer UI] Upload file functionality

yebrahim:
This PR adds simple multi-file upload functionality, which works seamlessly for files that are not too big. This reads the file into memory, then uses the Jupyter's file save API, which takes the entire file in payload on notebook==4.x. Moving to notebook==5.x allows for sending file chunks, which removes that limit. See https://github.com/jupyter/notebook/pull/2162.

image

**For large uploads**: At ~20MB on my laptop I started noticing UI freezing (from the `FileReader`), and around ~100MB the browser would crash when trying to send the file's contents. We can separately investigate how to better read the file without freezing the UI (maybe service workers?).

# Update:
Added a warning dialog if at least one selected file is larger than 25MB.
image

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-files/datalab-files.html
+++ b/sources/web/datalab/polymer/components/datalab-files/datalab-files.html
@@ -178,6 +178,7 @@ the License.
         <iron-icon icon="file-upload"></iron-icon>
         <span>Upload</span>
       </paper-button>
+      <!--This is hidden but its functionality is used by the alt upload button above-->
       <input type="file" id="altFileUpload" multiple style="display: none"
              on-change="_upload">
       <paper-button class="toolbar-button" on-click="_copySelectedItem"
--- a/sources/web/datalab/polymer/components/datalab-files/datalab-files.html
+++ b/sources/web/datalab/polymer/components/datalab-files/datalab-files.html
@@ -174,10 +174,12 @@ the License.
         <iron-icon icon="add-box"></iron-icon>
         <span>Folder</span>
       </paper-button>
-      <paper-button class="toolbar-button">
+      <paper-button class="toolbar-button" on-click="_altUpload">
         <iron-icon icon="file-upload"></iron-icon>
         <span>Upload</span>
       </paper-button>
+      <input type="file" id="altFileUpload" multiple style="display: none"
+             on-change="_upload">
       <paper-button class="toolbar-button" on-click="_copySelectedItem"
                     disabled$={{!selectedFile}}>
         <iron-icon icon="content-copy"></iron-icon>
--- a/sources/web/datalab/polymer/components/datalab-files/datalab-files.html
+++ b/sources/web/datalab/polymer/components/datalab-files/datalab-files.html
@@ -21,6 +21,7 @@ the License.
 
 <link rel="import" href="../../bower_components/iron-icon/iron-icon.html">
 <link rel="import" href="../../bower_components/iron-icons/editor-icons.html">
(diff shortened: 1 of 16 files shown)


sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -375,6 +375,84 @@ class FilesElement extends Polymer.Element {
...
+
+            resolve(model);
+          }
+
+          // TODO: handle file reading errors.
+          reader.onerror = () => {
+            reject('Error reading file.');
+          }
+
+          // TODO: this will freeze the UI on large files (>~20MB on my laptop) until
```
jimmc: If there is a size beyond which the UI might freeze, I think we should limit the upload size to files smaller than that. You should be able to check file.size and reject the request if the file is too big.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -375,6 +375,84 @@ class FilesElement extends Polymer.Element {
...
+   * This is the entry point for file upload functionality. This is here to avoid using
+   * the very ugly <input> element for file uploads.
+   */
+  _altUpload() {
+    this.$.altFileUpload.click();
+  }
+
+  /**
+   * Gets called after the user selected one or more files from the upload modal.
+   * For each of the selected files, reads its contents, converts it to base64, then
```
jimmc: Do you need to read the file into memory in the client? Can you use the xhr upload API?

sources/web/datalab/polymer/components/datalab-files/datalab-files.html
```diff
@@ -174,10 +174,13 @@
         <iron-icon icon="add-box"></iron-icon>
         <span>Folder</span>
       </paper-button>
-      <paper-button class="toolbar-button">
+      <paper-button class="toolbar-button" on-click="_altUpload">
```
jimmc: Can you post a screenshot? I don't see how the user selects a file to upload if the file input field is hidden.

yebrahim: Sure. The upload button is still visible though. This is just a trick to avoid having the user interact with the ugly `<input>` element.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: We're limited by the backend APIs Jupyter offers for now, which expects a `PUT` request with the file's content in payload. When we move away from Jupyter for content management, we'll be able to roll out a possible better upload functionality.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: The freezing is proportional to the file size.
I think we need more discussion before we make such a decision. The user might not mind waiting for a few seconds until the read/upload is complete.

jimmc: I agree with "we need more discussion". I would like to understand how this works and what the use cases are. The way you describe this feature, it sounds like it would be easy for the user to have a pretty bad experience with this. Maybe we can talk in person tomorrow.

sources/web/datalab/polymer/components/datalab-files/datalab-files.html
yebrahim: The label can't be changed easily unfortunately, since it's a browser implementation. The only way I found to change it is to write our own open dialog.

PR #1492, opened 2017-07-07
[Polymer UI] Add a text editor

yebrahim:
This PR adds a text editor that looks close to the old UI, and is similarly based on CodeMirror. This adds another entry point (`/editor`) that uses a new `datalab-editor` element, which accepts a file path. The path is now being passed as a querystring parameter.

The change also adds an "Edit as Text" toolbar button so that notebooks can be opened in the editor.

Currently, the editor's toolbar buttons don't do anything, I'll add that functionality in a separate PR.

_Updated to show file name path, haven't updated theme screenshots._

Edit as text button:
image

File name/path:
image

Light theme:
image

Dark theme:
image


sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
```diff
@@ -0,0 +1,127 @@
...
+              console.log('Could not load specified file: ', e);
+              return '';
+            })
+        } else {
+          return '';
+        }
+      })
+      // Create the codemirror element and load the contents in it.
+      .then((content: string) => {
+        // TODO: try to detect the language of the file before creating
```
jimmc: Can you pass the filename in here as well and use that to help determine the file type?

sources/web/datalab/static.ts
```diff
@@ -211,6 +212,9 @@ function requestHandler(request: http.ServerRequest, response: http.ServerRespon
     if (pathname === '/files' || pathname === '/sessions' || pathname === '/terminal') {
       pathname = '/index.html';
     }
+    if (pathname === '/editor') {
```
jimmc: This would perhaps read better as else-if.

sources/web/datalab/static.ts
```diff
@@ -211,6 +212,9 @@ function requestHandler(request: http.ServerRequest, response: http.ServerRespon
...
+      pathname = '/editor.html';
+    }
     if (pathname === '/index.css') {
```
jimmc: This could also be else-if.

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.html
```diff
@@ -0,0 +1,106 @@
...
+  <!--Dark theme-->
+  <link rel="import" type="css" href="../../bower_components/codemirror/theme/icecoder.css">
+  <template>
+    <style include="datalab-shared-styles">
+      :host {
+        background-color: var(--primary-bg-color);
+        height: 100%;
+      }
+      #editorContainer {
+        height: calc(100% - 2 * var(--toolbar-height) - 1px); /* two toolbars, once progress bar*/
```
yebrahim: Sounds good, done.

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.html
```diff
@@ -0,0 +1,106 @@
...
+      </paper-button>
+    </div>
+    <div id="editorContainer"></div>
+
+  </template>
+</dom-module>
+
+<script src="../../bower_components/codemirror/lib/codemirror.js"></script>
+<script src="../../bower_components/codemirror/mode/python/python.js"></script>
+<script src="../../modules/Utils.js"></script>
```
yebrahim: Done.

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    // Get the theme.
+    SettingsManager.getUserSettingsAsync()
+      .then((settings: common.UserSettings) => {
+        if (settings.theme) {
+          this._theme = settings.theme;
+        }
+      })
+      // Get the file contents, or empty string if no path is specified.
+      .then(() => {
```
yebrahim: Done.

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
```diff
@@ -0,0 +1,127 @@
...
+        if (this.filePath !== '') {
+          this._busy = true;
+          return ApiManager.getJupyterFile(this.filePath, true /*asText*/)
+            .then((file: JupyterFile) => file.content)
+            .catch((e: Error) => {
+              // TODO: Handle error visibly to the user.
+              console.log('Could not load specified file: ', e);
+              return '';
```
yebrahim: It depends how we want to handle the error. My thinking was if we fail to load the file, the editor should still load after showing an error to the user (error message not implemented yet). Otherwise the page will only half-load and show a broken UI.

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
yebrahim: Sure.

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
```diff
@@ -0,0 +1,127 @@
...
+    this._editor.setOption('theme', this._getThemeValue(datalabTheme));
+  }
+
+  /**
+   * Translates the Datalab theme value (e.g. "light") into one of the
+   * CodeMirror's themes. This theme's stylesheet needs to be loaded in
+   * the element's light DOM.
+   * @param datalabTheme Datalab theme value
+   */
+  _getThemeValue(datalabTheme: string) {
```
yebrahim: Done. Renamed to `_getCodeMirrorTheme`.

sources/web/datalab/static.ts
yebrahim: Done.

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
jimmc: Looking at the screenshots again, it looks like the filename does not appear anywhere on the page, so as a user I don't know what file I am editing. This could be a problem if I have two tabs open on two files that are similar. Am I missing something? Is the filename available to the user anywhere on that page? It must know it so it can save the file, right?
Given that the filename appears somewhere, let's make sure that, when the file fails to load, the filename attached to the editor window gets cleared. I want to make sure we won't fail to load the file, present an empty editing screen, then write that back to the original file when the user says Save.

yebrahim: You're right, I totally missed that. I'll work on adding it now.

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.html
```diff
@@ -0,0 +1,100 @@
...
+        <iron-icon icon="file-download"></iron-icon>
+        <span>Download</span>
+      </paper-button>
+    </div>
+    <div id="editorContainer"></div>
+
+  </template>
+</dom-module>
+
+<script src="../../bower_components/codemirror/lib/codemirror.js"></script>
```
blois: Why are these dependencies here rather than in datalab-editor.js? They are dependencies of the code, no?

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
```diff
@@ -0,0 +1,131 @@
...
+        if (this.filePath !== '') {
+          this._busy = true;
+          // Passing the asText=true parameter guarantees the returned type is not a directory.
+          // An error is thrown if it is.
+          return ApiManager.getJupyterFile(this.filePath, true /*asText*/)
+            .catch((e: Error) => {
+              // TODO: Handle error visibly to the user.
+              console.log('Could not load specified file: ', e);
+              return null;
+            })
```
blois: Is there a linter that would catch missing trailing semicolons?

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
```diff
@@ -0,0 +1,131 @@
...
+        this._editor = CodeMirror(this.$.editorContainer,
+                                  {
+                                    value: content,
+                                    mode: 'python',
+                                    lineNumbers: true,
+                                    lineWrapping: true,
+                                    theme: this._getCodeMirrorTheme(this._theme),
+                                  });
+      })
+      .catch((e: Error) => console.log('Error loading file: ' + e))
```
blois: Should this be doing more than just log?

sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
```diff
@@ -0,0 +1,131 @@
...
+        // the codemirror element. Perhaps use the file extension?
+        // TODO: load the mode dynamically instead of starting out with python.
+        let content = '';
+        if (file) {
+          content = <string>file.content;
+        }
+        this._editor = CodeMirror(this.$.editorContainer,
+                                  {
+                                    value: content,
+                                    mode: 'python',
```
blois: Is this only python files?

sources/web/datalab/polymer/components/datalab-files/datalab-files.html
```diff
@@ -91,7 +92,7 @@
       }
       #navbar {
         display: flex;
-        height: calc(var(--toolbar-height) - 1px); /* for progress bar */
+        height: calc(var(--toolbar-height) - var(--progressbar-height)); /* for progress bar */
```
blois: Explicit heights are icky for a case like this- no way to use flexbox?

sources/web/datalab/polymer/editor.html
```diff
@@ -0,0 +1,47 @@
...
+        bottom: 0;
+        left: 0;
+        right: 0;
+        min-width: 470px;
+      }
+    </style>
+    <script src="editor.js"></script>
+    <datalab-editor></datalab-editor>
+  </body>
+</html>
```
blois: Should files always have newlines?

sources/web/datalab/polymer/editor.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * This file contains functionality for the entry point of the Datalab editor. It
+ * wraps the <datalab-editor> element, and takes a "file" querystring parameter to
+ * load it in the editor.
+ */
+document.addEventListener('DOMContentLoaded', () => {
+
+  const editorElement = <DatalabEditorElement>document.querySelector('datalab-editor');
+
+  document.addEventListener('ThemeChanged', (e: CustomEvent) => {
```
blois: Seems like 'ThemeChanged' should be a constant somewhere.

sources/web/datalab/polymer/components/datalab-settings/datalab-settings.ts
```diff
@@ -80,7 +80,10 @@ class SettingsElement extends Polymer.Element {
    */
   _themeChanged() {
     return SettingsManager.setUserSettingAsync('theme', this.theme)
-      .then(() => document.dispatchEvent(new Event('ThemeChanged')));
+      .then(() => {
+        const e = new CustomEvent('ThemeChanged', {'detail': this.theme})
+        document.dispatchEvent(e);
```
blois: Any reason not to add explicit set/get methods on SettingsManager and have a static 'x_changed' event on that? Generally seems better to avoid going through the global document for events.

sources/web/datalab/polymer/editor.ts
```diff
@@ -0,0 +1,46 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/**
+ * This file contains functionality for the entry point of the Datalab editor. It
+ * wraps the <datalab-editor> element, and takes a "file" querystring parameter to
+ * load it in the editor.
+ */
+document.addEventListener('DOMContentLoaded', () => {
```
blois: Why does this need to wait on DOMContentLoaded? Are all dependencies not loaded when this script is executed?

sources/web/datalab/polymer/modules/Utils.ts
```diff
@@ -102,6 +102,26 @@ class Utils {
   }
 
   /**
+   * Parses the given querystring and returns a dictionary of keys and values.
+   * @param search querystring part of the url to parse, typically window.location.search
+   */
+  static getSearchParams(search: string) {
```
blois: Why not new URLSearchParams? This isn't something that I would recommend doing yourself.

sources/web/datalab/polymer/editor.ts
yebrahim: Would you suggest a common typescript file with event literals that's imported by all users? Do you have a better idea?

sources/web/datalab/polymer/modules/Utils.ts
yebrahim: wasn't aware this existed. :)

yebrahim:
👍. I'll check this in but we please feel free to continue the discussion. I'll tackle any further changes in another PR.

PR #1495, opened 2017-07-10
[Polymer UI] Enable multi-selection with modifier keys, (Un)Select All

yebrahim:
This PR adds the ability to select multiple items in the `item-list` element by clicking a select/unselect all button, or by using modifier keys: shift+click remembers the last selected item and selects all between that item and current selection, ctrl+click toggles selection of the current item, adding or removing it to the selected group.

The change also cleans up the code for `item-list` to move more things to Polymer's automatic event bindings, and avoid manual tracking of selected indices by using array mutation syntax.

Note that multi-selection is not currently useful yet, since toolbar buttons are activated on single-item selection, so this will be added next.

image

### DIFF
--- a/sources/web/datalab/polymer/components/item-list/item-list.ts
+++ b/sources/web/datalab/polymer/components/item-list/item-list.ts
@@ -33,14 +33,16 @@ class ItemClickEvent extends CustomEvent {
 
 /**
  * Two-column list element.
- * This element takes a list of two column names, and a list of row
- * objects, each containing values for each of the columns, an icon
- * name, and a selected property. The items are displayed in a table
- * form. Clicking an item selects it and unselects all other items.
- * Clicking the checkbox next to an item allows for multi-selection.
- * Double clicking an item fires an 'ItemClickEvent' event with this
- * item's index. Similarly, selection fires an 'ItemClickEvent' with
- * the most recent clicked item's index.
+ * This element takes a list of two column names, and a list of row objects,
+ * each containing values for each of the columns, an icon name, and a selected
+ * property. The items are displayed in a table form. Clicking an item selects
+ * it and unselects all other items. Clicking the checkbox next to an item
+ * allows for multi-selection. Shift and ctrl keys can also be used to select
+ * multiple items.
+ * Double clicking an item fires an 'ItemClickEvent' event with this item's index.
+ * Selecting an item by single clicking it changes the selectedIndices
+ * property. This also notifies the host, which can listen to the
+ * selected-indices-changed event.
  * If the "hide-header" attribute is specified, the header is hidden.
  * If the "disable-selection" attribute is specified, the checkboxes are
  * hidden, and clicking items does not select them.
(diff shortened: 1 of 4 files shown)


sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -90,93 +96,138 @@ class ItemListElement extends Polymer.Element {
...
-    let currentElement = childElement;
-    while (currentElement.tagName !== 'PAPER-ITEM' && !currentElement.classList.contains('row'))
-      if (currentElement.parentElement)
-        currentElement = currentElement.parentElement;
-      else
-        // This should not happen
-        throw new Error('Could not find parent row element for: ' + childElement.tagName);
-    return currentElement;
+    // Save this index to enable multi-selection using shift later.
+    this._lastSelectedIndex = index;
```
jimmc: Are you following a particular spec for how this works? From my brief testing on Goobuntu and OSX, those two don't work they same way. But I don't know which is "right", so perhaps it doesn't matter.

yebrahim: Shift and ctrl are global modifier keys for UI design. While shift works the same way on most systems, including MacOS, ctrl doesn't, because it's used as a right click modifier on MacOS. We might want to look into supporting cmd on MacOS, but it's not natively supported by javascript, so we'll have to do some tricks.

jimmc: To be more precise:
Using Finder on my mac, if I click on item 4, then command-click on item 2, then shift-click on item 6, the result is that items 2, 4, 5, and 6 are selected. If I do the same on the file browser on Goobutnu, the result is that all items 2 through 6 are selected.
On the Mac, if I click on 4 then 2 then 6, the result is that all items 2 through 6 are selected, whereas if I do that on Goobutu, the result is that items 4 through 6 are selected.
The devil is in the details.
I don't have a strong opinion as to how it should work, but I expect someone will, so it would be good to be able to point to a spec, or say that we do it the same as (insert whoever we do it exactly the same as).

blois: metaKey?

yebrahim: @blois elaborate? :)
@jimmc The behavior is the same for all systems for me. Nautilus and Finder (and I believe Windows Explorer) behave the same. This is also the same behavior as Drive. Single selecting item 4, then 2, then shift selecting 6 should end up with items 2 through 6 selected.

jimmc: In all of my examples, the first click was a single-select, the second click was an add-to select (control or command), and the third click was a range-select (shift). So after the first two clicks, items 2 and 4 are selected, and the result after the next click depends on which was selected first and what platform it is on. Finder appears to remember the last item clicked and uses that for the range, as your code does, but Nautilus appears to remember the first item clicked that is still selected - or something like that, I did not try to fully characterize.

blois: Event.metaKey is what maps to Cmd on Mac, which is often used on OSX in place of Ctrl.

yebrahim: I see. Thanks, I'll add that.

PR #1496, opened 2017-07-10
[Polymer UI] Add skeleton Data pane with dummy data.

jimmc:
This PR implements an initial Data pane. The intent of this pane is to provide easy browsing of BigQuery tables and eventually other data. At the moment, this data pane includes only dummy data.
Clicking the Search button populates the Collections list. Clicking on a Collection populates the Datasets list. Clicking on a Dataset populates the Details pane on the right.
For now, the Data button in the left nav bar only appears when showing the Data pane, so for now you must enter the URL to get to that pane.
datatab

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-data/datalab-data.html
+++ b/sources/web/datalab/polymer/components/datalab-data/datalab-data.html
@@ -80,7 +80,7 @@ the License.
     <!--Toolbar with action buttons -->
     <div id="toolbar">
       <paper-button class="toolbar-button" on-click="_search">
-        <iron-icon icon="refresh"></iron-icon>
+        <iron-icon icon="search"></iron-icon>
         <span>Search</span>
       </paper-button>
     </div>
@@ -93,7 +93,7 @@ the License.
         <div>
           Collections Filter:
             <paper-input id="collectionsFilterBox" label="Optional filter value" spellcheck=false autofocus
-                         value="{{collectionsFilterValue}}" no-label-float>
+                         value="{{_collectionsFilterValue}}" no-label-float>
             </paper-input>
         </div>
         <!-- list of collections -->
@@ -107,7 +107,7 @@ the License.
         <div>
           Tables Filter:
             <paper-input id="tablesFilterBox" label="Optional filter value" spellcheck=false autofocus
-                         value="{{tablesFilterValue}}" no-label-float>
+                         value="{{_tablesFilterValue}}" no-label-float>
             </paper-input>
         </div>
         <!-- list of tables -->
--- a/sources/web/datalab/polymer/components/datalab-data/datalab-data.html
+++ b/sources/web/datalab/polymer/components/datalab-data/datalab-data.html
@@ -1,0 +1,135 @@
+<!--
+Copyright 2017 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+in compliance with the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License
+is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+or implied. See the License for the specific language governing permissions and limitations under
+the License.
+-->
+
+<link rel="import" href="../../components/item-list/item-list.html">
+<link rel="import" href="../../components/shared-styles/shared-styles.html">
+
+<dom-module id="datalab-data">
+  <template>
+    <style include="datalab-shared-styles">
+      :host {
+        --data-style: {
+          font-family: monospace;
+          font-size: 14px;
+        }
+      }
+      div {
+        font-family: var(--app-content-font-family);
+        font-size: var(--app-content-font-size);
+        color: var(--app-primary-color);
+      }
+      div.datasetDrillDown {
+        display: flex;
+        margin: 15px;
+      }
+      .collectionsColumn {
+        width: 33%;
+      }
+      .datasetsColumn {
+        width: 33%;
+      }
+      .filteredColumn {
+        margin: 10px;
+      }
+      detailsColumn {
+        width: 33%;
+      }
+      #toolbar {
+        height: calc(var(--toolbar-height) - 1px); /* for border-bottom */
+        width: 100%;
+        background-color: var(--app-secondary-color);
+        border-bottom: 1px solid #eee;
+      }
+      .toolbar-button {
+        color: var(--app-primary-color);
+        margin: 4px;
+        font-weight: normal;
+        border-radius: 0px;
+      }
+      .toolbar-button > iron-icon {
+        width: 20px;
+        padding-right: 5px;
+      }
+      .collections-container {
+        height: 100%;
+        display: flex;
+      }
+      .datasets-container {
+        height: 100%;
+        display: flex;
+      }
+      #collections {
+        flex-grow: 1;
+      }
+      #datasets {
+        flex-grow: 1;
+      }
+    </style>
+
+    <!--Toolbar with action buttons -->
+    <div id="toolbar">
+      <paper-button class="toolbar-button" on-click="_search">
+        <iron-icon icon="refresh"></iron-icon>
+        <span>Search</span>
+      </paper-button>
+    </div>
+
+    <!-- Multi-column layout: collections (BQ datasets), datasets (BQ tables) -->
+    <div class="datasetDrillDown">
+
+      <!-- Collections (BQ datasets) -->
+      <div class="collectionsColumn filteredColumn">
+        <div>
+          Project:
+            <paper-input id="projectBox" label="Select a project" spellcheck=false autofocus
+                         value="{{projectValue}}" no-label-float>
+            </paper-input>
+        </div>
+        <div>
+          Collections Filter:
+            <paper-input id="collectionsFilterBox" label="Optional filter value" spellcheck=false autofocus
+                         value="{{collectionsFilterValue}}" no-label-float>
+            </paper-input>
+        </div>
+        <!-- list of collections -->
+        <div class='collections-container'>
+          <item-list id="collections" hide-header="true"></item-list>
+        </div>
+      </div>
+
+      <!-- Datasets (BQ tables) -->
+      <div class="datasetsColumn filteredColumn">
+        <div>
+          Datasets Filter:
+            <paper-input id="datasetsFilterBox" label="Optional filter value" spellcheck=false autofocus
+                         value="{{datasetsFilterValue}}" no-label-float>
+            </paper-input>
+        </div>
+        <!-- list of datasets (tables) -->
+        <div class='datasets-container'>
+          <item-list id="datasets" hide-header="true"></item-list>
+        </div>
+      </div>
+
+      <!-- Details pane -->
+      <div class="detailsColumn">
+        <div id="detailsPane"></div>
+      </div>
+
+    </div>
+
+  </template>
+</dom-module>
+
+<script src="datalab-data.js"></script>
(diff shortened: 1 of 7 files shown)


sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -0,0 +1,135 @@
...
+<link rel="import" href="../../components/item-list/item-list.html">
+<link rel="import" href="../../components/shared-styles/shared-styles.html">
+
+<dom-module id="datalab-data">
+  <template>
+    <style include="datalab-shared-styles">
+      :host {
+        --data-style: {
+          font-family: monospace;
+          font-size: 14px;
```
yebrahim: We have a shared variable for this: `--app-content-font-size`, although it's 13px. If that's too small we should add another variable for this page, even if it's here.

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -0,0 +1,135 @@
...
+        font-family: var(--app-content-font-family);
+        font-size: var(--app-content-font-size);
+        color: var(--app-primary-color);
+      }
+      div.datasetDrillDown {
+        display: flex;
+        margin: 15px;
+      }
+      .collectionsColumn {
+        width: 33%;
```
yebrahim: Are these two columns always going to have the same width? Maybe we can use flex to allow one to be bigger than the other. Also maybe not for this skeleton PR, but worth adding a TODO?

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -0,0 +1,135 @@
...
+        margin: 10px;
+      }
+      detailsColumn {
+        width: 33%;
+      }
+      #toolbar {
+        height: calc(var(--toolbar-height) - 1px); /* for border-bottom */
+        width: 100%;
+        background-color: var(--app-secondary-color);
+        border-bottom: 1px solid #eee;
```
yebrahim: We have a `--border-color` var.

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -0,0 +1,135 @@
...
+      }
+      .toolbar-button {
```
yebrahim: FYI, I think these rules should move to `shared-styles` at some point. It seems we're using these on every page.

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -0,0 +1,135 @@
...
+      }
+      #datasets {
+        flex-grow: 1;
+      }
+    </style>
+
+    <!--Toolbar with action buttons -->
+    <div id="toolbar">
+      <paper-button class="toolbar-button" on-click="_search">
+        <iron-icon icon="refresh"></iron-icon>
```
yebrahim: Icon.

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -0,0 +1,135 @@
...
+      </paper-button>
+    </div>
+
+    <!-- Multi-column layout: collections (BQ datasets), datasets (BQ tables) -->
+    <div class="datasetDrillDown">
+
+      <!-- Collections (BQ datasets) -->
+      <div class="collectionsColumn filteredColumn">
+        <div>
+          Project:
```
yebrahim: I think the project selector should move outside the multi-page selector, and has its own 100%-width div? The pages of the selector should all have the same height to help with discoverability.

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -0,0 +1,135 @@
...
+
+    <!--Toolbar with action buttons -->
+    <div id="toolbar">
+      <paper-button class="toolbar-button" on-click="_search">
+        <iron-icon icon="refresh"></iron-icon>
+        <span>Search</span>
+      </paper-button>
+    </div>
+
+    <!-- Multi-column layout: collections (BQ datasets), datasets (BQ tables) -->
```
yebrahim: I think we should figure out which terminology we want to use early on, otherwise reading the code will be confusing. I think we should stick to using BQ terminology, since it's the main scenario for big data. Using GCS is an "also-supported" in my opinion.
I also think calling a GCS file a table might work, so how about calling the two columns: containers, and tables?

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -0,0 +1,210 @@
...
+/// <reference path="../input-dialog/input-dialog.ts" />
+/// <reference path="../item-list/item-list.ts" />
+
+interface Collection {
+  name : string;
+  status : string;
+  datasets? : Array<Dataset>;
+}
+
+interface Dataset {
```
yebrahim: Yes, we definitely need to resolve the language here! This confused me for a few seconds. :)

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -0,0 +1,210 @@
...
+  status : string;
+}
+
+/**
+ * Data element for Datalab.
+ * Contains a form for initial user input, and a list of datasets.
+ */
+class DataElement extends Polymer.Element {
+
+  public projectValue : string;
```
yebrahim: Comments for at least the public properties.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -0,0 +1,210 @@
...
+        value: 'no-filter',
+      },
+    };
+  }
+
+  ready() {
+    super.ready();
+
+    this._collectionsList = [];
+    const collectionsElement = this.shadowRoot.querySelector('#collections')
```
yebrahim: Polymer provides: `= this.$.collections`.
Note this will only work with camel case names in HTML.

Also, I think we should cast the element to its type here in TS, especially if it's an actual Polymer element, this way you'll get type checking for its members. So for this case:
```
const collectionsElement = <ItemListElement>this.$.collections;
```

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -0,0 +1,210 @@
...
+  _search() {
+    this._collectionsList = this._generateFakeCollectionsListForTesting();
+    this._renderCollectionsList();
+  }
+
+  /**
+   * Creates a new ItemListRow object for each entry in the collections list, and sends
+   * the created list to the item-list to render.
+   */
+  _renderCollectionsList() {
```
yebrahim: This should probably be an observable on `_collectionsList`, instead of calling it manually every time the list changes.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -0,0 +1,210 @@
...
+    this._renderDatasetsList();
+  }
+
+  _clearDatasetsList() {
+    this._datasetsList = [];
+    this._renderDatasetsList();
+  }
+
+  _showDetailsForDataset(dataset: Dataset) {
+    this.$.detailsPane.innerHTML = 'Selected dataset: ' + dataset.name;
```
yebrahim: Are we planning on adding an element for showing the details? Maybe we should rename the `details-pane` to `file-details` and add a `table-details` element, then here we can do `this.$.detailsPane.table = table`.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  _clearDatasetDetails() {
+    this.$.detailsPane.innerHTML = '';
+  }
+
+  /**
+   * Creates a new ItemListRow object for each entry in the collections list, and sends
+   * the created list to the item-list to render.
+   */
+  _renderDatasetsList() {
```
yebrahim: Should similarly be an observable on `_datasetList` changes.

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.ts
```diff
@@ -24,17 +24,27 @@ class SidebarElement extends Polymer.Element {
    */
   public page: string;
 
+  private _showDataTab: boolean;
```
yebrahim: You could also remove the link from the sidebar, then manually typing `/data` in the address bar would open the data page, even without this code here.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -0,0 +1,210 @@
...
+      this._showDetailsForDataset(this._datasetsList[selectedIndices[0]]);
+    } else {
+      this._clearDatasetDetails();
+    }
+  }
+
+  /**
+   * Called when the element is detached from the DOM. Cleans up event listeners.
+   */
+  disconnectedCallback() {
```
yebrahim: You probably won't need this, so I'd get rid of it before forgetting it.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
jimmc: I expect I will be adding some kind of table-details element. I'll figure that out later.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
jimmc: Done.

sources/web/datalab/polymer/components/datalab-sidebar/datalab-sidebar.ts
jimmc: Yeah, that seems reasonable. Done.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
jimmc: Done.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
jimmc: Done.

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
yebrahim: Oh I meant using an actual search icon for search instead of refresh. :)

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -85,18 +85,12 @@
       </paper-button>
     </div>
 
-    <!-- Multi-column layout: collections (BQ datasets), datasets (BQ tables) -->
-    <div class="datasetDrillDown">
+    <!-- Multi-column layout: collections (BQ datasets), tables -->
```
yebrahim: BQ datasets or GCS buckets?

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+/**
+ * Data element for Datalab.
+ * Contains a form for initial user input, and a list of tables.
+ */
+class DataElement extends Polymer.Element {
+
+  /** The value the user can enter to filter the collections. */
+  public collectionsFilterValue : string;
```
yebrahim: I'd add a TODO here about what this is here for. I'm assuming there should be an observer that should call some filter function on the `item-list` when this changes?

yebrahim: Also, being a public property implies it's an attribute that can be used to customize the element from the host, which doesn't seem like the case here. Should these two be private properties?

PR #1502, opened 2017-07-13
Comment out broken --idle-timeout command line option.

jimmc:
The --idle-timeout command line option got broken after some recent code changes. This PR disables this command line option for now until we figure out how we want to fix it.

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -280,6 +280,7 @@ def flags(parser):
         default=_DATALAB_DEFAULT_DISK_SIZE_GB,
         help='size of the persistent disk in GB.')
 
+    """ Broken due to changes to the notebook server, disable until we fix it.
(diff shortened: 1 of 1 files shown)


tools/cli/commands/create.py
```diff
@@ -280,6 +280,7 @@ def flags(parser):
         default=_DATALAB_DEFAULT_DISK_SIZE_GB,
         help='size of the persistent disk in GB.')
 
+    """ Broken due to changes to the notebook server, disable until we fix it.
```
ojarjur: Let's just remove it for now and then add it back when we are ready with the fix.

It will still be available in the git history, so there's no need to keep it commented out in the code.

jimmc: Done. Removed all the --idle-timeout code.

PR #1503, opened 2017-07-14
Report warnings from gcloud

ojarjur:
This change extends the `call_gcloud_quietly` method so that it
checks if gcloud reported any warnings, and if so cascades them
up to the user.

This prevents us from accidentally obscuring warnings about things
like calling gcloud in a way that is deprecated.


ojarjur:
This fixed #1503

PR #1506, opened 2017-07-17
Re-implement --idle-timeout CLI option, clean up settings update.

jimmc:
This PR re-implements the --idle-timeout command line option in the "datalab create" command that was recently removed (before being released), as it was broken due to a change in the implementation of how the idle timeout initial value was handled.

The implementation in this PR includes code that allows for specifying initial values for any user setting, which are applied when the user settings file is created during initialization to override the default user settings values. At present, the --idle-timeout command line option is the only user setting that is managed this way, but it would be easy to allow other initial user settings to be done using the same mechanism in the future by adding additional command line options.

The --idle-timeout command line option implemented in this PR relies on new code in the Datalab image that is also in this PR. It will do nothing if used with an earlier Datalab image that does not include this code.

This PR also includes a few code changes to the user settings updateing code that I am hoping might help track down or eliminate an intermittent problem I have had in which sometimes my settings.json file becomes corrupt with a "}" in place of the "," immediately following the startuppath value.

### DIFF
--- a/sources/web/datalab/idleTimeout.ts
+++ b/sources/web/datalab/idleTimeout.ts
@@ -45,10 +45,25 @@ export function init() {
   }
   if (shutdownCommand) {
     let idleTimeoutStr = userSettings.idleTimeoutInterval;
-    logging.getLogger().debug('idleTimeoutStr from user settings: ' + idleTimeoutStr);
+    if (idleTimeoutStr === undefined) {
+      logging.getLogger().debug('idleTimeoutStr from user settings is undefined');
+    } else {
+      logging.getLogger().debug('idleTimeoutStr from user settings: ' + idleTimeoutStr);
+    }
     if (!idleTimeoutStr) {
       idleTimeoutStr = process.env.DATALAB_IDLE_TIMEOUT;
-      logging.getLogger().debug('idleTimeoutStr from env: ' + idleTimeoutStr);
+      if (idleTimeoutStr === undefined) {
+        logging.getLogger().debug('idleTimeoutStr from env is undefined');
+      } else {
+        logging.getLogger().debug('idleTimeoutStr from env: ' + idleTimeoutStr);
+      }
+    }
+    // For instances (actually, PDs) created before idle-timeout was implemented,
+    // the user's settings file may not have a value for idleTimeoutInterval.
+    // In this case, we set it to 0s to continue using the the no-idle-timeout
+    // behavior that was in place when that PD was created.
+    if (idleTimeoutStr === undefined) {
+      idleTimeoutStr = '0s';
     }
     setIdleTimeoutInterval(idleTimeoutStr);
   } else {
--- a/sources/web/datalab/settings.ts
+++ b/sources/web/datalab/settings.ts
@@ -118,8 +118,54 @@ function copyDefaultUserSettings(userId: string) {
   const defaultUserSettingsPath = path.join(__dirname, 'config', DEFAULT_USER_SETTINGS_FILE);
   console.log('Copying default settings: ' + defaultUserSettingsPath + ' to: ' + userSettingsPath);
   // Copy the default user settings file into user's directory.
-  const data = fs.readFileSync(defaultUserSettingsPath);
-  fs.writeFileSync(userSettingsPath, data);
+  const defaultUserSettings = fs.readFileSync(defaultUserSettingsPath, {encoding: 'utf8'});
+  const initialUserSettings = process.env.DATALAB_INITIAL_USER_SETTINGS;
+  const mergedUserSettings : string = initialUserSettings ?
+      mergeUserSettings(defaultUserSettings, initialUserSettings) : defaultUserSettings;
+  fs.writeFileSync(userSettingsPath, mergedUserSettings);
+  // writeFileSync does not return a status; let's see if it wrote a file.
+  if (!fs.existsSync(userSettingsPath)) {
+    console.log('Failed to write new user settings file ' + userSettingsPath);
+  }
+}
+
+/**
+ * Merges two sets of user settings, giving priority to the second.
+ * Exported for testing.
+ */
+export function mergeUserSettings(defaultUserSettings: string, initialUserSettings: string): string {
+  let parsedDefaultUserSettings;
+  try {
+    if (defaultUserSettings) {
+      parsedDefaultUserSettings = JSON.parse(defaultUserSettings)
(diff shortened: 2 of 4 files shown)


sources/web/datalab/idleTimeout.ts
```diff
@@ -45,10 +45,25 @@ export function init() {
   }
   if (shutdownCommand) {
     let idleTimeoutStr = userSettings.idleTimeoutInterval;
-    logging.getLogger().debug('idleTimeoutStr from user settings: ' + idleTimeoutStr);
+    if (idleTimeoutStr === undefined) {
```
yebrahim: Do we need the if? Doesn't the original line already print `undefined` if it's undefined?

sources/web/datalab/settings.ts
```diff
@@ -118,8 +118,54 @@ function copyDefaultUserSettings(userId: string) {
...
-  const data = fs.readFileSync(defaultUserSettingsPath);
-  fs.writeFileSync(userSettingsPath, data);
+  const defaultUserSettings = fs.readFileSync(defaultUserSettingsPath, {encoding: 'utf8'});
+  const initialUserSettings = process.env.DATALAB_INITIAL_USER_SETTINGS;
+  const mergedUserSettings : string = initialUserSettings ?
+      mergeUserSettings(defaultUserSettings, initialUserSettings) : defaultUserSettings;
+  fs.writeFileSync(userSettingsPath, mergedUserSettings);
+  // writeFileSync does not return a status; let's see if it wrote a file.
+  if (!fs.existsSync(userSettingsPath)) {
+    console.log('Failed to write new user settings file ' + userSettingsPath);
```
yebrahim: Should this be an error log instead?

sources/web/datalab/settings.ts
```diff
@@ -118,8 +118,54 @@ function copyDefaultUserSettings(userId: string) {
...
+}
+
+/**
+ * Merges two sets of user settings, giving priority to the second.
+ * Exported for testing.
+ */
+export function mergeUserSettings(defaultUserSettings: string, initialUserSettings: string): string {
+  let parsedDefaultUserSettings;
+  try {
+    if (defaultUserSettings) {
```
yebrahim: Instead of these five lines (same below), can we just do `JSON.parse(defaultUserSettings || '{}')`?

sources/web/datalab/settings.ts
yebrahim: Same for other log lines below.

sources/web/datalab/settings.ts
```diff
@@ -172,33 +223,45 @@ function ensureDirExists(fullPath: string): boolean {
...
-  if (fs.existsSync(settingsPath)) {
-    try {
-      settings = <common.UserSettings>JSON.parse(fs.readFileSync(settingsPath, 'utf8') || '{}');
-    }
-    catch (e) {
-      console.log(e);
-      return false;
-    }
+  try {
+    settings = <common.UserSettings>JSON.parse(fs.readFileSync(settingsPath, 'utf8') || '{}');
```
yebrahim: It looks like this code path was trying to avoid synchronous read/write operations. Can we turn this whole path into a promise (or async/await) chain?

test/server-unit/settings_spec.js
```diff
@@ -0,0 +1,45 @@
...
+      const init = '';
+      const expected = '{}';
+      expect(settings.mergeUserSettings(dflt, init)).toBe(expected);
+    });
+
+    it('merges empty initial settings to unchanged default settings', function() {
+      const dflt = '{"a": 123, "b": "xyz"}';
+      const init = '';
+      const expected = '{"a":123,"b":"xyz"}';
+      expect(settings.mergeUserSettings(dflt, init)).toBe(expected);
```
yebrahim: I'd add one more `expect` line with the order of `dflt` and `init` swapped.

tools/cli/commands/create.py
```diff
@@ -579,11 +592,13 @@ def run(args, gcloud_compute, gcloud_repos,
         cmd.extend(['--zone', args.zone])
 
     enable_backups = "false" if args.no_backups else "true"
+    idle_timeout = args.idle_timeout
     console_log_level = args.log_level or "warn"
     user_email = args.for_user or email
     service_account = args.service_account or "default"
     # We have to escape the user's email before using it in the YAML template.
     escaped_email = user_email.replace("'", "''")
+    initial_user_settings = '{{"idleTimeoutInterval": "{0}"}}'.format(idle_timeout) if idle_timeout else ''
```
yebrahim: If we're building the initial settings here, the code should point out that this is JSON. I think we should be using `json.dumps` to make it obvious for future additions.

sources/web/datalab/settings.ts
jimmc: Yes, that's a better way. Done.

sources/web/datalab/settings.ts
jimmc: The previous code included a call to fs.existsSync, so it was already not asynchronous. I could work on making updateUserSetting a fully asynchronous function, but I'd rather not try to do that in the same PR as the idle-timeout changes. For this PR, I can revert the changes to updateUserSetting, leave it as is with these changes (my preference), or make one or two other small changes. Which would you prefer?

test/server-unit/settings_spec.js
jimmc: Done.

sources/web/datalab/settings.ts
yebrahim: Ah, you're right. I guess I'm fine with the console log calls then.
I don't see a lot of value in splitting the file, but feel free to do it if you think it cleans it up.

sources/web/datalab/settings.ts
yebrahim: I'm fine with adding the minimum changes needed to make this PR work, then separating the async changes in a different PR.
Having synchronous `file.exists` checks is much lighter than reads and writes, but you're right, we the code path already wasn't fully asynchronous.

tools/cli/commands/create.py
yebrahim: Ping on this one.

jimmc: Sorry, github's commenting system swallowed this one when I made a different change. This is now done.

yebrahim: Np. LGTM.

PR #1508, opened 2017-07-17
Skip reverse proxy requests from static handler

### LINKED ISSUE #1507: Tensorboard doesn't work with DATALAB_EXPERIMENTAL_UI=true
To reproduce the issue:

from google.datalab.ml import TensorBoard
TensorBoard.start(model_dir);

The output cell includes a link. Click the link, turn on dev mode in chrome, and you will see 500 errors after refreshing the page. It works without the flag.

yebrahim:
Fixes #1507.

Note this also moves handling the reverse proxy requests separately from other requests.

### DIFF
--- a/sources/web/datalab/server.ts
+++ b/sources/web/datalab/server.ts
@@ -275,7 +275,7 @@ function uncheckedRequestHandler(request: http.ServerRequest, response: http.Ser
       urlpath.indexOf('/oauthcallback') == 0) {
     // Start or return from auth flow.
     auth.handleAuthFlow(request, response, parsed_url, appSettings);
-  } else if (isStaticResource(urlpath)) {
+  } else if (!reverseProxy.isReverseProxyRequest(request) && isStaticResource(urlpath)) {
     staticHandler(request, response);
   } else {
     handleRequest(request, response, urlpath);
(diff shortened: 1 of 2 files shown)


sources/web/datalab/server.ts
```diff
@@ -275,7 +275,7 @@ function uncheckedRequestHandler(request: http.ServerRequest, response: http.Ser
       urlpath.indexOf('/oauthcallback') == 0) {
     // Start or return from auth flow.
     auth.handleAuthFlow(request, response, parsed_url, appSettings);
-  } else if (isStaticResource(urlpath)) {
+  } else if (!reverseProxy.isReverseProxyRequest(request) && isStaticResource(urlpath)) {
```
qimingj: If you move the check here, then you don't need to handle it again in https://github.com/yebrahim/datalab/blob/master/sources/web/datalab/server.ts#L148?

jimmc: While I think this code is OK for now, I do find the request handling code a bit tedious to follow. We have requestHandler() that calls uncheckedRequestHandler() that calls handleRequest(), which doesn't provide much distinction. The comment for uncheckedRequestHandler implies there are two ways we process requests, and the method then has two branches, but I'm not convinced that those two branches actually correspond to the two ways to process requests implied in the comment. It might be good to do a bit of refactoring or renaming to clarify, or at least so that the function comment is more closely aligned with the names of the functions that it is calling.

yebrahim: Done.

PR #1509, opened 2017-07-18
[Polymer UI] Show an error toast on connection errors

yebrahim:
This PR adds a paper-toast element that shows a connection error message whenever an XHR request fails with an unexpected code. The notification is dismissed whenever requests succeed.

A new custom even `NotificationEvent` can be fired on any child element, and is handled by the shell element `datalab-app` to show/hide the notification toast. Note this is currently not used anywhere, but it's part of this PR to open the discussion about the design of notifications.

image

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -55,6 +55,14 @@ class DatalabAppElement extends Polymer.Element {
 
     // Will be called after the custom element is done rendering.
     window.addEventListener('WebComponentsReady', this._boundResizeHandler, true);
+
+    ApiManager.disconnectedHandler = () => {
+      this.showNotification('Failed to connect to the server.', true /*sticky*/);
+    }
+    ApiManager.connectedHandler = () => {
+      this.hideNotification();
+    }
+
   }
 
   static get is() { return 'datalab-app'; }
@@ -131,20 +139,20 @@ class DatalabAppElement extends Polymer.Element {
    * @param sticky whether the toast should stick around until explicitly dismissed in code
    */
   showNotification(message: string, sticky?: boolean) {
-    this.$.connectionToast.close();
-    this.$.connectionToast.text = message;
+    this.$.notificationToast.close();
+    this.$.notificationToast.text = message;
     if (sticky) {
-      this.$.connectionToast.duration = 0;
+      this.$.notificationToast.duration = 0;
     }
-    this.$.connectionToast.open();
+    this.$.notificationToast.open();
   }
 
   /**
    * Hides the notification toast if it's open.
    */
   hideNotification() {
-    if (this.$.connectionToast.opened) {
-      this.$.connectionToast.close();
+    if (this.$.notificationToast.opened) {
+      this.$.notificationToast.close();
     }
   }
 
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -63,6 +63,14 @@ class DatalabAppElement extends Polymer.Element {
       this.hideNotification();
     }
 
+    // Handle notification events bubbled up from children.
+    this.addEventListener('notification', (e: NotificationEvent) => {
+      if (e.detail.show) {
+        this.showNotification(e.detail.message, e.detail.sticky);
+      } else {
+        this.hideNotification();
+      }
+    });
   }
 
   static get is() { return 'datalab-app'; }
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -125,6 +125,20 @@ class DatalabAppElement extends Polymer.Element {
     }
   }
 
+  showNotification(message: string, isError?: boolean, sticky?: boolean) {
+    this.$.toast.close();
+    this.$.toast.text = message;
+    this.$.toast.classList = isError ? ['error-toast'] : [];
+    if (sticky) {
+      this.$.toast.duration = 0;
+    }
+    this.$.toast.open();
+  }
+
+  hideNotification() {
+    this.$.toast.close();
+  }
+
 }
 
 customElements.define(DatalabAppElement.is, DatalabAppElement);
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -125,18 +125,27 @@ class DatalabAppElement extends Polymer.Element {
     }
   }
 
-  showNotification(message: string, isError?: boolean, sticky?: boolean) {
-    this.$.toast.close();
-    this.$.toast.text = message;
-    this.$.toast.classList = isError ? ['error-toast'] : [];
+  /**
+   * Shows a notification toast at the bottom of the page with the given message.
+   * @param message message to show in the notification toast
+   * @param sticky whether the toast should stick around until explicitly dismissed in code
+   */
+  showNotification(message: string, sticky?: boolean) {
+    this.$.connectionToast.close();
+    this.$.connectionToast.text = message;
     if (sticky) {
-      this.$.toast.duration = 0;
+      this.$.connectionToast.duration = 0;
     }
-    this.$.toast.open();
+    this.$.connectionToast.open();
   }
 
+  /**
+   * Hides the notification toast if it's open.
+   */
   hideNotification() {
-    this.$.toast.close();
+    if (this.$.connectionToast.opened) {
+      this.$.connectionToast.close();
+    }
   }
 
 }
--- a/sources/web/datalab/polymer/components/shared-styles/shared-styles.html
+++ b/sources/web/datalab/polymer/components/shared-styles/shared-styles.html
@@ -25,12 +25,6 @@ the License.
         --app-content-font-family: 'Roboto', 'Noto', sans-serif;
         --primary-text-color : var(--primary-fg-color);
       }
-      *, *::after, *::before {
-        /* Animate the color changes for when themes are switched. */
-        transition: border-color var(--app-animation-duration),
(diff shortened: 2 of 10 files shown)


sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -125,6 +125,29 @@ class DatalabAppElement extends Polymer.Element {
     }
   }
 
+  /**
+   * Shows a notification toast at the bottom of the page with the given message.
+   * @param message message to show in the notification toast
+   * @param sticky whether the toast should stick around until explicitly dismissed in code
+   */
+  showNotification(message: string, sticky?: boolean) {
```
jimmc: Given that you are only using this function to show connection methods, and you clear the toast whenever a connection succeeds, I think this should probably be called showConnectionNotification. Or maybe have a ConnectionToast component.

sources/web/datalab/polymer/modules/Utils.ts
```diff
@@ -112,4 +112,20 @@ class Utils {
     }
   }
 
+  /**
+   * Called when an API call fails, so that a connection error message can be shown.
+   */
+  static connectionFailed() {
```
jimmc: Are you planning on calling these methods from anywhere other than AppManager? It seems a little odd to bounce these calls through Utils.

sources/web/datalab/polymer/components/shared-styles/shared-styles.html
```diff
@@ -25,12 +25,6 @@
         --app-content-font-family: 'Roboto', 'Noto', sans-serif;
         --primary-text-color : var(--primary-fg-color);
       }
-      *, *::after, *::before {
```
jimmc: Does this change have something to do with the connection toast change?

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: I'm planning to also use this notification for quick snackbar-style messages, such as when a file operation succeeds. In that case, a different message will be shown, and sticky will be set to false.

sources/web/datalab/polymer/modules/Utils.ts
yebrahim: I'm open to suggestions. The reason I put them there was that `ApiManager` is solely responsible for talking to the backend, and shouldn't know anything about frontend notifications. Similarly, it shouldn't know about the shell element `datalab-app`.

sources/web/datalab/polymer/components/shared-styles/shared-styles.html
yebrahim: Yes. For some reason this interferes with the toast animation.
This wasn't very well thought out anyway, it's expensive and doesn't really add much other than theme switching, which isn't a frequent operation.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
jimmc: If you are going to be using this toast for things other than connection messages, then the component should probably not be called connectionToast.

sources/web/datalab/polymer/modules/Utils.ts
jimmc: I don't have a great idea, so go ahead.

sources/web/datalab/polymer/modules/Utils.ts
```diff
@@ -112,4 +112,20 @@ class Utils {
...
+    const host = <DatalabAppElement>document.querySelector('datalab-app');
+    host.showNotification('Failed to connect to the server.', true /*sticky*/);
+  }
+
+  /**
+   * Called when an API call succeeds, so that the connection error message can be hidden.
+   */
+  static connectionSucceeded() {
+    const host = <DatalabAppElement>document.querySelector('datalab-app');
+    host.hideNotification();
```
jimmc: Once you start using the toast for things other than connection failures, are you pretty confident that you will never be displaying some other toast when a connection succeeds and clears the toast? Maybe there are two XHRs outstanding after a connection failure, and one succeeds but then has some other error that shows a toast, but then the other completes just after that, clearing the non-connection-failure toast.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: You're right. Renamed to `notificationToast`.

sources/web/datalab/polymer/modules/Utils.ts
yebrahim: Here's an alternative approach: ApiManager now has handlers for connected and disconnected that can be set by its caller. This way it's separated from its usage, and the containing element, `datalab-app` in this case, can listen for changes in connected status.

A better way to do this would be to observe the value change on the `isConnected` property, but I didn't want to add the complication of observer code, so I'm allowing exactly one listener until we have the need for multiple.

sources/web/datalab/polymer/modules/Utils.ts
yebrahim: Whatever happens last wins. If the last xhr request fails then the toast appears with the error message, if the last one succeeded, this means we really are connected, so the toast should disappear.
Or am I missing something?

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -143,6 +144,18 @@ class ApiManager {
...
   /**
+   * Current connection status. Set to the last connection status.
+   */
+  static isConnected = true;
+
+  /**
+   * Handlers for connected/disconnected status. These will be called when the
+   * isConnected property changes value.
+   */
+  static connectedHandler: Function;
```
blois: What is the method signature for these? () => void?

Seems like it's common enough that they should be proper events.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -55,6 +55,14 @@ class DatalabAppElement extends Polymer.Element {
 
     // Will be called after the custom element is done rendering.
     window.addEventListener('WebComponentsReady', this._boundResizeHandler, true);
+
+    ApiManager.disconnectedHandler = () => {
```
jimmc: I like this better than the previous iteration. Adding the isConnected status in ApiManager is also an improvement.

To clarify the comment I made in Utils.ts about connect clearing the toast, as far as I can see everything will work fine for now with the current code because you are using the toast only for the connection status. I am concerned that an issue may arise later when you start using the toast for other things besides connection status. The essence of my concern is, what if a toast is posted about something other than connection status, but then immediately after that happens, a connection is reestablished; then the toast about that other thing will be cleared. But I think we can probably wait on that until it happens, if ever.

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -421,17 +437,34 @@ class ApiManager {
...
     const request = new XMLHttpRequest();
     const noCache = options.noCache || false;
+    const failureCodes = options.failureCodes;
 
     return new Promise((resolve, reject) => {
       request.onreadystatechange = () => {
         if (request.readyState === 4) {
           if (request.status === successCode) {
+
+            // If this is the first success after failures, call the connected handler
```
blois: This is Chrome only, right? If so, use fetch?

Much cleaner/better than XHR & readyStates & all.

sources/web/datalab/polymer/modules/ApiManager.ts
```diff
@@ -307,6 +322,7 @@ class ApiManager {
     const xhrOptions: XhrOptions = {
       method: 'POST',
       successCode: 201,
+      failureCodes: [409],
```
blois: 'failureCodes' implies a failure- perhaps 'recoverableStatusCodes'?

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: Hmmm.. my thinking was that the notification will show for success events, which necessarily require the connection to be established. Do you have a scenario in mind where we would need to show a notification for something that happens client-only?

sources/web/datalab/polymer/modules/ApiManager.ts
yebrahim: Added signature.
I'd love to use an event target and listeners here, but I don't know how without using a DOM element. Let me know if you have any ideas.

sources/web/datalab/polymer/modules/ApiManager.ts
yebrahim: These really are failures though, but on the server side. The client should just be able to gracefully handle them.
Would naming it `possibleFailureCodes` be better?

sources/web/datalab/polymer/modules/ApiManager.ts
yebrahim: Nothing in the Datalab front-end is Chrome-only. Looks like fetch is supported by all evergreen browsers except IE. Do you think it's worth looking for a polyfill?

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
jimmc: I do not have a specific scenario in mind. Your addition of the isConnected flag helps to reduce the possibility of race conditions in clearing the toast. I am OK with moving ahead with it as is.

sources/web/datalab/polymer/modules/ApiManager.ts
blois: The logic to properly handle XHRs is frequently done incorrectly and would be better off at least encapsulated in a reusable helper.

yebrahim: This is what we have here. This is the only place where XHRs are explicitly used, then it's encapsulated in `ApiManager._xhrTextAsync`.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: Cool. Waiting for another approval then.

PR #1512, opened 2017-07-18
[Polymer UI] Show success notification on file operations

yebrahim:
This change uses the recently added toast notification to show a success message when a file operation succeeds. A big part of this change is white spaces for the file upload change so I recommend using the `?w=1` querystring.

image


sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -392,90 +392,87 @@ class FilesElement extends Polymer.Element {
...
-          okLabel: 'Upload Anyway',
-        };
-
-        const result: BaseDialogCloseResult = 
-          await Utils.showDialog(BaseDialogElement, dialogOptions);
-          proceedWithUpload = result.confirmed === true;
-      }
+      const result: BaseDialogCloseResult = 
+        await Utils.showDialog(BaseDialogElement, dialogOptions);
+        proceedWithUpload = result.confirmed === true;
```
blois: Indentation?

Any progress on a lint/fix?

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -392,90 +392,87 @@ class FilesElement extends Polymer.Element {
...
-            reader.readAsDataURL(file);
-          });
-
-          // Now upload the file data to the backend server.
-          const uploadPromise = loadPromise
-            .then((model: JupyterFile) => ApiManager.saveJupyterFile(model));
-          uploadPromises.push(uploadPromise);
-        }
-      }
+    if (proceedWithUpload) {
```
blois: Why not return early? Would make the code more readable.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -392,90 +392,87 @@ class FilesElement extends Polymer.Element {
...
-          title: 'Warning: Large File',
-          messageHtml: warningMsg,
-          okLabel: 'Upload Anyway',
-        };
-
-        const result: BaseDialogCloseResult = 
-          await Utils.showDialog(BaseDialogElement, dialogOptions);
-          proceedWithUpload = result.confirmed === true;
-      }
+      const result: BaseDialogCloseResult =
```
blois: Random: is this type needed?

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -392,90 +392,87 @@ class FilesElement extends Polymer.Element {
...
+              format: 'base64',
+              content: itemData,
+            };
+
+            resolve(model);
+          }
+
+          // TODO: handle file reading errors.
+          reader.onerror = () => {
+            reject('Error reading file.');
```
blois: Should always reject with Errors, as these are directly translated to throws.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -392,90 +392,87 @@ class FilesElement extends Polymer.Element {
...
-          const uploadPromise = loadPromise
-            .then((model: JupyterFile) => ApiManager.saveJupyterFile(model));
-          uploadPromises.push(uploadPromise);
-        }
-      }
+    if (proceedWithUpload) {
+      files.forEach((file: File) => {
+
+        // First, load the file data into memory.
+        const loadPromise = new Promise((resolve, reject) => {
```
blois: Nit: promise creation should be as narrowly scoped as possible to make it easier to validate that it is always resolved.

Would recommend just:

```
const loadPromise = new Promise((resolve, reject) => {
  const reader = new FileReader();
  reader.onload = () => reader.result;
  reader.onerror = (e) => throw new Error('Error reading file');
  reader.readAsDataURL(file);
});
```

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -392,90 +392,87 @@ class FilesElement extends Polymer.Element {
...
+
+          const reader = new FileReader();
+
+          reader.onload = function() {
```
blois: This should probably be an arrow function, as it is elsewhere in this method.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -392,90 +392,87 @@ class FilesElement extends Polymer.Element {
...
+        });
+
+        // Now upload the file data to the backend server.
+        const uploadPromise = loadPromise
+          .then((model: JupyterFile) => ApiManager.saveJupyterFile(model));
+        uploadPromises.push(uploadPromise);
+      });
 
       // Wait on all upload requests before declaring success or failure.
       return Promise.all(uploadPromises)
```
blois: It's an async function- why not use that?

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -392,90 +392,87 @@ class FilesElement extends Polymer.Element {
    */
   async _upload() {
```
blois: Why is all of this logic in the UI layer? Really seems like it should be at a lower level.

Tests?

yebrahim: The polymer element contains all of the UI as well as underlying logic. Plus, this logic here is all UI related, it's mostly building the warning message if any and displaying it for large files.

I'm afraid tests will have to come as a separate change.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: I'd like to keep it for clarity. Even thought `showDialog` returns a `BaseDialogCloseResult`, this is a base type, and the actual type returned will differ according to the dialog type passed. So it's just a coincidence that we really are expecting a `BaseDialogCloseResult` in this case, and I think it's worth being explicit about that.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: Fixed and sent in a separate PR.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: Done.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: Done.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: Done.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: Done.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: Done.

PR #1514, opened 2017-07-19
[Polymer UI] Add tslinter

yebrahim:
This PR adds tslinter to the npm manifest, and lints the files by calling the linter manually, it's not yet part of the build process, I'll do this separately. (Should this be required? Should it break the build?)

The most important part of this PR is the lint rules defined in the tslint.json file. These rules are obviously opinionated, so feel free to discuss changes to them.

These are some rules I think might start some discussions:
- `member-access: false`: this is disabled mainly because of the way tyepscript is integrated with Polymer in Datalab. When adding an observer function in Polymer, it's done through a string literal with the function name, so the Typescript parser complains that the function isn't actually used anywhere if it's private. Leaving out the accessors in Polymer elements works because they default to public.
- `no-reference: false`: this is needed again because of the way Typescript is set up for the project. Because we use HTML imports instead a module loader, we can't really use `import` in script, and have to use references instead.
- `max-classes-per-file: false`. Just personal taste.


yebrahim:
SGTM.

PR #1521, opened 2017-07-20
[Polymer UI] Add error popups for file operation failures

yebrahim:
This PR adds an error popup when a file operation fails, using the failure returned from the backend.

image

### DIFF
--- a/sources/web/datalab/polymer/components/base-dialog/base-dialog.html
+++ b/sources/web/datalab/polymer/components/base-dialog/base-dialog.html
@@ -31,6 +31,10 @@ the License.
         min-width: 300px;
         min-height: 200px;
       }
+      .error-title--true {
+        color: #fff;
+        background-color: #ff7777;
+      }
       #body {
         margin: 0px;
         flex-grow: 1;
@@ -59,7 +63,7 @@ the License.
     </style>
     <paper-dialog id="theDialog" class$="dialog-big--{{big}}" with-backdrop
                   on-keydown="_checkEnter">
-      <div class="dialog-title">{{title}}</div>
+      <div class$="dialog-title error-title--{{isError}}">{{title}}</div>
       <!--message value is bound in javascript because it needs to bind to innerHTML-->
       <div id="message" hidden$="{{!messageHtml}}">{{messageHtml}}</div>
       <!--body acts as a place holder for extending elements, their custom
@@ -70,7 +74,9 @@ the License.
         <paper-button id="okButton" on-click="_confirmClose" class="confirm" >
           {{okLabel}}
         </paper-button>
-        <paper-button on-click="_cancelClose">{{cancelLabel}}</paper-button>
+        <paper-button hidden$={{isError}} on-click="_cancelClose">
+          {{cancelLabel}}
+        </paper-button>
       </div>
     </paper-dialog>
   </template>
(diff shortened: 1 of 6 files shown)


sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -562,20 +575,26 @@ class FilesElement extends Polymer.Element {
...
                 this.dispatchEvent(new NotificationEvent(message));
-              });
-              // TODO: [yebrahim] Re-select the renamed item after refresh
+              })
+              .catch((e: Error) => {
+                success = false;
+                Utils.showErrorDialog('Error renaming item', e.message);
+              })
+                // Only refresh the file list on success
+              .then(() => success === true ? this._fetchFileList() : null);
```
yebrahim: The reason I added the success flag is I didn't want `_fetchFileList` to be part of the first `.then`, because if it fails, it'll cause the `.catch` to be called with the wrong error.
I could also return a boolean, but an explicit flag seemed more readable.

sources/web/datalab/polymer/components/base-dialog/base-dialog.html
```diff
@@ -70,7 +74,9 @@
         <paper-button id="okButton" on-click="_confirmClose" class="confirm" >
           {{okLabel}}
```
yebrahim: Lol. Fair enough. I also think "Close" might be a better label. Done.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
jimmc: Another possibility here is to use then(successFunction,errorFunction):
  ApiManager.renameItem()
  .then( ... dispatch event notification ... )
  .then( ()=>fetchFileList, (e)=>showErrorDialog )
That would do what you want using standard flow control, without having to add a flag.

yebrahim: Yes I thought about that as well, but I found it too unclear. I won't remember why I did that when I read this code a few months from now. :)

jimmc: It feels wrong to me to be using a success flag when the Promise mechanism is supposed to have success and failure control flow built in by using then and catch.
What if you add a throw (or return Promise.reject()) at the end of your catch block so that the following then() does not get executed? If you don't catch that elsewhere, does it bubble out as an error somewhere, or will it just be swallowed?
I'll approve, in case you still think the current code is the best way of doing it, but would be happier with a solution that uses then and catch to manage control flow.

yebrahim: It bubbles up as an uncaught exception.

How about this, I'm catching exceptions from fetch and ignoring them explicitly. We can choose to handle them later if we think that's a good idea. Is this a better fix?

jimmc: Yes, I like that better. Since you are doing the fetchFileList-and-catch something like five times, how about adding a method fetchFileListAndIgnoreErrors that encapsulates that additional catch?

yebrahim: Since we're not handling errors from `_fetchFileList`, I'll disable those by default instead.

PR #1522, opened 2017-07-20
[Polymer UI] Refresh on page focus

yebrahim:
This PR changes the way files and sessions pages are refreshed. It reduces the frequency of the auto-refresh interval to a one each minute, and adds a focus handler to trigger the refresh. This makes user operations and page switches the main trigger for refresh, rather than an auto refresh in the background.

Note we still have an issue where multiple tabs with the files page open will separately make refresh calls, so if they're on different paths, the user's last visited path will be one of these at random. The only way to fix this is to add a blur event listener on the window, which is currently not working for me.

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -97,6 +97,8 @@ class DatalabAppElement extends Polymer.Element {
     super.ready();
 
     window.addEventListener('focus', () => this._focusHandler());
+    // TODO: Need to also add a blur event listener that calls the current page's blur
+    // handler, but I cannot get this to work as of now.
   }
 
   /**
@@ -133,7 +135,9 @@ class DatalabAppElement extends Polymer.Element {
     const newElement = this._getPageElement(newPage);
     const oldElement = this._getPageElement(oldPage);
 
-    // Call proper event handlers on changed pages
+    // Call proper event handlers on changed pages.
+    // TODO: Explore making all datalab pages extend a custom element that has
+    // these event handlers defined to keep things consistent and get rid of the checks.
     if (newElement) {
       if (newElement._focusHandler) {
         newElement._focusHandler();
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -93,6 +93,12 @@ class DatalabAppElement extends Polymer.Element {
     ];
   }
 
+  ready() {
+    super.ready();
+
+    window.addEventListener('focus', () => this._focusHandler());
+  }
+
   /**
    * Called when the element is detached from the DOM. Cleans up event listeners.
    */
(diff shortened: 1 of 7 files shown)


sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -117,15 +125,40 @@ class DatalabAppElement extends Polymer.Element {
    * We do this to lazy load pages as the user clicks them instead of letting
    * the browser pre-load all the pages on the first request.
    */
-  _pageChanged(page: string) {
+  _pageChanged(newPage: string, oldPage: string) {
     // Build the path using the page name as suffix for directory
     // and file names.
-    const subpath = 'datalab-' + page;
+    const subpath = 'datalab-' + newPage;
     const resolvedPageUrl = this.resolveUrl('../' + subpath + '/' + subpath + '.html');
```
jimmc: This might read better using string interpolation with a template literal:
  `../${subpath}/${subpath}.html`

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -117,15 +125,40 @@ class DatalabAppElement extends Polymer.Element {
...
-    // If the new page has a resize handler, call it.
-    this._resizeHandler();
+    const newElement = this._getPageElement(newPage);
+    const oldElement = this._getPageElement(oldPage);
+
+    // Call proper event handlers on changed pages.
+    // TODO: Explore making all datalab pages extend a custom element that has
+    // these event handlers defined to keep things consistent and get rid of the checks.
+    if (newElement) {
+      if (newElement._focusHandler) {
```
jimmc: I think code like this inner if-statement is often written as a one-liner:
  newElement._focusHandler && newElement._focusHandler();

sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
```diff
@@ -103,25 +94,31 @@ class SessionsElement extends Polymer.Element {
...
   _fetchSessionList() {
-    const self = this;
-    self._fetching = true;
+    // Don't overlap fetch requests. This can happen because we set up fetch from several sources:
+    // - Initialization in the ready() event handler.
+    // - Refresh mechanism called by the setInterval().
+    // - User clicking refresh button.
+    // - Sessions page gaining focus.
+    if (this._fetching) {
+      return Promise.resolve(null);
```
jimmc: In what scenario would it be useful for the caller to use this return value rather than just ignoring it and continuing its execution?
It might be better not to return a value from this function.
Likewise for fetchFileList

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -93,6 +93,14 @@ class DatalabAppElement extends Polymer.Element {
     ];
   }
 
+  ready() {
+    super.ready();
+
+    window.addEventListener('focus', () => this._focusHandler());
+    // TODO: Need to also add a blur event listener that calls the current page's blur
```
jimmc: Maybe try a focusout event handler?

yebrahim: Not working. It seems to me there's something about Polymer and these events.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: No IE support though, and probably not worth looking for a lib/polyfill.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
yebrahim: The linter screams at this, and I think it's not as readable as explicit if.

sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
yebrahim: This is an async function (perhaps we should revisit naming for all functions to add the `async suffix), so it needs to always return a promise so the caller can wait for it to finish.

jimmc: I am suggesting it may be better for this function not to return a Promise. That's why I asked the question: when does it make sense for a caller to wait on this promise? The function might return right away with a no-op promise, so the caller can't rely on anything having actually happened before continuing, so why ever wait to continue?
Alternatively, set up the promise on line 103 so that it does not resolve until this._fetching gets cleared.

yebrahim: I see. Yea I guess we don't need it to then-able currently.
I'll take it out for now, although we might need that for tests later.

PR #1523, opened 2017-07-20
Add initial BigQuery API access.

jimmc:
This PR implements some basic API calls to BigQuery. At the moment, these are only used from a sample debugging call that gets executed when the Search button is pressed, and which just logs to the console. The next step is to figure out how the user can supply the values that are passed to these functions, and how to display the results.

### DIFF
--- a/sources/web/datalab/polymer/modules/GapiManager.ts
+++ b/sources/web/datalab/polymer/modules/GapiManager.ts
@@ -35,6 +35,8 @@ class GapiManager {
 
   /**
    * Loads the gapi module and the auth2 modules.
+   * @param signInChangedCallback callback to be called when the signed-in state changes
+   * @returns a promise that completes when the load is done or has failed
    */
   public static loadGapi(signInChangedCallback: (signedIn: boolean) => void) {
     // Loads the gapi client library and the auth2 library together for efficiency.
--- a/sources/web/datalab/polymer/modules/GapiManager.ts
+++ b/sources/web/datalab/polymer/modules/GapiManager.ts
@@ -38,7 +38,7 @@ class GapiManager {
    * @param signInChangedCallback callback to be called when the signed-in state changes
    * @returns a promise that completes when the load is done or has failed
    */
-  public static loadGapi(signInChangedCallback: (signedIn: boolean) => void) {
+   public static loadGapi(signInChangedCallback: (signedIn: boolean) => void): Promise<void> {
     // Loads the gapi client library and the auth2 library together for efficiency.
     // Loading the auth2 library is optional here since `gapi.client.init` function will load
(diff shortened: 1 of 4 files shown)


sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -34,7 +40,8 @@ class GapiManager {
     // Loads the gapi client library and the auth2 library together for efficiency.
     // Loading the auth2 library is optional here since `gapi.client.init` function will load
     // it if not already loaded. Loading it upfront can save one network request.
-    GapiManager._loadClientId()
+    console.log('== loading gapi');
+    return GapiManager._loadClientId()
```
yebrahim: Add `Async` to the function `loadGapi` if it's now asynchronous.

third_party/externs/ts/gapi/bigquery.d.ts
```diff
@@ -0,0 +1,85 @@
...
+      projectId: string;
+      all?: boolean;    // true to list all datasets, including hidden ones
+      filter?: string;  // syntax: "labels.<name>[:<value>]"
+            // multiple filters can be ANDed by connecting with a space
+      maxResults?: number  // (unsigned int)
+      pageToken?: string  // page token as returned by a previous call
+    }
+
+    interface ListDatasetsResponse {
+      /*
```
yebrahim: Why are these commented out?

third_party/externs/ts/gapi/gapi.d.ts
```diff
@@ -9,12 +9,24 @@ declare module gapi.client {
...
   }): Promise<void>;
 
+  export function load(name: string, version: string): Promise<void>;
+
   interface RequestOptions {
     path: string;
     params?: any;
   }
 
   export function request(args: RequestOptions): Promise<any>;
```
yebrahim: Where is this used?

sources/web/datalab/polymer/modules/GapiManager.ts
jimmc: Are you proposing that all functions that return a Promise should have a name that ends in Async? I don't think I agree with that (and we have a ton of "Async"s to add if we're going to do it). It sounds kind of like hungarian notation.
Or am I misunderstanding your request?

yebrahim: Yes this was my proposition, I'm also not a huge fan of tying the type system to the naming, but I think it's useful to quickly know the logic flow as I'm calling the function.
How about we stick to an explicit function return type then?

jimmc: I think putting explicit return types on all public async functions is a fine idea. Done (for GapiManager).

PR #1524, opened 2017-07-21
[Polymer UI] Add optional input box to directory picker

yebrahim:
I'll use this for the editor's new file save functionality. We could also choose to use it for copying/moving files with a new name.

Note the screenshot is only to show how it looks like with the feature enabled, but the default directory picker for copying/moving files does not have it on by default.
image

### DIFF
--- a/sources/web/datalab/polymer/components/directory-picker-dialog/directory-picker-dialog.html
+++ b/sources/web/datalab/polymer/components/directory-picker-dialog/directory-picker-dialog.html
@@ -27,10 +27,10 @@ the License.
       .files {
         flex: 1 1 auto;
       }
-      .input {
+      .fileName {
         flex: 0 0 auto;
       }
-      #inputBox {
+      #fileNameBox {
         width: 300px;
         margin: 10px auto;
         /*Disable input underline animation*/
@@ -45,9 +45,9 @@ the License.
       <div class="files">
         <datalab-files id="filePicker" small></datalab-files>
       </div>
-      <div class="input" hidden$={{!withFileName}}>
+      <div class="fileName" hidden$={{!withFileName}}>
         <hr>
-        <paper-input id="inputBox" label="File name" spellcheck=false autofocus
+        <paper-input id="fileNameBox" label="File name" spellcheck=false autofocus
                     value="{{fileName}}" no-label-float>
         </paper-input>
       </div>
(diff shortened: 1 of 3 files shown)


sources/web/datalab/polymer/components/directory-picker-dialog/directory-picker-dialog.ts
```diff
@@ -34,8 +35,32 @@ class DirectoryPickerDialogElement extends BaseDialogElement {
 
   private static _memoizedTemplate: PolymerTemplate;
 
+  /**
+   * Initial value of input box.
+   */
+  public inputValue: string;
```
jimmc: I think it would be better to call this something like fileNameValue, and likewise call the boolean withFileName and update the names of the html fields to be more specific.

sources/web/datalab/polymer/components/directory-picker-dialog/directory-picker-dialog.ts
```diff
@@ -20,6 +20,7 @@
  */
 interface DirectoryPickerDialogCloseResult extends BaseDialogCloseResult {
```
jimmc: Now that you have the ability to pick a file as well, this class name is not quite accurate. Maybe not a big deal for now, but could be nice to change if you come up with a better and still succinct name.

yebrahim: I'd like to keep this name until we actually have a full file picker. Right now, we still can't pick files, as in, you can't really select a file and click a button to confirm, you're still only selecting the current open directory, but now you have an additional input box.

sources/web/datalab/polymer/components/directory-picker-dialog/directory-picker-dialog.ts
yebrahim: Done.

sources/web/datalab/polymer/components/directory-picker-dialog/directory-picker-dialog.html
```diff
@@ -14,13 +14,45 @@
...
+      }
     </style>
-    <div class="files">
-      <datalab-files id="filePicker" small></datalab-files>
+
+    <div class="container">
+      <div class="files">
+        <datalab-files id="filePicker" small></datalab-files>
+      </div>
+      <div class="input" hidden$={{!withFileName}}>
```
jimmc: Were you going to change these names also, to fileName and fileNameBox, or something? Or did you want to leave them as-is? I'll approve in case you have a reason for leaving them as-is, but I'd say it would be better to make those names more precise.

yebrahim: Sure, done.

PR #1525, opened 2017-07-21
[Polymer UI] Save functionality for editor

yebrahim:
For now, this only works with files that open in the editor, not new files.

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
+++ b/sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
@@ -14,6 +14,7 @@
 
 /**
  * Type declarations for CodeMirror
+ * TODO: Get these from DefinitelyTyped
  */
 
 interface CodeMirrorOptions {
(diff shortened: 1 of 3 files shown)


sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
```diff
@@ -116,6 +122,30 @@ class DatalabEditorElement extends Polymer.Element {
...
+        format: this._file.format,
+        mimetype: this._file.mimetype,
+        name: this._file.name,
+        path: dirPath,
+        type: this._file.type,
+      };
+      return ApiManager.saveJupyterFile(model);
+      // TODO: Handle save success/failure here.
+    } else {
+      return Promise.resolve(null);
```
jimmc: Should there be a dialog telling the user of the exp code that this is not yet implemented, or will you be adding that functionality right away?

yebrahim: I think the TODO's are enough for now, I'll work on adding the dialog soon.

PR #1527, opened 2017-07-23
[Polymer UI] <item-list> element tests

yebrahim:
This PR adds tests for the item-list element. It also adds a `package.json` file at the root of the experimental UI directory for npm dependencies. Later on, we can download TS types automatically. For now, this manifest is used for the current build process, and to start tests. To build the front-end, you can navigate to sources/web/datalab/polymer and run `npm run build`. To start tests, you can do `npm test` from the same directory.

Testing the item-list is done in isolation, by feeding rows and columns arrays, and checking various interactions.

One gotcha here is for webcomponent tester to play nice with the Chrome devtools page, the sauce plugin needs to be disabled, otherwise it keeps refreshing the page, which is very annoying for debugging tests. See https://github.com/Polymer/web-component-tester/issues/242#issuecomment-163438535.

### DIFF
--- a/sources/web/datalab/polymer/test/item-list-test.ts
+++ b/sources/web/datalab/polymer/test/item-list-test.ts
@@ -1,0 +1,135 @@
+declare function flush(fn?: () => any): null;
+declare function assert(condition: boolean, message?: string): null;
+declare function fixture(element: string): any;
+
+/// <reference path="../node_modules/@types/mocha/index.d.ts" />
+/// <reference path="../node_modules/@types/chai/index.d.ts" />
+/// <reference path="../components/item-list/item-list.ts" />
+
+describe('<item-list>', () => {
+  let testFixture: ItemListElement;
+
+  function isSelected(i: number) {
+    const row = getRow(i);
+    const box = getCheckbox(i);
+    // Check three things: the index is in the selectedIndices array, the row
+    // has the selected attribute, and its checkbox is checked
+    return testFixture.selectedIndices.indexOf(i) > -1
+        && row.hasAttribute('selected')
+        && box.checked;
+  }
+
+  function getRow(i: number) {
+    return testFixture.$.listContainer.querySelectorAll('paper-item')[i];
+  }
+
+  function getCheckbox(i: number) {
+    const row = getRow(i);
+    return row.querySelector('paper-checkbox');
+  }
+
+  beforeEach(() => {
+    testFixture = fixture('item-list-fixture');
+    const rows: ItemListRow[] = [
+      {
+        firstCol: 'first column 1',
+        icon: 'folder',
+        secondCol: 'second column 1',
+        selected: false,
+      },
+      {
+        firstCol: 'first column 2',
+        icon: 'folder',
+        secondCol: 'second column 2',
+        selected: false,
+      },
+      {
+        firstCol: 'first column 3',
+        icon: 'folder',
+        secondCol: 'second column 3',
+        selected: false,
+      }
+    ];
+    testFixture.rows = rows;
+    flush();
+  });
+
+  it('displays a row per item in list', () => {
+    const children = testFixture.$.listContainer.querySelectorAll('paper-item');
+    assert(children.length === 3, 'three rows should appear');
+  });
+
+  it('selected items are selected', () => {
+    testFixture._selectItem(1);
+    flush();
+
+    assert(!isSelected(0), 'first item should not be selected');
+    assert(isSelected(1), 'second item should be selected');
+    assert(!isSelected(2), 'third item should not be selected');
+  });
+
+  it('selected items are returned', () => {
+    testFixture._selectItem(0);
+    testFixture._selectItem(2);
+    flush();
+
+    assert(JSON.stringify(testFixture.selectedIndices) === '[0,2]',
+        'first and third items should be selected');
+  });
+
+  it('can select/unselect all', () => {
+    const c = testFixture.$.header.querySelector('#selectAllCheckbox') as HTMLElement;
+    c.click();
+    flush();
+
+    assert(JSON.stringify(testFixture.selectedIndices) === '[0,1,2]');
+    assert(isSelected(0), 'all items should be selected');
+    assert(isSelected(1), 'all items should be selected');
+    assert(isSelected(2), 'all items should be selected');
+
+    c.click();
+    flush();
+
+    assert(JSON.stringify(testFixture.selectedIndices) === '[]');
+    assert(!isSelected(0), 'all items should be unselected');
+    assert(!isSelected(1), 'all items should be unselected');
+    assert(!isSelected(2), 'all items should be unselected');
+  });
+
+  it('clicking an item always selects it', () => {
+    const firstRow = getRow(0);
+    firstRow.click();
+    flush();
+
+    assert(isSelected(0), 'first item should be selected');
+
+    firstRow.click();
+    flush();
+
+    assert(isSelected(0), 'first item should still be selected');
+  });
+
+  it('selects only the clicked item if the checkbox was not clicked', () => {
+    const firstRow = getRow(0);
+    const secondRow = getRow(1);
+
+    firstRow.click();
+    secondRow.click();
+    flush();
+
+    assert(isSelected(1), 'only the second item should still be selected');
+  });
+
+  it('can do multi-selection using the checkboxes', () => {
+    const firstCheckbox = getCheckbox(0);
+    const secondCheckbox = getCheckbox(1);
+
+    firstCheckbox.click();
+    flush();
+    assert(isSelected(0), 'first item should be selected');
+
+    secondCheckbox.click();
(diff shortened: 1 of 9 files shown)


sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -0,0 +1,211 @@
...
+  it('displays column names in the header row', () => {
+    testFixture.columns = ['col1', 'col2'];
+    Polymer.dom.flush();
+
+    // Column 0 is for the checkbox
+    assert(testFixture.$.header.children[1].innerText === 'col1');
+    assert(testFixture.$.header.children[2].innerText === 'col2');
+  });
+
+  it('selected items are selected', () => {
```
jimmc: I like my 'it' strings to read as sentences, so they do so in the test output.
it('selects items') would probably be sufficient here.
Likewise for many of the other 'it' strings below.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  it('selected items are returned', () => {
+    testFixture._selectItem(0);
+    testFixture._selectItem(2);
+
+    assert(JSON.stringify(testFixture.selectedIndices) === '[0,2]',
+        'first and third items should be selected');
+  });
+
+  it('can select/unselect all', () => {
```
jimmc: Do you want to test the behavior when some items are selected and the select-all box is clicked?

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -0,0 +1,211 @@
...
+  });
+
+  it('selects only the clicked item if the checkbox was not clicked', () => {
+    const firstRow = getRow(0);
+    const secondRow = getRow(1);
+
+    firstRow.click();
+    secondRow.click();
+
+    assert(isSelected(1), 'only the second item should still be selected');
```
jimmc: I think you also need to assert that (0) is not selected, or that selectedIndices==[1]

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -0,0 +1,211 @@
...
+    testFixture.columns = ['col1', 'col2'];
+    Polymer.dom.flush();
+
+    // Column 0 is for the checkbox
+    assert(testFixture.$.header.children[1].innerText === 'col1');
+    assert(testFixture.$.header.children[2].innerText === 'col2');
+  });
+
+  it('selected items are selected', () => {
+    testFixture._selectItem(1);
```
jimmc: Maybe you should add an assertion that no items are selected before you do anything. Probably a bunch of tests will break if that is ever not the case, but having an explicit test of that might make it quicker to figure that out.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -0,0 +1,211 @@
...
+    const firstCheckbox = getCheckbox(0);
+    const thirdCheckbox = getCheckbox(2);
+
+    firstCheckbox.click();
+    assert(isSelected(0), 'first item should be selected');
+
+    thirdCheckbox.click();
+    assert(isSelected(0) && isSelected(2), 'both first and third items should be selected');
+    assert(!isSelected(1), 'second item should not be selected');
+  });
```
jimmc: Could add an assertion that another click on one of the checkboxes leaves just the other item selected, as you do in the ctrl-click - or is that not how it works here?

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -0,0 +1,211 @@
...
+    }));
+
+    assert(isSelected(0) && isSelected(2), 'both first and third items should be selected');
+    assert(!isSelected(1), 'second item should not be selected');
+
+    firstRow.dispatchEvent(new MouseEvent('click', {
+      ctrlKey: true,
+    }));
+
+    assert(!isSelected(0), 'first row should be unselected on ctrl click');
```
jimmc: Assert the third row is still selected.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -0,0 +1,211 @@
...
+  });
+
+  it('can do multi-selection using the shift key', () => {
```
jimmc: There are some other interesting cases using shift, such as: click(3), ctrl-click(1), shift-click(5). Or click(1), ctrl-click(3), shift-click(5). Those behave differently on different platforms.

yebrahim: Done. Note the test should pass on all platforms. Even though you can't really make a ctrl+click on MacOS, you can still manually dispatch the event in javascript.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -87,27 +97,45 @@ describe('<item-list>', () => {
 
   it('displays a row per item in list', () => {
     const children = testFixture.$.listContainer.querySelectorAll('paper-item');
-    assert(children.length === 3, 'three rows should appear');
+    assert(children.length === 5, 'five rows should appear');
+  });
+
+  it('starts out with all items unselected', () => {
+    assert(!isSelected(0) && !isSelected(1) && !isSelected(2), 'all items should be unselected');
```
jimmc: Now that you have five items, this test is not testing what it says it is. :-)
Probably would be better to assert that selectedIndices == []

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -119,16 +147,50 @@ describe('<item-list>', () => {
...
     assert(!isSelected(0) && !isSelected(1) && !isSelected(2), 'all items should be unselected');
   });
 
-  it('clicking an item always selects it', () => {
+  it('selects all items if the Select All checkbox is clicked with one item selected', () => {
+    testFixture._selectItem(1);
+    const c = testFixture.$.header.querySelector('#selectAllCheckbox') as HTMLElement;
+    c.click();
+
+    assert(isSelected(0) && isSelected(1) && isSelected(2), 'all items should be selected');
```
jimmc: five items should be selected

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -119,16 +147,50 @@ describe('<item-list>', () => {
     const c = testFixture.$.header.querySelector('#selectAllCheckbox') as HTMLElement;
     c.click();
 
-    assert(JSON.stringify(testFixture.selectedIndices) === '[0,1,2]');
+    assert(JSON.stringify(testFixture.selectedIndices) === '[0,1,2,3,4]',
+        'all items should be in the selected indices array');
     assert(isSelected(0) && isSelected(1) && isSelected(2), 'all items should be selected');
```
jimmc: Five items should be unselected. Maybe you should write a couple of helper methods assertAllItemsAreSelected and assertAllItemsAreUnselected.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -54,31 +70,25 @@ describe('<item-list>', () => {
   }
 
   /**
-   * Rows must be recreated on each test with the fixture, to avoid state
-   * leakage. Also be sure to call Polymer's flush() after any code that will
-   * cause shadow dom redistribution, such as observed array mutation, wich is
-   * used by the dom-repeater in this case.
+   * Rows must be recreated on each test with the fixture, to avoid state leakage
```
jimmc: Add a period at the end of this sentence.

sources/web/datalab/polymer/test/item-list-test.ts
jimmc: Right, I meant that the file selection browsers on Mac and Ubuntu behave differently given the above scenario. I don't have a strong opinion as to how ours should behave, but I like having a test so we know if it changes.

sources/web/datalab/polymer/test/item-list-test.ts
yebrahim: I'd like to have at least one place where `isSelected` is called on all elements to maximize its benefit.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -0,0 +1,301 @@
...
+    firstCheckbox.click();
+    assert(isSelected(0), 'first item should be selected');
+
+    thirdCheckbox.click();
+    assert(isSelected(0) && isSelected(2), 'both first and third items should be selected');
+    assert(!isSelected(1) && !isSelected(3) && !isSelected(4), 'the rest should not be selected');
+
+    firstCheckbox.click();
+    assert(!isSelected(0) && !isSelected(1) && isSelected(2) && !isSelected(3) && !isSelected(4),
+        'first item should ben unselected after the second click');
```
jimmc: s/ben/be/

PR #1529, opened 2017-07-24
Update base image. 

qimingj:
- Update Scikit-learn. 
- Add LIME (used for ML feature sensitivity analysis) package.
- Update TF for PY3.


ojarjur:
If we're adding another package, then we should udpate `third_party/license.txt` and `sources/web/datalab/static/datalab.txt` to include it.

qimingj:
Done.

PR #1531, opened 2017-07-27
Start the Datalab container using a systemd service

### LINKED ISSUE #1373: Consider using systemd instead of the kubelet for running the containers in a Datalab VM
Right now, Datalab VMs run the `kubelet` command to run a Kubernetes Pod on the VM that includes the containers we need (the Datalab container and a log-saver container).

However, the Container Optimized OS image that we use also has systemd installed on it, and we may have a cleaner setup if we switch to that.

An example cloud-init script that writes out systemd service configurations is [here](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/master.yaml)

ojarjur:
This commit changes the VMs created by the `datalab` tool
to use systemd services for running their containers rather
than using a kubelet.

This removes one rather large runtime dependency and in my
testing this cuts the VM startup time down by more than one
minute.

This fixes #1373

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -159,76 +159,72 @@ cleanup_tmp
 journalctl -u google-startup-scripts --no-pager > /var/log/startupscript.log
 """
 
-_DATALAB_CONTAINER_MANIFEST_URL = (
-    'http://metadata.google.internal/' +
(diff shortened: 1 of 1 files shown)


nikhilk:
Awesome to see this change... (the removal of large dependency + improvement in startup)

PR #1533, opened 2017-07-27
[Polymer] Data tab: single-column mixed-data format, BQ data

jimmc:
This PR replaces the previous two-column layout with a single-column layout that has one text box for search terms and one list for heterogenous search results. It also wires in the BigQuery API so that the text in the search box is actually used to search for projects, datasets, and tables, which appear in the single results list with different icons. The details area on the right now has a Visibility icon that shows or hides that panel, same as the preview pane in the Files tab.
Turns out the BQ API is not really well suited for this approach, so it might take a while to make it work smoothly. Alternatively, I may try to figure out how to call Datauhb instead and see how far I can get with that.

data-tab-bq-samples

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-data/datalab-data.html
+++ b/sources/web/datalab/polymer/components/datalab-data/datalab-data.html
@@ -15,6 +15,11 @@ the License.
 <link rel="import" href="../../components/item-list/item-list.html">
 <link rel="import" href="../../components/shared-styles/shared-styles.html">
 
+<link rel="import" href="../../bower_components/iron-icon/iron-icon.html">
+<link rel="import" href="../../bower_components/iron-a11y-keys/iron-a11y-keys.html">
+<link rel="import" href="../../bower_components/paper-button/paper-button.html">
+<link rel="import" href="../../bower_components/paper-input/paper-input.html">
+
 <dom-module id="datalab-data">
   <template>
     <style include="datalab-shared-styles">
@@ -29,25 +34,10 @@ the License.
         font-size: var(--app-content-font-size);
         color: var(--app-primary-color);
       }
-      div.tableDrillDown {
-        display: flex;
-        margin: 15px;
-      }
-      .collectionsColumn {
-        width: 33%;
-      }
-      .tablesColumn {
-        width: 33%;
-      }
-      .filteredColumn {
-        margin: 10px;
-      }
-      detailsColumn {
-        width: 33%;
-      }
       #toolbar {
         height: calc(var(--toolbar-height) - 1px); /* for border-bottom */
         width: 100%;
+        display: flex;
         background-color: var(--app-secondary-color);
         border-bottom: 1px solid var(--border-color);
       }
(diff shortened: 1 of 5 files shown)


sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -61,64 +51,61 @@
...
-        <div>
-          Collections Filter:
-            <paper-input id="collectionsFilterBox" label="Optional filter value" spellcheck=false autofocus
-                         value="{{_collectionsFilterValue}}" no-label-float>
-            </paper-input>
+    <div id="data-container">
+      <div class="resultsColumn">
+        <!-- Input box to enter search terms -->
+        <div id="search-div">
+          <iron-a11y-keys id="searchKeys" keys="enter" on-keys-pressed="_search"></iron-a11y-keys>
```
yebrahim: We probably want `_search` to trigger on any changes to the input box, not just keystrokes. Maybe use the `paper-input`'s [value-changed](https://www.webcomponents.org/element/PolymerElements/paper-input/elements/paper-input#event-value-changed)?

yebrahim: Better yet, we can fire after a delay, say 300ms, so that we buffer the user's keystrokes, instead of triggering service calls on each key. I'd add this as a TODO since it's just an improvement.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -18,208 +18,170 @@
...
-  _generateFakeCollectionsListForTesting() {
-    const collectionsList = [];
-    const count = Math.floor(Math.random() * 20);
-    for (let i = 0; i < count; i++) {
-      collectionsList.push(this._generateFakeCollectionForTesting());
-    }
-    return collectionsList;
+  /** Sends the user's query to the search API, renders results as they get returned. */
+  _search() {
+    this._resultsList = [];
```
yebrahim: This could make an unnecessary UI refresh if the list is unchanged. Can we move it to the `_handleQueryResults` where we don't do anything if the results are the same? I assume this will happen a lot as the user is writing a query to narrow down results.
Also, I'm not sure how polished you want this code to be now, so feel free to ignore comments like this, but it'd be good to at least add a TODO to revisit later.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -18,208 +18,170 @@
...
-  _generateFakeTableForTesting() {
-    const fakeTable = {
-      name: 'fakeTable' + Math.floor(Math.random() * 1000000),
-      status: 'fake',
-    };
-    return fakeTable;
+  // Make some calls to the BigQuery API and pass the results to the resultHandler
+  _callBigQuery(searchValue: string, resultHandler: (partialResults: Result[]) => void) {
+    const sampleProject = 'bigquery-public-data';
+    GapiManager.listBigQueryProjects()
```
yebrahim: `searchValue` is not used for searching projects too?

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -18,208 +18,170 @@
...
-    console.log('== collection selected:', collection);
-    if (!collection.tables) {
-      collection.tables = this._generateFakeTablesListForTesting();
-    }
-    this._tablesList = collection.tables;
-  }
-
-  _clearTablesList() {
-    this._tablesList = [];
+  _typeToIcon(type: string): string {
```
yebrahim: If we put each type's icon in its `_<type>ToResult` function we don't have to worry about changes to those types later on.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -18,208 +18,170 @@
...
 
-  _generateFakeTablesListForTesting() {
-    const tablesList = [];
-    const count = Math.floor(Math.random() * 20);
-    for (let i = 0; i < count; i++) {
-      tablesList.push(this._generateFakeTableForTesting());
-    }
-    return tablesList;
+  _handleQueryResults(partialResults: Result[]) {
+    this._resultsList = this._resultsList.concat(partialResults);
```
yebrahim: I'm not sure this is a clean separation between the query operation and displaying results. Some staggering results might spill over after emptying the list, causing results that possible don't match the query (i.e. from the previous query) to appear right?

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
```diff
@@ -61,64 +51,61 @@
...
         display: flex;
       }
-      #collections {
-        flex-grow: 1;
+      #results {
+        width: 100%;
       }
-      #tables {
-        flex-grow: 1;
+      #detailsPane {
```
yebrahim: We'll want some `box-shadow` here.

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
jimmc: I added a TODO.

sources/web/datalab/polymer/components/datalab-data/datalab-data.html
jimmc: Added a TODO to look at this when I make it to working on the preview content.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
jimmc: Right now the search API is pretty bad: I am calling three separate APIs that return paged results, with different kinds of filtering (or none), that are responding at different speeds. Once I figure out how to make that work better - or switch to something completely different - then I can revisit this issue. Meanwhile, I added a TODO.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
jimmc: Yeah, that could happen. I added a TODO to look at that. This is all part of the messy search API that I need to figure out and fix. This code is all pretty much POC at the moment.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
jimmc: No, there is no filtering available in the search-projects call.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
jimmc: Per our offline conversation, leaving this as-is.

PR #1536, opened 2017-07-31
[Data First] Add Table Details Pane

yebrahim:
This PR adds a details pane to the Data First page that shows the selected table's details. One caveat is the schema only shows the columns in the first level, as I haven't yet added flattening (or some other solution) to the schema. This will be coming next.

image

### DIFF
--- a/sources/web/datalab/polymer/components/table-details/table-details.html
+++ b/sources/web/datalab/polymer/components/table-details/table-details.html
@@ -39,7 +39,6 @@ the License.
       }
       table {
         border-spacing: 1px;
-        border-left: 0px solid transparent;
         padding: 0 0 4px 0;
         width: 100%;
       }
@@ -100,6 +99,7 @@ the License.
       <br>
 
       <table>
+        <!--TODO: Flatten out the fields list-->
         <template is="dom-repeat" items={{_table.schema.fields}} as="column">
           <tr>
             <td class="field">{{column.name}}</td>
--- a/sources/web/datalab/polymer/components/table-details/table-details.html
+++ b/sources/web/datalab/polymer/components/table-details/table-details.html
@@ -59,6 +59,7 @@ the License.
       <paper-spinner active hidden$={{!_busy}}></paper-spinner>
     </div>
     <div id="container" hidden$={{!_table}}>
+      <br>
       <h3>Table Details: {{_table.tableReference.tableId}}</h3>
       <span class="strong">Project: </span><span>{{_table.tableReference.projectId}}</span>
       <br>
--- a/sources/web/datalab/polymer/components/table-details/table-details.html
+++ b/sources/web/datalab/polymer/components/table-details/table-details.html
@@ -14,46 +14,101 @@ the License.
 
 <link rel="import" href="../../components/shared-styles/shared-styles.html">
 
+<link rel="import" href="../../bower_components/paper-spinner/paper-spinner.html">
+
 <dom-module id="table-details">
   <template>
     <style include="datalab-shared-styles">
+      #placeholder {
+        height: 100%;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+      }
+      paper-spinner {
+        --paper-spinner-stroke-width: 2px;
+      }
       #container {
         padding: 0px 20px;
         height: 100%;
         overflow-y: auto;
+        color: var(--primary-fg-color);
       }
       .strong {
         font-weight: bold;
       }
+      table {
+        border-spacing: 1px;
+        border-left: 0px solid transparent;
+        padding: 0 0 4px 0;
+        width: 100%;
+      }
+      td.field {
+        padding: 10px;
+        background-color: #eee;
+        font-weight: bold;
+      }
+      td.value {
+        padding: 10px;
+        background-color: #fff;
+      }
+      .separator {
+        margin: 20px 0px;
+      }
     </style>
-    <div id="container" hidden$="{{!table}}">
-      <h3>Table Details: {{table.tableReference.tableId}}</h3>
-      <span class="strong">Project: </span><span>{{table.tableReference.projectId}}</span>
+    <div id="placeholder" hidden$={{_table}}>
+      <paper-spinner active hidden$={{!_busy}}></paper-spinner>
+    </div>
+    <div id="container" hidden$={{!_table}}>
+      <h3>Table Details: {{_table.tableReference.tableId}}</h3>
+      <span class="strong">Project: </span><span>{{_table.tableReference.projectId}}</span>
       <br>
-      <span class="strong">Dataset: </span><span>{{table.tableReference.datasetId}}</span>
+      <br>
+      <span class="strong">Dataset: </span><span>{{_table.tableReference.datasetId}}</span>
+      <br>
       <br>
       <table>
         <tr>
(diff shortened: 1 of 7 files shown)


sources/web/datalab/polymer/components/table-details/table-details.html
```diff
@@ -0,0 +1,116 @@
...
+        </tr>
+      </table>
+
+      <hr class="separator">
+
+      <div class="strong">Schema</div>
+      <br>
+
+      <table>
+        <!--TODO: Flatten out the fields list-->
```
jimmc: Or maybe use an expander for nested fields, especially if there is a top-level type name that you can display for the group of fields.

sources/web/datalab/polymer/components/table-details/table-details.ts
```diff
@@ -0,0 +1,136 @@
...
+    return table ? this._bytesToReadableSize(table.numBytes) : '';
+  }
+
+  /**
+   * Converts the given number of bytes into a human readable string with units
+   * and a two-decimal-point number
+   */
+  _bytesToReadableSize(bytesStr: string) {
+    const kilo = 1024;
+    const units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
```
jimmc: This list of units should last us for a while. :-)

sources/web/datalab/polymer/components/table-details/table-details.ts
```diff
@@ -0,0 +1,136 @@
...
+    let bytes = parseInt(bytesStr, 10);
+    let level = 0;
+
+    while (bytes >= kilo) {
+      bytes /= kilo;
```
jimmc: Once you start dividing, this variable no longer contains the number of bytes. "significand" might be better, or even just "size" since that doesn't imply any units.

third_party/externs/ts/gapi/bigquery.d.ts
```diff
@@ -60,13 +61,48 @@ declare namespace gapi.client {
...
+    }
+
     interface ListTablesResponse {
       kind: string;   // Should always be "bigquery#tableList"
       etag: string;
       nextPageToken: string;
       tables: Array<TableResource>;
     }
 
+    interface BigqueryTable {
```
jimmc: All the symbols in this file are implicitly qualified as BigTable, so I don't think that needs to be included in the interface name. Since this is the response to the GetTableRequest, I think it should be called GetTableResponse.

sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -104,6 +104,19 @@ class GapiManager {
     return GapiManager._loadBigQuery().then(() => gapi.client.bigquery.tables.list(request));
   }
 
+  /**
+   * Fetches table details from BigQuery
+   */
+  public static getTableDetails(projectId: string, datasetId: string, tableId: string):
```
jimmc: I don't know if we want to assume the only tables we have come from BigQuery. How about we call this getBigQueryTable or getBigQueryTableDetails?

sources/web/datalab/polymer/components/table-details/table-details.ts
```diff
@@ -0,0 +1,136 @@
...
+    const matches = this.tableId.match(/^(.*):(.*)\.(.*)$/);
+    if (matches && matches.length === 4) { // The whole string is matched as first result
+      this._busy = true;
+      const projectId = matches[1];
+      const datasetId = matches[2];
+      const tableId = matches[3];
+
+      GapiManager.getTableDetails(projectId, datasetId, tableId)
+        .then((result) => {
+          this._table = JSON.parse(result.body) as gapi.client.bigquery.BigqueryTable;
```
jimmc: Instead of result here you can call it response and declare its type as HttpResponse<gapi.client.bigquery.BigqueryTable> (or GetTableResponse assuming you change the response name to that as suggested elsewhere), then you can directly access response.result as the table details rather than calling JSON.parse.

sources/web/datalab/polymer/components/table-details/table-details.ts
```diff
@@ -0,0 +1,136 @@
...
+        }, (error) => console.log('Failed to get table details: ' + error.body))
```
jimmc: We should start using console.error instead of console.log when logging errors.

sources/web/datalab/polymer/components/table-details/table-details.html
yebrahim: My only worry is the use of space, so I don't want to tackle this decision in this PR, I'd rather think about it later.

sources/web/datalab/polymer/components/table-details/table-details.ts
yebrahim: Done.

sources/web/datalab/polymer/components/table-details/table-details.ts
yebrahim: I had fun reading about where we go after that, but "yottabyte" is the biggest official byte unit-size as far as I can see: https://en.wikipedia.org/wiki/Byte#Unit_symbol :)

sources/web/datalab/polymer/components/table-details/table-details.ts
yebrahim: "Bytes" doesn't imply units either, it's just the number of bytes left from division.
I changed it to "bytesLeft" anyway.

sources/web/datalab/polymer/modules/GapiManager.ts
yebrahim: Done.

third_party/externs/ts/gapi/bigquery.d.ts
yebrahim: Renamed to just "Table".
This object is not the response though, it represents an actual table. A response would be a thin wrapper around it with its `body` property set to this object.

sources/web/datalab/polymer/components/table-details/table-details.ts
yebrahim: Renamed the response object.
The `JSON.parse` is still needed to turn the response body from a string to an object.

sources/web/datalab/polymer/components/table-details/table-details.ts
```diff
@@ -72,41 +72,42 @@ class TableDetailsElement extends Polymer.Element {
       const datasetId = matches[2];
       const tableId = matches[3];
 
-      GapiManager.getTableDetails(projectId, datasetId, tableId)
-        .then((result) => {
-          this._table = JSON.parse(result.body) as gapi.client.bigquery.BigqueryTable;
-        }, (error) => console.log('Failed to get table details: ' + error.body))
+      GapiManager.getBigqueryTableDetails(projectId, datasetId, tableId)
+        .then((response: HttpResponse<gapi.client.bigquery.Table>) => {
+          this._table = JSON.parse(response.body);
```
jimmc: The HttpResponse struct includes a 'result' field that should contain the already-parsed version of 'body'. You should be able to say "this._table = response.result;" here.

yebrahim: Ah, good to know. Done.

PR #1537, opened 2017-08-01
Add <table-details> tests

yebrahim:
This PR adds element test for `<table-details>`. To make this easy I added ids for some of the HTML element accessed by the test.


sources/web/datalab/polymer/test/table-details-test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Rows must be recreated on each test with the fixture, to avoid state leakage.
+   */
+  beforeEach(() => {
+    GapiManager.getBigqueryTableDetails = (pid, did, tid) => {
+      assert(!!pid && !!did && !!tid,
+          'getTableDetails should be called with project, dataset, and table');
+
+      const request = {
```
jimmc: I think this should be 'response' rather than 'request'.

sources/web/datalab/polymer/test/table-details-test.ts
```diff
@@ -0,0 +1,132 @@
...
+        result: mockTable,
+      };
+      return Promise.resolve(request) as any;
```
jimmc: Does this need to be cast to any? If it's going to be cast, seems like it should be cast to HttpResponse<gapi.client.BigQuery.Table>.

sources/web/datalab/polymer/test/table-details-test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      const rows = testFixture.$.schema.rows;
+      mockTable.schema.fields.forEach((field, i) => {
+        assert(rows[i].children[0].innerText === field.name, 'field name not matching');
+        assert(rows[i].children[1].innerText === field.type, 'field type not matching');
+        assert(rows[i].children[2].innerText === (field.mode || 'NULLABLE'),
+            'field mode not matching');
+      });
+
+      done();
```
jimmc: Does the test fail if you don't call done() here?

sources/web/datalab/polymer/test/table-details-test.ts
```diff
@@ -0,0 +1,132 @@
...
+      assert(testFixture.$.datasetId.innerText === mockTable.tableReference.datasetId,
+            'dataset id should be parsed');
+      assert(testFixture.$.tableSize.innerText === '1023.00 KB', 'should see readable table size');
+      assert(testFixture.$.longTermTableSize.innerText === '50.00 MB',
+          'should see readable long term table size');
+      assert(testFixture.$.numRows.innerText === '1,234,567,890',
+          'should see comma-separated number of rows');
+      assert(testFixture.$.creationTime.innerText === '7/31/2017, 3:47:51 PM',
+          'should parse timestamp into readable text');
+      assert(testFixture.$.lastModifiedTime.innerText === '7/31/2017, 3:47:51 PM',
```
jimmc: Would be better if the lastModifiedTime were different from the lastCreationTime.

sources/web/datalab/polymer/test/table-details-test.ts
yebrahim: Done.

sources/web/datalab/polymer/test/table-details-test.ts
yebrahim: Sure, cast to `gapi.client.HttpRequest<gapi.client.bigquery.Table>`.

sources/web/datalab/polymer/test/table-details-test.ts
yebrahim: Done.

sources/web/datalab/polymer/test/table-details-test.ts
yebrahim: It's an async test, so it needs to be waited on. Otherwise any number of side effects can happen. For example, if it does fail in one of the assertions this will probably not get caught because the element has already been removed.

jimmc: Right. I was noticing that all of the asserts are within the callback block, so if for some reason the event never fired, and the callback never got called, no assertions would fail. Thus I just wanted to be sure that the test fails if done() is not called. Please tell me the test fails if you comment out the call to done().

yebrahim: Ah, I misunderstood your first question. Yes the test will timeout if the `done()` call isn't made.

PR #1539, opened 2017-08-01
Make sure gapi is loaded before making GapiManager operations

yebrahim:
This PR changes `GapiManager` to add a dependency on `gapi.load` for all its operations. This way any of its public methods can be called at any point, and it would first make sure the module has loaded. The actual loading happens only once, then its resulting promise is returned every time it's called.

### DIFF
--- a/sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
+++ b/sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
@@ -71,10 +71,11 @@ class AuthPanel extends Polymer.Element {
         .then((email: string) => {
           this._userInfo = 'Signed in as ' + email;
           this._projectInfo = 'No project is set';  // TODO
+
+          const ev = new Event('signInOutDone');
+          this.dispatchEvent(ev);
         });
     }
-    const ev = new Event('signInOutDone');
-    this.dispatchEvent(ev);
   }
 }
 
--- a/sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
+++ b/sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
@@ -47,7 +47,10 @@ class AuthPanel extends Polymer.Element {
 
   ready() {
     super.ready();
-    GapiManager.loadGapi(this._signInChanged.bind(this));
+    GapiManager.listenForSignInChanges(this._signInChanged.bind(this))
+      .catch(() => {
+        // TODO: handle errors authenticating
+      });
   }
 
   _signInClicked() {
@@ -64,8 +67,11 @@ class AuthPanel extends Polymer.Element {
   _signInChanged(signedIn: boolean) {
     this._signedIn = signedIn;
     if (signedIn) {
-      this._userInfo = 'Signed in as ' + GapiManager.getSignedInEmail();
-      this._projectInfo = 'No project is set';  // TODO
+      GapiManager.getSignedInEmail()
+        .then((email: string) => {
+          this._userInfo = 'Signed in as ' + email;
+          this._projectInfo = 'No project is set';  // TODO
+        });
     }
     const ev = new Event('signInOutDone');
     this.dispatchEvent(ev);
--- a/sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
+++ b/sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
@@ -91,6 +91,9 @@ class DataElement extends Polymer.Element {
           console.log('== projects: ', response);
           const projectResults: Result[] = response.result.projects.map(this._bqProjectToResult.bind(this)) as Result[];
           resultHandler(projectResults);
+        })
+        .catch(() => {
+          // TODO: handle errors getting projects
         });
     // The filter arg when querying for datasets must be of the form labels.<name>[:<value>],
     // see https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/list
@@ -99,12 +102,18 @@ class DataElement extends Polymer.Element {
           console.log('== datasets: ', response);
           const datasetResults: Result[] = response.result.datasets.map(this._bqDatasetToResult.bind(this)) as Result[];
           resultHandler(datasetResults);
+        })
+        .catch(() => {
+          // TODO: handle errors getting projects
         });
     GapiManager.listBigQueryTables(sampleProject, searchValue /* datasetId */)
         .then((response: HttpResponse<gapi.client.bigquery.ListTablesResponse>) => {
           console.log('== tables: ', response);
           const tableResults: Result[] = response.result.tables.map(this._bqTableToResult.bind(this)) as Result[];
           resultHandler(tableResults);
+        })
+        .catch(() => {
+          // TODO: handle errors getting projects
         });
   }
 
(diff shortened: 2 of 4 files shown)


sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -32,20 +36,33 @@ class GapiManager {
...
-    return GapiManager._loadClientId()
-      .then(() => gapi.load('client:auth2', this._initClient.bind(this, signInChangedCallback)))
-      .catch((e: Error) => console.log('Failed to get client ID: ', e));
+    if (!this._loadPromise) {
+      this._loadPromise = new Promise((resolve, reject) => {
+        return this._loadClientId()
+          .then(() => gapi.load('client:auth2', resolve))
+          .catch((e: Error) => {
+            if (e instanceof MissingClientIdError) {
+              console.log(e.message);
```
jimmc: console.error

sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -114,50 +143,61 @@ class GapiManager {
...
 
-  private static _initClient(signInChangedCallback: (signedIn: boolean) => void) {
-    // Initialize the client with API key and People API, and initialize OAuth with an
-    // OAuth 2.0 client ID and scopes (space delimited string) to request access.
-    gapi.client.init({
+  public static listenForSignInChanges(signInChangedCallback: (signedIn: boolean) => void):
+      Promise<void> {
+    return this.loadGapi()
+      .then(() => {
+        // If auth status is already know, initialize the callback now
```
jimmc: "already known"

sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -114,50 +143,61 @@ class GapiManager {
...
+        if (gapi && gapi.auth2) {
+          const signedIn = gapi.auth2.getAuthInstance().isSignedIn.get();
+          signInChangedCallback(signedIn);
+        }
+
+        // Listen for auth changes
+        gapi.auth2.getAuthInstance().isSignedIn.listen(() => {
```
jimmc: If line 155 is valid, then this line doesn't look right. If the condition on line 155 returns false, then this line will fail because gapi.auth2 is not defined.

sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -62,46 +79,58 @@ class GapiManager {
...
   /**
    * Signs the user out using gapi.
    */
    public static signOut(): Promise<void> {
-    return gapi.auth2.getAuthInstance().signOut();
+    return this.loadGapi()
+      .then(() => gapi.auth2.getAuthInstance().signOut());
   }
 
   /** Returns the signed-in user's email address. */
```
jimmc: Probably should reword this description, as it no longer actually returns the email address.

sources/web/datalab/polymer/modules/GapiManager.ts
```diff
@@ -114,50 +143,61 @@ class GapiManager {
...
-  private static _initClient(signInChangedCallback: (signedIn: boolean) => void) {
-    // Initialize the client with API key and People API, and initialize OAuth with an
-    // OAuth 2.0 client ID and scopes (space delimited string) to request access.
-    gapi.client.init({
+  public static listenForSignInChanges(signInChangedCallback: (signedIn: boolean) => void):
+      Promise<void> {
+    return this.loadGapi()
+      .then(() => {
+        // If auth status is already know, initialize the callback now
+        if (gapi && gapi.auth2) {
```
jimmc: Once loadGapi succeeds, doesn't that mean the auth load also succeeded, so gapi.auth2 is known to be populated? Under what condition can this expression be false?

sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
```diff
@@ -47,7 +47,10 @@ class AuthPanel extends Polymer.Element {
 
   ready() {
     super.ready();
-    GapiManager.loadGapi(this._signInChanged.bind(this));
+    GapiManager.listenForSignInChanges(this._signInChanged.bind(this))
+      .catch(() => {
+        // TODO: handle errors authenticating
```
jimmc: Do we want to call console.error in all these places while waiting to decide how to handle them?

sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
```diff
@@ -64,8 +67,11 @@ class AuthPanel extends Polymer.Element {
...
-      this._userInfo = 'Signed in as ' + GapiManager.getSignedInEmail();
-      this._projectInfo = 'No project is set';  // TODO
+      GapiManager.getSignedInEmail()
+        .then((email: string) => {
+          this._userInfo = 'Signed in as ' + email;
+          this._projectInfo = 'No project is set';  // TODO
+        });
     }
     const ev = new Event('signInOutDone');
     this.dispatchEvent(ev);
```
jimmc: Should these two lines be inside the then, after line 73? That would make the Account menu close only after the email is retrieved. Might not make any difference, as I think the only path to _signInChanged is after auth is done, which would mean the email is always already loaded, which I think means the promise gets resolved before returning, so the code at line 71 will always run before the code at line 76 anyway. But I'm not positive about that.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -91,6 +91,9 @@ class DataElement extends Polymer.Element {
           console.log('== projects: ', response);
           const projectResults: Result[] = response.result.projects.map(this._bqProjectToResult.bind(this)) as Result[];
           resultHandler(projectResults);
+        })
+        .catch(() => {
+          // TODO: handle errors getting projects
```
jimmc: console.error for now?

sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
yebrahim: I'm not sure we want to do this. So far, having a client ID to make this auth possible is only an option, so it shouldn't be an error.

sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
yebrahim: I think your analysis is correct, but it's good to put code in the right place. Done.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
yebrahim: Same as above discussion.

sources/web/datalab/polymer/modules/GapiManager.ts
yebrahim: Done.

sources/web/datalab/polymer/modules/GapiManager.ts
yebrahim: Done.

sources/web/datalab/polymer/modules/GapiManager.ts
yebrahim: Done.

sources/web/datalab/polymer/modules/GapiManager.ts
yebrahim: Yes, this code used to live outside the promise so the check was necessary. Done.

sources/web/datalab/polymer/modules/GapiManager.ts
yebrahim: Done.

PR #1540, opened 2017-08-02
[Data First] Flatten out table schema

yebrahim:
This PR adds support for nested schema tables in Data First page by flattening out the schema to separate rows. This is similar to the experience on bigquery.cloud.google.com. This is not necessarily the best experience, but it's a good starting point.

The PR also adds a small delay before sending API requests when the user is typing in the search box as a means of buffering.

image

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
+++ b/sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
@@ -104,7 +104,7 @@ class DataElement extends Polymer.Element {
           resultHandler(datasetResults);
         })
         .catch(() => {
-          // TODO: handle errors getting projects
+          // TODO: handle errors getting datasets
         });
     GapiManager.listBigQueryTables(sampleProject, searchValue /* datasetId */)
         .then((response: HttpResponse<gapi.client.bigquery.ListTablesResponse>) => {
@@ -113,7 +113,7 @@ class DataElement extends Polymer.Element {
           resultHandler(tableResults);
         })
         .catch(() => {
-          // TODO: handle errors getting projects
+          // TODO: handle errors getting tables
         });
   }
 
--- a/sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
+++ b/sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
@@ -31,8 +31,6 @@ class DataElement extends Polymer.Element {
 
   private _resultsList: Result[];
   private _searchValue: string;
-  private _searchInputDelayTimeout: number;
-  private readonly _searchInputDelay = 500;
 
   static get is() { return 'datalab-data'; }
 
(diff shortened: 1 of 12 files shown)


sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -31,6 +31,8 @@ class DataElement extends Polymer.Element {
 
   private _resultsList: Result[];
   private _searchValue: string;
+  private _searchInputDelayTimeout: number;
```
jimmc: This change to the way the search field behaves is completely orthogonal to the change for flattening the schema, and non-trivial. Please put in in a separate PR.

sources/web/datalab/polymer/components/table-details/table-details.ts
```diff
@@ -108,6 +115,31 @@ class TableDetailsElement extends Polymer.Element {
...
+    fields.forEach((field) => {
+
+      // First push the record field itself
+      flatFields.push(field);
+
+      // Then flatten it and push its children
+      if (field.type === 'RECORD' && field.fields) {
+        // Make sure we copy the flattened nested fields before modifying their
+        // name to prepend the parent field name
+        const nestedFields = [...this._flattenFields(field.fields)];
```
jimmc: This looks like it recursively flattens, so if the table has something nested a few levels, these names could get pretty long. You know what would be nice to add? A resize bar between the results list and the preview pane so the user could make the preview pane wider in cases like this. In a separate PR, of course.

nikhilk:
Take a look at what nested schemas show within the notebook. We should try to do the same here.

That said, this (https://youtu.be/GrD7ymUPt3M?t=21m27s) is what I've wanted, even within the notebook page -- something that better uses screen real estate for long and/or complex schemas ... rather than showing every field, and in a flattened manner.

sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
yebrahim: Sure.

yebrahim:
This is what we currently have in the notebook, and it's not better in my opinion:
image

We can choose to put a more sophisticated viewer in the expanded view of the table, but for the sake of the details pane I wanted to put something simple in there.

It's not clear from the video what your desired experience looks like, can you explain it a bit more?

nikhilk:
Its better in two ways:
- the most nested field doesn't determine the width of the first column
- the nested section can be expanded/collapsed.

(I suggest trying tables with more complex schema - like the analytics one as used in https://github.com/googledatalab/notebooks/blob/master/samples/Conversion%20Analysis%20with%20Google%20Analytics%20Data.ipynb)

From the screen in the video
- The use of "..." to truncate an overly long schema.
- The nested section shows up in a separate list so as to better use width of available real estate (in case of notebooks).

(Hopefully that is clearer)

screen shot 2017-08-01 at 6 19 19 pm

I think the width/screen space is going to come up here as well -- the sidebar is unsuited for deeply nested schemas, as well as for previews of data... and so IMO, we need to think of showing this in context of results where there is more screen real estate.

yebrahim:
I don't see these points though:
- The most nested field still dictates the width of the table. Without the use of indentation, all fields are flattened to the first level. The only thing that view is adding is just hiding exactly one level of nesting behind an expanded view.
- Expand/collapse doesn't help much here if multiple nested levels can't be collapsed onto any level. This is related to the point above.

I agree we need a neater UX for navigating nested schemas, but I don't think this one is it.

As for the screen in the video, sure we could get this to work, but it won't fit into the details pane obviously, so we can do this in the expanded view when the user really wants to analyze a certain table.

Let me know if we can proceed with what we have here or if you still think we should write a similar viewer to the one in the notebooks.

PR #1541, opened 2017-08-02
[Data First] Styling search box

yebrahim:
image

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
+++ b/sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
@@ -42,7 +42,6 @@ class DataElement extends Polymer.Element {
         value: () => [],
       },
       _searchValue: {
-        observer: '_search',
         type: String,
         value: '',
       },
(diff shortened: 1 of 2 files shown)


sources/web/datalab/polymer/components/datalab-data/datalab-data.ts
```diff
@@ -65,12 +67,25 @@ class DataElement extends Polymer.Element {
...
+      clearInterval(this._searchInputDelayTimeout);
+      this._searchInputDelayTimeout = 0;
+    }
+
+    this._searchInputDelayTimeout = setTimeout(() => {
+      // TODO - clearing the resultsList may cause unnecessary refreshes, clean
+      // this up when we figure out how we actually want to handle the search
+      // call.
+      this._resultsList = [];
+      this._sendQuery(this._searchValue, this._handleQueryResults.bind(this));
```
jimmc: I think you should clear this._searchInputDelayTimeout in this block.

yebrahim:
I actually thought that was the intended behavior of the original code, I did not notice the search icon was actually a button. :)

Let me do that instead and make the button more clear.

yebrahim:
Reusing this PR to revert to the search button and make it more obvious. PTAL.

yebrahim:
PTAL now.

yebrahim:
How about now? :)

PR #1544, opened 2017-08-02
[Polymer UI] Add timeout panel to Account menu

jimmc:
This PR adds the timeout info display to the account menu in the Polymer UI. As in the old UI, clicking on the timer icon toggles whether the idle timeout is enabled. The timeout line is not displayed if the timeout has been disabled by setting the value to 0.

The value displayed in the timeout line counts down as the time until shutdown runs out, but the Polymer code currently is pretty aggressive about requesting info, such as asking for the list of sessions every minute, so in practice the timer will not count down until we fix that issue and make our code not ask for that kind of information when the corresponding tab is not visible.

Display when enabled:
idle-timeout-enabled

Display when disabled by clicking on the timer icon:
idle-timeout-disabled

### DIFF
--- a/sources/web/datalab/polymer/components/timeout-panel/timeout-panel.ts
+++ b/sources/web/datalab/polymer/components/timeout-panel/timeout-panel.ts
@@ -28,11 +28,6 @@
 class TimeoutPanel extends Polymer.Element {
 
   private static _displayInterval = 1000;  // Update display every second
-  private static _hoursPerDay = 24;
-  private static _minutesPerHour = 60;
-  private static _secondsPerMinute = 60;
-  private static _secondsPerHour = TimeoutPanel._secondsPerMinute * TimeoutPanel._minutesPerHour;
-  private static _secondsPerDay = TimeoutPanel._secondsPerHour * TimeoutPanel._hoursPerDay;
   private static _queryIntervalWhenEnabled = 15 * 1000;  // How often to query when timeout is enabled
   private static _queryIntervalWhenDisabled = 60 * 1000;  // How often to query when timeout is disabled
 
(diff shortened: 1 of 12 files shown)


sources/web/datalab/polymer/test/timeout-panel-test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  it('starts with controls not displayed', () => {
+    assert(testFixture.$.timeoutControls.hidden, 'timeoutControls should be hidden');
+  });
+
+  it('shows one icon when opened', (done: () => void) => {
+    TimeoutManager.getTimeout = () => {
+      return Promise.resolve(mockTimeoutInfo);
+    };
+    testFixture.onOpenChange(true).then(() => {
```
yebrahim: One other option for testing this method without returning the promise is to make `_timeoutText` a property with `notify` set to true, and listen for the `_timeout-text-changed` event here, then calling `done()`.

sources/web/datalab/polymer/test/timeout-panel-test.ts
```diff
@@ -0,0 +1,58 @@
...
+      Polymer.dom.flush();
+      assert(!testFixture.$.timeoutEnabledIcon.hidden, 'timeoutEnabledIcon should be showing');
+      assert(testFixture.$.timeoutDisabledIcon.hidden, 'timeoutDisabledIcon should be hidden');
+      done();
+    });
+  });
```
yebrahim: Do we want to add more tests here? For example, making sure the text shows the timer, or testing the second approximation?

jimmc: Added some more tests. Once we figure out what we want to do about the "utility" functions (the stuff that might move elsewhere) I can add more tests of that code when it's not tied to the UI.

sources/web/datalab/polymer/components/timeout-panel/timeout-panel.ts
```diff
@@ -0,0 +1,196 @@
...
+    return TimeoutManager.setTimeoutEnabled(true)
+        .then(() => this._updateTimeoutInfo());
+  }
+
+  private _updateTimeoutInfo(): Promise<void> {
+    console.log('== updateTimeoutInfo');
+    this._lastQueryTime = Date.now();
+    return TimeoutManager.getTimeout()
+        .then((timeoutInfo) => {
+          this._timeoutInfo = timeoutInfo;
```
yebrahim: I see, I'm fine leaving it this way then.

sources/web/datalab/polymer/components/timeout-panel/timeout-panel.ts
```diff
@@ -0,0 +1,196 @@
...
+      this._timeoutText = 'Idle timeout is disabled';
+    }
+
+    this._runTimer();
+  }
+
+  // Rounds seconds to an approximate time, based on magnitude.
+  // We use rounding instead of truncating so that the user will generally see
+  // the number they specified, such as "3h", rather than something like "2h 50m".
+  private _roundToApproximateTime(seconds: number): number {
```
yebrahim: I think that would be better. This seems to be the sort of thing you want to unit test easily instead of have it hidden in the internals of an element.
Running tests on those modules should work the same way, an HTML file sourcing the module and the test script. We could also choose to have one HTML file that sources all the modules and their test scripts.

sources/web/datalab/polymer/components/timeout-panel/timeout-panel.ts
```diff
@@ -0,0 +1,196 @@
...
+    if (seconds <= 120) {
+      return seconds;  // No rounding when less than 2 minutes.
+    }
+    const minutes = Math.round(seconds / 60);
+    if (minutes <= 20) {
+      return minutes * 60; // Round to nearest minute when under 20.5 minutes.
```
yebrahim: Ah I see, the rounding threw me off a little.

sources/web/datalab/polymer/test/timeout-panel-test.ts
yebrahim: No this is fine too.

sources/web/datalab/polymer/components/timeout-panel/timeout-panel.ts
jimmc: Done.

PR #1547, opened 2017-08-03
[Polymer UI] Divider element, use in Data First page

yebrahim:
This adds a vertical divider element with two panes. This element can be dropped in the document with two child divs with slots "left" and "right", and is customizable to provide the initial split percentage, and the minimum pane size.

The element is not visible by default, but the mouse pointer will change to a resize cursor when the user hovers on it, so it's up to the elements using the divider to distinguish the two panes somehow to make the divider visible if needed, for example by adding a right border to the left pane, or a box-shadow to either pane. While it's being dragged, its background color changes slightly.

Future work for this element could be making the divider customizable.

### DIFF
--- a/sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
+++ b/sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
@@ -12,6 +12,8 @@ or implied. See the License for the specific language governing permissions and 
 the License.
 -->
 
+<link rel="import" href="../../bower_components/polymer/polymer.html">
+
 <dom-module id="resizable-divider">
   <template>
 
@@ -30,15 +32,18 @@ the License.
         width: 50%;
         user-select: none;
       }
-      /*
+      /**
        * Slotted elements should fill the content slots, and be border-boxed so
-       * they can add padding
+       * they can add padding.
        */
       ::slotted(*) {
         width: 100%;
         height: 100%;
         box-sizing: border-box;
       }
+      /**
+      * Consider applying a style rule that can be passed to this element.
+      */
       #divider {
         cursor: col-resize;
         user-select: none;
--- a/sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
+++ b/sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
@@ -17,27 +17,30 @@ the License.
 
     <style include="datalab-shared-styles">
       :host {
-        --divider-width: 4px;
+        --divider-width: 5px;
       }
       #container {
         height: 100%;
         width: 100%;
         display: flex;
+        position: relative;
       }
-      .pane, #divider {
+      .pane {
         overflow: auto;
-        width: calc(50% - var(--divider-width));
+        width: 50%;
         user-select: none;
       }
       #divider {
         cursor: col-resize;
-        width: calc(2 * var(--divider-width));
         user-select: none;
-        box-shadow: inset 0px 0px 5px 1px #ccc;
-        z-index: 1;
+        position: absolute;
+        top: 0px;
+        bottom: 0px;
+        left: calc(50% - var(--divider-width));
+        right: calc(50% + var(--divider-width));
+        width: calc(2 * var(--divider-width));
       }
       #divider.active {
-        box-shadow: inset 0px 0px 5px 2px #ccc;
         user-select: none;
       }
     </style>
--- a/sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
+++ b/sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
@@ -25,7 +25,7 @@ the License.
         display: flex;
       }
       .pane, #divider {
-        overflow: hidden;
+        overflow: auto;
         width: calc(50% - var(--divider-width));
         user-select: none;
       }
@@ -45,13 +45,13 @@ the License.
     <div id="container">
 
       <div id="p1" class="pane">
-        Pane1
+        <slot name="left"></slot>
       </div>
 
       <div id="divider" on-mousedown="{{_mouseDownHandler}}"></div>
 
       <div id="p2" class="pane">
-        Pane2
+        <slot name="right"></slot>
       </div>
 
     </div>
--- a/sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
+++ b/sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
@@ -30,6 +30,10 @@ the License.
         width: 50%;
         user-select: none;
       }
+      /*
+       * Slotted elements should fill the content slots, and be border-boxed so
+       * they can add padding
+       */
       ::slotted(*) {
         width: 100%;
         height: 100%;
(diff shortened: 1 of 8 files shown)


sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
```diff
@@ -0,0 +1,80 @@
...
+-->
+
+<link rel="import" href="../../bower_components/polymer/polymer.html">
+
+<dom-module id="resizable-divider">
+  <template>
+
+    <style include="datalab-shared-styles">
+      :host {
+        --divider-width: 5px;
```
jimmc: Given how this is used, I think it should be called divider-half-width.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
```diff
@@ -0,0 +1,80 @@
...
+      }
+      #divider.active {
+        user-select: none;
+        background-color: var(--selection-bg-color);
+      }
+    </style>
+
+    <div id="container">
+
+      <div id="p1" class="pane">
```
jimmc: Should these ids be more descriptive, such as id="left-pane" and "right-pane", or maybe "divider-left-pane" to make id collisions a bit less likely.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  /**
+   * Minimum pane width in pixels. Defaults to 50(px);
+   */
+  public minimumWidthPx: number;
+
+  /**
+   * Initial position of the divider in percentage. Defaults to 50(%);
+   */
+  public initDividerPosition: number;
```
jimmc: I think we can afford to add the extra three letters to spell out "initial" in this variable and its equivalents.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
```diff
@@ -0,0 +1,114 @@
...
+  private _boundMouseMoveHandler: EventListenerOrEventListenerObject;
+  static get is() { return 'resizable-divider'; }
+
+  static get properties() {
+    return {
+      initDividerPosition: {
+        type: Number,
+        value: 50,
+      },
+      minimumWidthPx: {
```
jimmc: Would it be a lot more work to make this just minimumWidth and have it accept either "100px" or "50%" as a value? That would make it more consistent with the other width specifications in CSS.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
```diff
@@ -0,0 +1,114 @@
...
+  public initDividerPosition: number;
+
+  private _boundMouseDownHandler: EventListenerOrEventListenerObject;
+  private _boundMouseupHandler: EventListenerOrEventListenerObject;
+  private _boundMouseMoveHandler: EventListenerOrEventListenerObject;
+  static get is() { return 'resizable-divider'; }
+
+  static get properties() {
+    return {
+      initDividerPosition: {
```
jimmc: If this parameter only accepts a percent, it should probably be called initialDividerPercent to make that clear. Or maybe it should have "Left" in the name somewhere to indicate that this is the size of the left panel (otherwise, we are assuming that everyone will think left-to-right percent). In line with the comment a few lines down, maybe it could be "initialLeftPaneWidth" and accept either "70%" or "300px" as a value. Or allow "initialRightPaneWidth" instead. I'll let you decide how much of that is worth doing right now, but if you decide to minimize the work, I'll stick to my suggestion to at least change the name from "Position" to "Percent".

sources/web/datalab/polymer/test/resizable-divider-test.ts
```diff
@@ -0,0 +1,108 @@
...
+    left = testFixture.querySelector('#left-pane') as HTMLDivElement;
+    right = testFixture.querySelector('#right-pane') as HTMLDivElement;
+  });
+
+  it('inserts the given elements into panes', () => {
+
+    assert(!!left, 'left pane should exist under the divider element');
+    assert(!!right, 'right pane should exist under the divider element');
+
+    assert(left.innerText === 'left pane', 'left pane\'s contents should appear in first slot');
```
jimmc: Should this say "left slot"? I don't see anything that implies an ordering. If you put the left pane second in the html code of the test, it should still work, right?

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
yebrahim: Well what about `pt`, `em`, or `vw`? It's a lot of work to support the entire list, I think we should stick with what fulfills our requirements for now, and add support for things as we go.

The way I see it is you need to be able to specify the size of an element in percent points such that it scales up and down when the window resizes, but also its minimum size in pixels for when you have elements inside it that you don't want to get messed up. I think these two are most important for usability of the element, but if you have a use case in mind that you think we should support we can discuss it.

sources/web/datalab/polymer/test/resizable-divider-test.ts
yebrahim: Not sure what you mean. The `left pane` text identifies the pane not the slot, right?

jimmc: The slots are neither defined nor filled in any particular order, they are identified by their names. Calling it the "first slot", as in this test's error message, is misleading. You could change the order of the definition or use to put the right slot first, and the code would still work, but then the error message would be wrong.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
jimmc: Perhaps I misunderstand how ids work in an element. If this id were "left-pane", would that cause any problems in your test where you also use "left-pane" as an id? If not, it matters less. But I'm not keen on cryptic names. In general, about the only place such names are acceptable is "for i = ..." in a very short loop. Among other problems, very short names are hard to search for. In this case, you are using p1 and p2 in two different files, which is a much larger separation than a short for-loop.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
jimmc: I can live with this for now.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
jimmc: It's hard to argue against YAGNI. I submit to your persuasive argument. :-)

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
yebrahim: Sure, I can make the names clearer just because it's helpful to do that.
If the id were "left-pane" and I inserted a div with the same id as one of the panes, it appears in a different DOM altogether, so you'd reference them differently in the code. The one inside the divider element is in its shadow DOM, and you get a reference to it by calling `this.$.id`, and the one that gets inserted into the slot is in the light DOM, you can get it by calling `this.querySelector('#id')`.
In some sense, you can say these two elements are "different as night and day" :)

PR #1549, opened 2017-08-04
Testing work for <datalab-files> element

yebrahim:
This PR adds some work needed to start testing the <datalab-files> element, it does the following:
- Make the element's `ready` method async, and return a resolved promise if it had already run so it can be called multiple times.
- Remove unneeded `basePath` property. This wasn't actually doing anything, and the base path is being tracked and added to all requests by `ApiManager`.
- Make the `_fetchFileList` method thenable so the test can wait on it.
- Move some common test utilities into one file.
- Add some basic tests for the files element. There are a lot more tests to add here, but this is just to get started.


sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -195,22 +197,23 @@ class FilesElement extends Polymer.Element {
...
-  _fetchFileList(throwOnError = false) {
+  _fetchFileList(throwOnError = false): Promise<any> {
     // Don't overlap fetch requests. This can happen because we set up fetch from several sources:
     // - Initialization in the ready() event handler.
     // - Refresh mechanism called by the setInterval().
     // - User clicking refresh button.
     // - Files page gaining focus.
     if (this._fetching) {
-      return;
+      return Promise.resolve();
```
jimmc: Seems like this should be returning the promise that is currently in flight. Otherwise, the second caller will believe the file list has been fetched and will continue processing, with a null value for the file list.

sources/web/datalab/polymer/test/files-test.html
```diff
@@ -0,0 +1,41 @@
...
+  </head>
+  <body>
+
+    <test-fixture id="files-fixture">
+      <template>
+        <datalab-files></datalab-files>
+      </template>
+    </test-fixture>
+
+    <script src="files-test.js"></script>
```
jimmc: All the other test files so far are named <file-under-test>-test.{html,ts}. Following that pattern, these files should be named datalab-files-test.html and datalab-files-test.ts. Or you could rename datalab-files.html and datalab-files.ts to files.html and files.ts.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: Good point. Done.

sources/web/datalab/polymer/test/files-test.html
yebrahim: Will change everything to the same element name followed by `-test`.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -233,6 +234,8 @@ class FilesElement extends Polymer.Element {
         }
       })
       .then(() => this._fetching = false);
```
jimmc: Might want to clear _fileListFetchPromise here as well. If so, you might be able to remove _fetching and just use !!_fileListFetchPromise instead.

yebrahim: We still need `_fetching` for displaying the progress bar, and it's also used in other places, where using the promise directly won't be very clear. I'll keep using the flag for now.

PR #1564, opened 2017-08-10
Refactor some modules into custom elements

yebrahim:
This PR moves the logic from some pure javascript modules into custom elements. It also adds the abstraction of an ApiManager as a low-level means of sending xhr requests to a given backend. The rest of the ApiManager's operations are going to be moved into a different class, perhaps FileManager, in an upcoming PR.


sources/web/datalab/polymer/modules/api-manager-factory/api-manager-factory.ts
```diff
@@ -0,0 +1,52 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/**
+ * Dependency custom element for ApiManager
+ */
+const API_MANAGER_ELEMENT = {
+  daas: {
+    path: 'not/implemented',
```
jimmc: Might as well fill these in and refer to the nascent DaasApiManager.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -0,0 +1,116 @@
...
+   * Two requests are made to /api/contents and /api/sessions to get this data.
+   * @param path current path to list files under
+   */
+  listFiles(path: string): Promise<DatalabFile[]>;
+
+  /**
+   * Creates a new Datalab item
+   * @param itemType type of the created item
+   */
+  create(itemType: DatalabFileType, path?: string): Promise<DatalabFile>;
```
jimmc: The names are kind of inconsistent here. You have saveFile, create, and deleteItem. I think they should either all have the same suffix (File or Item), or they should all have no suffix (get, save, list, rename, etc).

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -0,0 +1,116 @@
...
+   */
+  deleteItem(path: string): Promise<boolean>;
+
+  /*
+   * Copies an item from source to destination. Item name collisions at the
+   * destination are handled by backend.
+   * @param itemPath path to item to copy
+   * @param destinationDirectory directory to copy the item into
+   */
+  copyItem(itemPath: string, destinationDirectory: string): Promise<DatalabFile>;
```
jimmc: Probably just "path" here to match delete.

sources/web/datalab/polymer/test/settings-manager-test.html
```diff
@@ -20,9 +20,9 @@
     <script src="../../webcomponentsjs/webcomponents-lite.js"></script>
     <script src="../../web-component-tester/browser.js"></script>
 
-    <script src="../modules/SettingsManager.js"></script>
+    <script src="../modules/settings-manager/settings-manager.js"></script>
```
jimmc: Shouldn't this now be a link statement pointing to setting-manager.html? And then not have the script statement on line 26?
Likewise in timeout-manager-test.

sources/web/datalab/polymer/modules/api-manager-factory/api-manager-factory.ts
yebrahim: Done.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
yebrahim: Done. No prefix is simpler.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
yebrahim: Done.

sources/web/datalab/polymer/test/settings-manager-test.html
yebrahim: Yes, fixed. The reason it worked was because it wasn't really a Polymer element.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -71,21 +71,21 @@ interface FileManager {
    *               useful for downloading notebooks, which are by default read
    *               as JSON, which doesn't preserve formatting.
    */
-  getFile(path: string, asText?: boolean): Promise<DatalabFile>;
+  get(path: string, asText?: boolean): Promise<DatalabFile>;
 
   /**
    * Uploads the given file object to the backend. The file's name, path, format,
    * and content are required fields.
    * @param model object containing file information to send to backend
```
jimmc: doc name (model) does not match function param name (file).
Should this return Promise<boolean> like delete, or is there additional information it needs to return?
If so, this could be done in a separate PR.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -104,13 +104,13 @@ interface FileManager {
    * Deletes an item
    * @param path item path to delete
    */
-  deleteItem(path: string): Promise<boolean>;
+  delete(path: string): Promise<boolean>;
 
   /*
    * Copies an item from source to destination. Item name collisions at the
    * destination are handled by backend.
    * @param itemPath path to item to copy
```
jimmc: change docname (itemPath) to match function (path)
As with save, should this return Promise<boolean>?

yebrahim:
I'll address these two comments in the next PR for the sake of moving fast. Merging this one as is.

PR #1565, opened 2017-08-10
Refactoring ApiManager into custom element

yebrahim:
This is a continuation of the Module -> Custom Element refactor work, this time for the ApiManager class.


sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
```diff
@@ -120,6 +121,15 @@ class FilesElement extends Polymer.Element {
...
+    super();
+
+    // TODO: Should choose the FileManager instance dynamically here based on
+    // the data source specified for this datalab-files instance
+    this._fileManager = FileManagerFactory.getInstance();
+    this._apiManager = ApiManagerFactory.getInstance();
+  }
+
   /**
    * Called when when the element's local DOM is ready and initialized We use it
```
jimmc: typo: "when when"

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -42,9 +42,14 @@ interface Notebook {
...
-  FILE,
-  NOTEBOOK,
+  DIRECTORY = 'directory',
+  FILE = 'file',
+  NOTEBOOK = 'notebook',
+}
+
+enum DatalabFileStatus {
+  IDLE = '',
+  RUNNING = 'Running',
```
jimmc: Are you directly using this string to display in the UI? Normally I would say that's a bad idea, but we are so far from having an internationalized UI that I don't know that it matters.

sources/web/datalab/polymer/components/datalab-files/datalab-files.ts
yebrahim: Done.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
yebrahim: Fixed by using a method for now, we'll think about moving that to a strings module.

PR #1568, opened 2017-08-10
More modules to custom elements

yebrahim:
- Converting GapiManager and TemplateManager to custom elements.
- Renaming datalab-files to file-browser.


sources/web/datalab/polymer/components/datalab-app/datalab-app.html
```diff
@@ -76,7 +76,7 @@
           <iron-pages id="pages" selected="{{page}}" attr-for-selected="name"
                       fallback-selection="files">
             <datalab-data class="page" name="data"></datalab-data>
-            <datalab-files class="page" name="files"></datalab-files>
+            <file-browser class="page" name="files"><file-browser>
```
jimmc: Needs a slash in the closing tag

sources/web/datalab/polymer/test/table-details-test.html
```diff
@@ -0,0 +1,39 @@
...
+  </head>
+  <body>
+
+    <test-fixture id="table-details-fixture">
+      <template>
+        <table-details></table-details>
+      </template>
+    </test-fixture>
+
+    <script src="table-details-test.js"></script>
```
jimmc: I don't see a table-details-test.js file, and table-details.test does not appear in index.html. I think either those two should be fixed, or we should just delete this file for now until we actually have those tests.

yebrahim: This file was added by mistake. Removed.

sources/web/datalab/polymer/components/datalab-app/datalab-app.html
yebrahim: Done.

PR #1569, opened 2017-08-11
Refactor gapi-manager to request scopes individually

yebrahim:
This changes the gapi-manager element to expose a public method to request granting individual scopes. The element is initialized with a set of "basic" scopes that do not require the user's approval. The datalab-data element then requests the BigQuery scope on initialization.


sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -13,31 +13,47 @@
...
-      'https://www.googleapis.com/auth/bigquery';
-
   private static _clientId = '';   // Gets set by _loadClientId()
   private static _loadPromise: Promise<void>;
 
   /**
+   * Request a new scope to be granted.
+   */
+  public static async grantScope(scope: GapiScopes): Promise<any> {
+    await this.loadGapi();
```
jimmc: If I understand this flow correctly, if this is the very first gapi call, when loadGapi calls initClient, that asks the user for auth on profile and email, and then when this method calls signIn on line 47, that immediately subsequently asks the user for auth on whatever was passed in here. Is that right? If so, maybe you can pass the currently requested scope to loadGapi so that, if it is the first time and it calls initClient, it can pass the requested scope to initClient and have it ask for all of them at the same time.

yebrahim: Pages that require a specific scope should call the `grantScope` method, which will ask the user to sign in with this scope if they're signed out. The basic profile requested in the separate Sign In operation does not require the user to approve anything.

PR #1570, opened 2017-08-11
Begin implementation of new Data browser, also BigQueryFileManager.

jimmc:
This PR starts the implementation of a new data browser to replace the current Data tab. This implementation uses the file-browser element along with the new BigQueryFileManager.

Since this does not yet implement the preview capability of the current Data tab, I am not yet replacing the current Data tab, but rather adding a new "/data2" path that links to this new data browser. Once it has the preview capability of the current Data tab, we can move it to /data and remove the old Data tab implementation.

Still TODO: remove the Status column, implement preview for tables, add a filter input field, display appropriate icons.

This example shows a project (jimmc-datalab) and a BigQuery dataset (test) with the two BigQuery tables in that dataset.
data2-tab

### DIFF
--- a/sources/web/datalab/polymer/components/data-browser/data-browser.ts
+++ b/sources/web/datalab/polymer/components/data-browser/data-browser.ts
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/// <reference path="../input-dialog/input-dialog.ts" />
+/// <reference path="../item-list/item-list.ts" />
+
+/**
+ * Data Browser element for Datalab.
+ */
+class DataBrowserElement extends Polymer.Element {
+
+  static get is() { return 'data-browser'; }
+
+  static get properties() {
+    return {
+    };
+  }
+
+  ready() {
+    super.ready();
+  }
+}
+
+customElements.define(DataBrowserElement.is, DataBrowserElement);
--- a/sources/web/datalab/polymer/components/data-browser/data-browser.ts
+++ b/sources/web/datalab/polymer/components/data-browser/data-browser.ts
@@ -22,14 +22,6 @@ class DataBrowserElement extends Polymer.Element {
 
   static get is() { return 'data-browser'; }
 
-  static get properties() {
-    return {
-    };
-  }
-
-  ready() {
-    super.ready();
-  }
 }
 
 customElements.define(DataBrowserElement.is, DataBrowserElement);
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -139,15 +139,8 @@ class FileBrowserElement extends Polymer.Element {
     };
   }
 
-  constructor() {
-    super();
-
-    this._apiManager = ApiManagerFactory.getInstance();
-  }
-
   /**
-   * Called when the element's local DOM is ready and initialized We use it to
-   * initialize element state.
+   * Called when the element's local DOM is ready and initialized.
    */
   async ready() {
     // Must set this to true before calling super.ready(), because the latter will cause
@@ -157,18 +150,16 @@ class FileBrowserElement extends Polymer.Element {
 
     super.ready();
 
+    this._apiManager = ApiManagerFactory.getInstance();
+
     if (!this.fileManagerType) {
+      // TODO - This element should nto have to set a default file manager type, but
+      // it is currently required to know that we should load the startuppath.
       this.fileManagerType = 'jupyter';
-    }
-    switch (this.fileManagerType) {
-      case 'bigquery':
-        this._fileManager =
-            FileManagerFactory.getInstanceForType(FileManagerType.BIG_QUERY);
-        break;
-      default:
-      case 'jupyter':
-        this._fileManager = FileManagerFactory.getInstance();
-        break;
+      this._fileManager = FileManagerFactory.getInstance();
+    } else {
+      this._fileManager = FileManagerFactory.getInstanceForType(
+          FileManagerFactory.fileManagerNameToType(this.fileManagerType));
     }
 
     // TODO: Using a ready promise might be common enough a need that we should
@@ -842,6 +833,8 @@ class FileBrowserElement extends Polymer.Element {
   }
 
   private _loadStartupPath() {
+    // TODO - move this to SettingsManager and make it able to store startuppaths
+    // for multiple file managers.
     if (this.fileManagerType === 'jupyter') {
       return SettingsManager.getUserSettingsAsync(true /*forceRefresh*/)
         .then((settings: common.UserSettings) => {
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -47,6 +47,16 @@ class FileBrowserElement extends Polymer.Element {
   public currentPath: string;
 
   /**
+   * The type of FileManager we want to use for this file-browser.
+   */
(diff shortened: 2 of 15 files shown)


sources/web/datalab/polymer/components/data-browser/data-browser.ts
```diff
@@ -0,0 +1,35 @@
...
+/// <reference path="../item-list/item-list.ts" />
+
+/**
+ * Data Browser element for Datalab.
+ */
+class DataBrowserElement extends Polymer.Element {
+
+  static get is() { return 'data-browser'; }
+
+  static get properties() {
```
yebrahim: You can remove both properties and ready.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -124,9 +142,6 @@ class FileBrowserElement extends Polymer.Element {
   constructor() {
     super();
 
-    // TODO: Should choose the FileManager instance dynamically here based on
-    // the data source specified for this file-browser instance
-    this._fileManager = FileManagerFactory.getInstance();
     this._apiManager = ApiManagerFactory.getInstance();
```
yebrahim: Let's move this one too to keep things consistent. If dependencies are built in the ready method let's do away with the constructor.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -47,6 +47,16 @@ class FileBrowserElement extends Polymer.Element {
   public currentPath: string;
 
   /**
+   * The type of FileManager we want to use for this file-browser.
```
yebrahim: Can we document the different values for the attribute, or point to the factory file maybe?

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -142,42 +157,26 @@ class FileBrowserElement extends Polymer.Element {
...
 
+    if (!this.fileManagerType) {
+      this.fileManagerType = 'jupyter';
+    }
+    switch (this.fileManagerType) {
+      case 'bigquery':
+        this._fileManager =
+            FileManagerFactory.getInstanceForType(FileManagerType.BIG_QUERY);
+        break;
+      default:
```
yebrahim: Let's throw an error here. If the attribute is not specified it gets populated up on line 160, so this catches bad values.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -840,6 +841,45 @@ class FileBrowserElement extends Polymer.Element {
     }
   }
 
+  private _loadStartupPath() {
+    if (this.fileManagerType === 'jupyter') {
```
yebrahim: I think this should be the SettingsManager's responsibility, the file browser shouldn't know where the user settings are stored. So let's keep the old code as is, and just add a TODO that the SettingsManager needs to multiplex?

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+}
+
+/**
+ * A file manager that wraps the BigQuery API so that we can browse BQ projects,
+ * datasets, and tables like a filesystem.
+ */
+class BigQueryFileManager implements FileManager {
+
+  /**
```
yebrahim: Can we remove these comments? They're not related to BigQuery. And I think we should keep these comments on the interface only anyway, we can add BigQueryFileManager related comments inside the methods.

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -0,0 +1,165 @@
...
+   * Returns a DatalabFile object representing the file or directory requested
+   * @param path string path to requested file
+   * @param asText whether the file should be downloaded as plain text. This is
+   *               useful for downloading notebooks, which are by default read
+   *               as JSON, which doesn't preserve formatting.
+   */
+  public async get(path: string, asText?: boolean): Promise<DatalabFile> {
+    path; asText; // Make compiler happy until we implement this function.
```
yebrahim: underscores work here?

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /**
+   * Returns a list of files at the target path, each implementing the
+   * DatalabFile interface. Two requests are made to /api/contents and
+   * /api/sessions to get this data.
+   * @param path current path to list files under
+   */
+  public list(path: string): Promise<DatalabFile[]> {
+    const pathParts = path.split('/').filter((part) => !!part);
```
yebrahim: Please add a comment about why this works, and our assumption that slashes are never in any of the parts.

sources/web/datalab/polymer/modules/file-manager-factory/file-manager-factory.ts
```diff
@@ -16,34 +16,54 @@
...
+  },
   jupyter: {
+    name: 'jupyter',
     path: 'modules/jupyter-file-manager/jupyter-file-manager.html',
     type: JupyterFileManager,
   },
 };
 
+enum FileManagerType {
+  BIG_QUERY,  // bigquery
```
yebrahim: Self explanatory? :)

sources/web/datalab/polymer/modules/file-manager-factory/file-manager-factory.ts
```diff
@@ -16,34 +16,54 @@
...
-    if (!FileManagerFactory._fileManager) {
-      const backendType = FileManagerFactory._getBackendType();
+    return FileManagerFactory.getInstanceForType(FileManagerType.JUPYTER);
+  }
+
+  public static getInstanceForType(fileManagerType: FileManagerType) {
+    const backendType = FileManagerFactory._getBackendType(fileManagerType);
+    if (!FileManagerFactory._fileManagers[backendType.name]) {
 
       Polymer.importHref(backendType.path, undefined, undefined, true);
```
yebrahim: Let's take this one out for now, and worry about it later if it's causing slow downs. It shouldn't be here if we have the paths imported in the html.

sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -246,7 +251,8 @@ class GapiManager {
 
   private static _loadBigQuery(): Promise<void> {
     return this.loadGapi()
-      .then(() => gapi.client.load('bigquery', 'v2'));
+      .then(() => gapi.client.load('bigquery', 'v2'))
```
yebrahim: I'm curious if we really need this line. Doesn't it work without it?

sources/web/datalab/polymer/test/file-browser-test.ts
```diff
@@ -37,10 +37,13 @@ describe('<file-browser>', () => {
     ApiManagerFactory.getInstance().getBasePath = () => {
       return Promise.resolve('');
     };
-    FileManagerFactory.getInstance().list = (path: string) => {
-      assert(path === startuppath, 'listFilesAsync should be called with the startup path');
-      return Promise.resolve(mockFiles);
-    };
+    const mockFileManager = {
```
yebrahim: It's interesting we need to do this, I can't see the difference, do you?

jimmc: By doing this, the test no longer calls any of the real code in FileManagerFactory.getInstance. The way it was previously done, it called getInstance(), which instantiated a jupyter-file-manager, then overwrote a method in that file manager.

sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
jimmc: It is necessary. Without this line, gapi.client.bigquery does not get filled, so calls such as to gapi.client.bigquery.projects.list fail.

nikhilk:
Should have probably chimed in earlier on this...

Can we consider couple of renames?

1. BigQuery -> Data
2. File -> Content (something generic that lends itself to non-File scenarios like data in fact)

jimmc:
The BigQuery name is used on the API module that access the BigQuery API, so it seems appropriate to me. I expect at some point to have a Datahub API module, which will be able to return similar data via the Datahub API.

I'm not opposed to doing some kind of rename from File to Content or maybe Document, but that's mostly unrelated to this set of changes, so we can do that later. We will probably be doing some more refactoring of how we handle files, so we could do it as part of that.

PR #1571, opened 2017-08-13
Breadcrumbs as an element

yebrahim:
Breadcrumbs were getting too complex to keep inside the file browser element. Moving them into a separate element allows for more logic, such as hiding overflowing links to the left.

image
image

### DIFF
--- a/sources/web/datalab/polymer/components/bread-crumbs/bread-crumbs.ts
+++ b/sources/web/datalab/polymer/components/bread-crumbs/bread-crumbs.ts
@@ -46,10 +46,17 @@ class BreadCrumbsElement extends Polymer.Element {
   }
 
   _resizeHandler() {
-    const container = this.$.breadcrumb as HTMLDivElement;
-    const maxWidth = container.offsetWidth - 30;
+    // Let the items render first, before hiding overflowing items
+    Polymer.dom.flush();
 
-    const children = container.querySelectorAll('div.crumb') as NodeListOf<HTMLDivElement>;
+    const container = this.$.breadcrumb as HTMLDivElement;
+    const maxWidth = container.offsetWidth - 55;
+
+    const children = container.querySelectorAll('div.part') as NodeListOf<HTMLDivElement>;
+
+    if (!children.length) {
+      return;
+    }
 
     // Find the index of the last breadcrumb we can fit, starting from the right
     // We must fit at least the right-most child, so this is the starting point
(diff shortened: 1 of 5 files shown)


sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -142,6 +137,17 @@ class FileBrowserElement extends Polymer.Element {
 
     super.ready();
 
+    this.$.breadCrumbs.addEventListener('crumbClicked', (e: CustomEvent) => {
```
jimmc: In bread-crumbs.ts you dispatch an ItemClickedEvent, should that be the type of e here?

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -251,22 +258,27 @@ class FileBrowserElement extends Polymer.Element {
...
     }
 
-    this._currentCrumbs = this.currentPath.split('/');
+    this.$.breadCrumbs.crumbs = this._splitCurrentPath();
+
+    return this._fetchFileList();
+  }
+
+  _splitCurrentPath() {
+    const pathTokens = this.currentPath.split('/');
```
jimmc: You could add .filter((p) => p) here (or !!p, or p!='') instead of the if-statement on line 278

sources/web/datalab/polymer/components/bread-crumbs/bread-crumbs.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    // Find the index of the last breadcrumb we can fit, starting from the right
+    // We must fit at least the right-most child, so this is the starting point.
+    this.lastVisibleIndex = children.length - 1;
+    children[this.lastVisibleIndex].classList.remove('hidden');
+    let runningWidth = children[this.lastVisibleIndex].offsetWidth;
+
+    for (let i = this.lastVisibleIndex - 1; i >= 0; --i) {
+      if (runningWidth + children[i].offsetWidth < maxWidth) {
```
yebrahim: I thought this would make the four loop's definition a little longer than usual, but I'm willing to give it a try if you think it's more readable.

Second solution wouldn't work because we need the loop to terminate after finding the first element that won't fit. Otherwise it'll hide elements in the middle of the path and show earlier elements if they're narrower.

jimmc: If you move the calculation of runningWidth to be the first thing in the loop rather than within the if statement, then I think that part works, as once it becomes greater than maxWidth, it will remain that way for the rest of the loop. But this way is also OK if you prefer it.

yebrahim:
Thanks for catching that. Had to rebase unfortunately, but things should look right now.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -162,6 +168,17 @@ class FileBrowserElement extends Polymer.Element {
           FileManagerFactory.fileManagerNameToType(this.fileManagerType));
     }
 
+    this.$.breadCrumbs.addEventListener('crumbClicked', (e: CustomEvent) => {
```
jimmc: ItemClickEvent?

PR #1574, opened 2017-08-14
Hide the Terminal button.

jimmc:


jimmc:
Yes, that's a better way. Done.

PR #1575, opened 2017-08-14
Add down-arrow by Kernel choice, show ellipsis when it is changing.

jimmc:
This PR adds a caret to the right of the kernel name in the notebook tool bar to indicate that there is a drop-down choicelist:
kernel-caret

When the user selects an item from that choicelist to switch kernels, the kernel name is replaced by "..." until the new kernel connects.


qimingj:
Oh. I did not see that. I think it is sufficient.

PR #1580, opened 2017-08-16
FileManager refactor to allow for using multiple file services

yebrahim:
This PR refactors a few of the classes that we have so that we can work with multiple file services.

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
+++ b/sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
@@ -61,10 +61,6 @@ class DatalabEditorElement extends Polymer.Element {
   async ready() {
     super.ready();
 
-    // TODO: This should get the FileManager instance that corresponds to the
-    // origin of the file opened in the editor
-    this._fileManager = FileManagerFactory.getInstanceForType(this.fileId.source);
-
     // Get the theme.
     const settings = await SettingsManager.getUserSettingsAsync()
       .catch(() => console.error('Could not load user settings.'));
@@ -76,6 +72,10 @@ class DatalabEditorElement extends Polymer.Element {
     // Get the file contents, or empty string if no path is specified
     let content = '';
     if (this.fileId) {
+      // TODO: This should get the FileManager instance that corresponds to the
+      // origin of the file opened in the editor
+      this._fileManager = FileManagerFactory.getInstanceForType(this.fileId.source);
+
       this._busy = true;
       // Passing the asText=true parameter guarantees the returned type is not a directory.
       // An error is thrown if it is.
--- a/sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
+++ b/sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
@@ -69,37 +69,40 @@ class DatalabEditorElement extends Polymer.Element {
       this._theme = settings.theme;
     }
 
-    // Get the file contents, or empty string if no path is specified
+    // Get the file contents, or empty string if no path is specified or the
+    // file could not be found.
     let content = '';
     if (this.fileId) {
-      // TODO: This should get the FileManager instance that corresponds to the
-      // origin of the file opened in the editor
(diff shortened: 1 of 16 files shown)


sources/web/datalab/polymer/components/datalab-editor/datalab-editor.ts
```diff
@@ -68,28 +64,34 @@ class DatalabEditorElement extends Polymer.Element {
...
       this._busy = false;
-
-      if (this._file) {
-        content = this._file.content as string;
-      }
+    } else {
+      // TODO: Make this more flexible instead of assuming the default
+      // destination is jupyter. If there's no fileId specified (blank editor),
+      // we should ask the user on save about the file destination.
+      this._fileManager = FileManagerFactory.getInstanceForType(FileManagerType.JUPYTER);
```
jimmc: Something we can consider for a future PR is to implement some kind of "null" FileManager that is not tied to anything. Then when the user asks to Save, we can figure out where the user wants to save, and replace this._fileManager with the appropriate FileManager for that destination.

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -18,28 +18,43 @@ class BigQueryMethodNotYetImplemented extends Error {
   }
 }
 
+class BigQueryFile extends DatalabFile {
+}
+
+class BigQueryFileId extends DatalabFileId {
```
jimmc: DatalabFileId now has a path field, so I believe we don't want this subclass.

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -18,28 +18,43 @@ class BigQueryMethodNotYetImplemented extends Error {
...
+  public async getRootFile() {
+    return this.get(new DatalabFileId('/', FileManagerType.BIG_QUERY));
+  }
+
+  public saveText(_file: DatalabFile, _content: string): Promise<DatalabFile> {
     throw new BigQueryMethodNotYetImplemented('save');
   }
 
-  public list(path: string): Promise<DatalabFile[]> {
+  public list(containerId: BigQueryFileId): Promise<DatalabFile[]> {
```
jimmc: DatalabFileId

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -53,70 +43,169 @@ enum DatalabFileStatus {
...
+  source: FileManagerType;
+
+  constructor(path: string, source: FileManagerType) {
+    this.path = path;
+    this.source = source;
+  }
+
+  public static fromQueryString(querystring: string) {
+    const tokens = querystring.split(DatalabFileId._delim);
+    return new DatalabFileId(tokens[1], FileManagerFactory.fileManagerNameToType(tokens[0]));
```
jimmc: Probably should verify tokens.length==2

sources/web/datalab/polymer/modules/jupyter-file-manager/jupyter-file-manager.ts
```diff
@@ -18,64 +18,174 @@
  * these APIs to help with type checking.
  */
 
+class JupyterFile extends DatalabFile {
```
jimmc: It might also be worth defining an interface type for the upstream jupyter file. I'm fine with postponing that to a future PR.

sources/web/datalab/polymer/test/file-browser-test.ts
```diff
@@ -12,14 +12,73 @@
  * the License.
  */
 
+class MockFileManager implements FileManager {
+  public get(_fileId: DatalabFileId): Promise<DatalabFile> {
+    throw new BigQueryMethodNotYetImplemented('get');
```
jimmc: Probably should not be using a BigQuery class in this test file.

yebrahim: Moved to a general purpose error.

PR #1594, opened 2017-08-18
Make loadUserSettings more robust when file is corrupt.

### LINKED ISSUE #1592: Datalab failed to open localhost - idleTimeout
I pulled the latest version of Datalab image ( Datalab 1.2.20170818) and use the same command to run datalab, however, it keeps throwing the following errors and I can't open datalab in my localhost.

`docker run -it -e DATALAB_DEBUG=true --name=datalab -p "127.0.0.1:8081:8080" -v "${HOME}/git/datalab:/content" -e "PROJECT_ID=my-project" gcr.io/cloud-datalab/datalab:local`

image

Still trying to get it working. Any suggestions? 
thank you

jimmc:
This fixes #1592 by returning an empty UserSettings object rather than null when the settings file is corrupt. It also changes the logging so that the message about an error in the settings file is logged at Error rather than at Debug.
This PR does not address or identify the cause of the file corruption.


yebrahim:
This would cause the user's settings to be lost though, because consecutive calls to save will end up saving an empty settings dict. Can we at maybe backup the file?

jimmc:
Added code to rename the file so it doesn't get lost.

sources/web/datalab/settings.ts
```diff
@@ -396,6 +399,21 @@ export function createHandler(): http.RequestHandler {
   return requestHandler;
 }
 
+function renameBadUserSettings(settingsPath: string) {
+  let newPath = settingsPath + ".bad";
+  let n = 0;
+  while (fs.existsSync(newPath)) {
```
yebrahim: Do we need a limit on this? Since we still don't know what's causing the bug, if this happens frequently enough we'll fill up the file system with these files.
Maybe we can just use one -bad name and override it?

jimmc: I'm not too worried about filling up the disk, these are small files. I'm more concerned about losing the user's settings file if we always move to the same file. I put in a limit of ten backups.

yebrahim: SGTM.

PR #1599, opened 2017-08-23
Drive double click actions

yebrahim:
Double clicking an item from a drive file browser now executes the right action.
There's still some cleanup to do here; obtaining the editor/notebook urls can be extracted outside of the individual file managers, since most of it is repeated logic.

Note this won't work on the Jupyter-based notebook editor as of now, since it uses the new notebook fileId path, which needs to be supported by the notebook editor.

### DIFF
--- a/sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
+++ b/sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
@@ -113,7 +113,7 @@ class DriveFileManager implements FileManager {
     // Notebooks that are stored on the VM require the basepath.
     const apiManager = ApiManagerFactory.getInstance();
     const basepath = await apiManager.getBasePath();
-		return location.protocol + '//' + location.host + basepath +
+    return location.protocol + '//' + location.host + basepath +
         '/notebook#fileId=' + fileId.toQueryString();
   }
 }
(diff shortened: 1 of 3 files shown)


sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
```diff
@@ -98,11 +101,19 @@ class DriveFileManager implements FileManager {
...
+        '/editor?file=' + fileId.toQueryString();
   }
 
-  public getNotebookUrl(): Promise<string> {
-    throw new UnsupportedMethod('getNotebookUrl', this);
+  public async getNotebookUrl(fileId: DatalabFileId): Promise<string> {
+    // Notebooks that are stored on the VM require the basepath.
+    const apiManager = ApiManagerFactory.getInstance();
+    const basepath = await apiManager.getBasePath();
+		return location.protocol + '//' + location.host + basepath +
```
jimmc: fix indentation

sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
```diff
@@ -98,11 +101,19 @@ class DriveFileManager implements FileManager {
...
+        '/notebook#fileId=' + fileId.toQueryString();
```
jimmc: Why is this using # rather than ? as in getEditorUrl?

sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
yebrahim: Done.

sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
yebrahim: Discussed offline, leaving as is for now.

PR #1605, opened 2017-08-25
Fix showing line numbers on notebook init

### LINKED ISSUE #1604: Show line numbers by default when re-opening notebook
When creating a new cell in a Python notebook (e.g. by clicking the "Add Code" button or typing `a` or `b` in command mode) the new cell has line numbers showing by default.  (I find this helpful, because I click on the line number margin to toggle between command mode and edit mode; without the line number margin, mode change is not possible with mouse, only via the Escape key.)

However, if I save the notebook, close it, and re-open it, all line numbers disappear.  If I want them back I'd have to use the [undocumented](https://github.com/googledatalab/datalab/issues/1532) keyboard shortcut to toggle line numbers on every single cell, which is impractical; I know of no way to toggle them for all cells.

So, I'd like to simply request line numbers are enabled on all code cells upon opening a notebook. (If for some reason this is untenable as default behavior, then my secondary ask would be to have a button or shortcut to turn line numbers on for all code cells.)

yebrahim:
The codecell custom configurations were incorrectly placed in the notebook app's postload section, which was causing existing cells to load without line numbers (and other customizations), but new cells to load with them. The fix is to just move that block to preload.

Fixes #1604.

### DIFF
--- a/sources/web/datalab/static/notebook-app.js
+++ b/sources/web/datalab/static/notebook-app.js
@@ -130,6 +130,64 @@ define(['appbar', 'minitoolbar', 'idle-timeout', 'util'], function(appbar, minit
       updateNavigation();
     });
 
+    require(['notebook/js/codecell'], function(ipy) {
+      var codeCell = ipy.CodeCell;
+      var codeCellProto = codeCell.prototype;
+
+      var statusContent = document.getElementById('cellStatus').text;
+      codeCellProto.set_input_prompt = function(number) {
+        this.input_prompt_number = number;
+
+        if (number !== undefined) {
+          if (number === null) {
+          }
+          else if (number == '*') {
+            this.element.addClass('session');
+            this.element.removeClass('completed');
+
+            var status = $(statusContent);
+            this.element.find('div.output').css('display', '').empty().append(status);
+            status.delay(2000).show(0);
+          }
+          else {
+            this.element.addClass('completed');
+            var status = this.element.find('div.status');
+            if (status) {
+              status.remove();
+            }
+          }
+        } else {
+          var status = this.element.find('div.status');
+          if (status) {
+            status.remove();
+          }
+        }
+      }
+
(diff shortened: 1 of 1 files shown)


nikhilk:
The behavior we had was to ensure only the cell with focus would show line numbers. Is that still the case?

IamJeffG:
@nikhilk Do you mean that the desired behavior was for line numbers to toggle automatically on a per-cell basis when I switch cells?  That is not what happens in your latest release, but if it were, wouldn't it have the effect of physically shifting blocks of code in the notebook while scrolling through the cells? (the line number pushes the code to the right) That sounds like a very jarring UI.

That said, I went back to jupyter notebook to better understand the good behavior I'm used to. I will comment more about this on my original issue #1604 .

IamJeffG:
> The behavior we had was to ensure only the cell with focus would show line numbers.

Upon closer inspection of datalab's behavior (see gif below) I think you're _not_ referring to hiding the left margin for line numbers (as I thought in my preceding comment), but rather about leaving the line number container but just hiding the line number text.  This is fine and is very satisfactory as long as their (empty) container remains.

https://user-images.githubusercontent.com/2002703/29717051-9c808510-8962-11e7-9995-555c7662562f.gif

yebrahim:
@nikhilk I don't remember seeing this behavior, and the code doesn't indicate this being the case, it's just enabling line numbers.

nikhilk:
I remember adding css to toggle line number visibility (the margin space would still be taken, so text doesn't move around). This is to avoid clutter of having line numbers for all cells showing all the time. Perhaps things changed in stylesheets along the way to have broken that?

IamJeffG:
My screenshot above confirms the behavior @nikhilk is describing (line numbers go invisible on inactive cells). However I took this screenshot on the 1.2.20170818 release, not on this feature branch.

PR #1607, opened 2017-08-25
Add a fake metadata server that can be used to inject credentials

ojarjur:
To use the new metadata server, first create the docker network on which it will run:

```sh
docker network create -d bridge --internal --subnet=169.254.169.254/16 metadata
```

Then run the container:

```sh
docker run -it --rm --name=datalab \
    --add-host="metadata.google.internal:169.254.169.254" \
    -p 127.0.0.1:8081:8080 \
    --env DATALAB_FAKE_METADATA_SERVER=true \
    --env PROJECT_NUMBER=$(gcloud projects describe $(gcloud config get-value core/project) --format "value(projectNumber)") \
    --env PROJECT_ID=$(gcloud config get-value core/project) \
    datalab
```

And attach it to that network:

```sh
docker network connect --ip 169.254.169.254 metadata datalab
```

You can then push your local credentials to the container with:

```sh
curl -X POST \
    -d "{\"account\": \"$(gcloud auth list --format 'value(account)' --filter 'status:ACTIVE')\", \"scopes\": \"https://www.googleapis.com/auth/cloud-platform\", \"access_token\": \"$(gcloud auth print-access-token)\", \"expires_in\": 300, \"token_type\": \"Bearer\"}" \
    http://localhost:8081/api/creds
```

... and then code cells run inside the container will use those credentials.

### DIFF
--- a/sources/web/datalab/metadata.ts
+++ b/sources/web/datalab/metadata.ts
@@ -53,8 +53,8 @@ function launchFakeServer(metadata: FakeMetadata): void {
                            metadata_host, port, JSON.stringify(metadata));
 
   const server = http.createServer((req, res) => {
-    var parsed_url = url.parse(req.url, true);
-    var urlpath = parsed_url.pathname;
+    const parsed_url = url.parse(req.url, true);
+    const urlpath = parsed_url.pathname;
     logging.getLogger().info('Service a fake metadata request at %s', urlpath);
 
     if (urlpath == '/computeMetadata/v1/project/numeric-project-id') {
@@ -86,7 +86,7 @@ function launchFakeServer(metadata: FakeMetadata): void {
       res.write(metadata.creds.scopes);
     } else if (urlpath == '/computeMetadata/v1/instance/service-accounts/default/token' ||
                urlpath == '/computeMetadata/v1/instance/service-accounts/' + metadata.creds.account + '/token') {
-      var token: any = {
+      const token: any = {
         access_token: metadata.creds.access_token,
         expires_in: metadata.creds.expires_in,
         token_type: metadata.creds.token_type,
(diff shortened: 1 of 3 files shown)


sources/web/datalab/metadata.ts
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Implements the token submission request handling.
+ * @param request the incoming token POST request.
+ * @param response the outgoing response.
+ */
+function requestHandler(request: http.ServerRequest, response: http.ServerResponse): void {
+  var requestUrl = url.parse(request.url);
+  var path = requestUrl.pathname;
+  if (request.url.indexOf('/api/creds') === 0 && 'POST' == request.method) {
```
chmeyers: Would this also match "/api/creds_not_really_the_creds_url"?  Not sure if we care.

ojarjur: Yes, it would have. I don't think we care, but there's no reason to allow that, so I've changed it so that will not match.

sources/web/datalab/metadata.ts
```diff
@@ -0,0 +1,162 @@
...
+  token_type: string;
+}
+
+interface FakeMetadata {
+  project: string;
+  project_number: string;
+  creds: Creds;
+}
+
+var metadata: FakeMetadata = {
```
jimmc: Use const or let everywhere instead of var

sources/web/datalab/metadata.ts
```diff
@@ -0,0 +1,162 @@
...
+ * @param response the outgoing response.
+ */
+function requestHandler(request: http.ServerRequest, response: http.ServerResponse): void {
+  var requestUrl = url.parse(request.url);
+  var path = requestUrl.pathname;
+  if (request.url == '/api/creds' && 'POST' == request.method) {
+    parseCreds(request, function(c: any): void {
+      const creds = metadata.creds;
+      for (const key of Object.keys(c)) {
+        (<any>creds)[key] = c[key];
```
jimmc: Use the "as" format for casts: creds as any

sources/web/datalab/metadata.ts
ojarjur: Done.

sources/web/datalab/metadata.ts
ojarjur: Done.

PR #1610, opened 2017-08-25
Skip basepath for settings requests

yebrahim:
Settings manager shouldn't care about basepath. For now the frontend server will serve those immediately, and this might change later.

### DIFF
--- a/sources/web/datalab/polymer/modules/base-api-manager/base-api-manager.ts
+++ b/sources/web/datalab/polymer/modules/base-api-manager/base-api-manager.ts
@@ -74,7 +74,7 @@ interface ApiManager {
    * base path, and expecting a JSON response. This method returns immediately
    * with a promise that resolves with the parsed object when the request completes.
    */
-  sendRequestAsync(url: string, options?: XhrOptions): Promise<any>;
+  sendRequestAsync(url: string, options?: XhrOptions, prependBasepath?: boolean): Promise<any>;
 
   /**
    * Sends an XMLHttpRequest to the specified URL, adding the required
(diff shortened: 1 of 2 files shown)


sources/web/datalab/polymer/modules/base-api-manager/base-api-manager.ts
```diff
@@ -98,14 +98,22 @@ abstract class BaseApiManager implements ApiManager {
 
   public abstract getBasePath(): Promise<string>;
 
-  public async sendRequestAsync(url: string, options?: XhrOptions): Promise<string> {
+  public async sendRequestAsync(url: string, options?: XhrOptions, prependBasepath = true)
+      : Promise<any> {
     const basepath = await this.getBasePath();
-    return this._xhrJsonAsync(basepath + url, options);
+    if (prependBasepath) {
```
ojarjur: I think the `const basepath = ...` line should go inside this branch rather than before it.

sources/web/datalab/polymer/modules/base-api-manager/base-api-manager.ts
```diff
@@ -98,14 +98,22 @@ abstract class BaseApiManager implements ApiManager {
...
+    }
+    return this._xhrJsonAsync(url, options);
   }
 
-  public async sendTextRequestAsync(url: string, options?: XhrOptions): Promise<string> {
+  public async sendTextRequestAsync(url: string, options?: XhrOptions, prependBasepath = true)
+      : Promise<string> {
     const basepath = await this.getBasePath();
-    return this._xhrTextAsync(basepath + url, options);
+    if (prependBasepath) {
```
ojarjur: Same comment here.

sources/web/datalab/polymer/modules/base-api-manager/base-api-manager.ts
yebrahim: Done.

sources/web/datalab/polymer/modules/base-api-manager/base-api-manager.ts
yebrahim: Done.

PR #1611, opened 2017-08-25
Fetch all BQ projects and sort the resulting list.

jimmc:


sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -12,6 +12,13 @@
  * the License.
  */
 
+type ListDatasetsResponse = gapi.client.bigquery.ListDatasetsResponse;
```
yebrahim: Much cleaner.
FYI, I'm waiting on [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/19083) so we can use the gapi types from a declaration file that's automatically generated from the sources.

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -92,19 +99,31 @@ class BigQueryFileManager implements FileManager {
     throw new UnsupportedMethod('getEditorUrl', this);
   }
 
+  private async _listMoreProjects(accumulatedProjects: ProjectResource[],
```
yebrahim: Do we want to name this function differently so it can work with projects/datasets/tables? Maybe pass in the type of objects or the actual API to call?

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -92,19 +99,31 @@ class BigQueryFileManager implements FileManager {
...
+      pageToken: string): Promise<DatalabFile[]> {
+    const response: HttpResponse<ListProjectsResponse> =
+        await GapiManager.bigquery.listProjects(pageToken);
+    const additionalProjects = response.result.projects || [];
+    const projects = accumulatedProjects.concat(additionalProjects);
+    if (response.result.nextPageToken) {
+      return this._listMoreProjects(projects, response.result.nextPageToken);
+    } else {
+      projects.sort((a: ProjectResource, b: ProjectResource) => {
```
yebrahim: Curious how long this takes on a 1000+ projects?

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
jimmc: Yes, that will be nice.

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
jimmc: About 12 seconds for my 1200 projects. :-(

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
jimmc: I thought about that ... but based on YAGNI I decided to wait until we actually have a project with more than 1000 datasets or a dataset with more than 1000 tables.
If you think it is worth doing this now, I can take another look at it.

yebrahim: All hail YAGNI, but I think this is something any user might hit, so we can't wait until this happens to us.
If you don't want to put it as part of this PR, maybe put a hi-pri TODO?

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
yebrahim: Just the sort or the entire thing?

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
jimmc: I started trying to make this function generic, but there were enough differences that I decided to just write the equivalent methods for listDatasets and listTables. PTAL.

sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -72,11 +72,14 @@ class GapiManager {
     /**
      * Gets the list of BigQuery projects, returns a Promise.
      */
-    public static listProjects():
+    public static listProjects(pageToken?: string):
         Promise<gapi.client.HttpRequestFulfilled<gapi.client.bigquery.ListProjectsResponse>> {
       const request = {
         maxResults: 1000,
```
yebrahim: Add pageToken here in the request definition directly to be consistent with dataset and table requests below?

jimmc: Done.

PR #1612, opened 2017-08-25
Allow specifying file manager type as a query param

yebrahim:


sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -158,8 +158,15 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
 
     this._apiManager = ApiManagerFactory.getInstance();
 
+    // Allow forcing an file manager type.
```
jimmc: Is this supposed to be "any file manager type"?

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -158,8 +158,15 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
...
+    // TODO: Consider writing a config element instead of parsing URL parameters
+    //       everywhere configs are needed.
+    const params = new URLSearchParams(window.location.search);
+    if (params.has('filemanager')) {
```
jimmc: Seems like we would not need this if we instead added a file parameter that starts with a filemanager type:
?file=drive:12345
But if this is just for short-term debugging and demos, it's fine. In that case, maybe add a TODO that this should be removed once the file parameter is added.

yebrahim: Yes, until we get the path as a querystring we should remove this. Added a TODO.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
yebrahim: Done.

PR #1616, opened 2017-08-28
Create a new browser instance for each test

yebrahim:
This should help with the notebook test flakiness. Reusing the same browser object meant navigating away from notebooks after running them, which often failed because of the confirmation popup, which in turn wasn't getting dismissed properly because of timing issues.
This PR creates a new browser for each notebook.


ojarjur:
Thanks for digging in to this.

PR #1625, opened 2017-08-29
Remaining DriveFileManager APIs

yebrahim:


sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
```diff
@@ -56,13 +57,15 @@ class DriveFileManager implements FileManager {
       try {
         return new NotebookContent(JSON.parse(content).cells);
       } catch (e) {
-        throw new Error('Could not parse notebook: ' + e.message);
+        Utils.log.error('Could not parse notebook: ' + e.message);
+        return new TextContent(content);
```
jimmc: It seems a bit strange to log an error, but not return an error at this level. Maybe the caller catch the error and decide it is acceptable.

yebrahim: This is to handle cases where the file extension or mimetype are misleading, in which case I assume the general case and return the string content as is.
Also, the caller will still need to get the content even if an error occurs, and it seems odd to me that the exception itself would contain the content.

nikhilk: It might be better to have an error saying invalid file format -- or were you intending that the user actually gets dropped into text editing mode? Not sure we really want the user to operate at the level of notebook json, esp. if we can't tell where in the file there might have been an error. Plus our text editor isn't all that great of a JSON editor.

yebrahim: The main code path for invoking this functionality are the editor and the preview pane, both of which need to handle this gracefully in my opinion. If you have a malformed json you'll need the editor to fix it.

nikhilk: I don't think preview will want to show the raw json, correct?

In the case of editor, at the very least, you'll also want to display an error message about why the user is seeing a plain text view instead of a notebook view. Will the caller know when the result is a TextContent object?

Maybe TextContent should have an optional error facility that allows flowing that bit of information, so it can be displayed appropriately rather than logged.

yebrahim: Ok I think I addressed the concerns here about preview doing the parsing itself, but this quickly turned into a cleanup change, so I can move that into a separate PR if you prefer, but I'd like to move forward fast, so I'm sending it here.

The editor should not understand notebook parsing, as it's just a text editor, and should in fact be the go-to to edit a broken notebook.
Also, our text editor can be a pretty good json editor with the right mode :), with a auto-format button and all. Maybe we can add that later.

PR #1626, opened 2017-08-30
Editor fixes

yebrahim:
- CodeMirror wouldn't render full height inside a flex container.
- Focus the editor after each button click and after content reload.
- Move progress bar under toolbar for clarity.

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-editor/datalab-editor.html
+++ b/sources/web/datalab/polymer/components/datalab-editor/datalab-editor.html
@@ -33,12 +33,9 @@ the License.
       :host {
         background-color: var(--primary-bg-color);
         height: 100%;
-        display: flex;
-        flex-direction: column;
       }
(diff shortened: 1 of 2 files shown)


sources/web/datalab/polymer/components/datalab-editor/datalab-editor.html
```diff
@@ -47,7 +44,7 @@
       }
       #editorToolbar {
         display: flex;
-        flex: 0 0 var(--toolbar-height);
+        height: calc(var(--toolbar-height) - 1px); /* for border-bottom */
```
jimmc: Is it worth using a variable for the border thickness?

yebrahim: I don't think these two occurrences warrant a variable, so I'll merge this as is, and we can revisit if you think it's much cleaner.

PR #1629, opened 2017-08-30
Use resizable divider on preview pane in file browser.

jimmc:
* Use resizable divider in file-browser element
* Make divider always visible to give the user an affordance
* Update file-browser visibility button to work with resizable divider
* Allow right pane in file browser to be opened and closed either using the visibility icon or by sliding the divider

This is what the divider now looks like in the file browser. It changes to a slightly different shade when active. We can improve the visuals later.
resizable-divider

### DIFF
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.html
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.html
@@ -254,9 +254,8 @@ the License.
 
       <!--File listing-->
       <resizable-divider id="filesContainer" class="files-container"
-          divider-position="{{_dividerPosition}}" minimum-width-px="0"
-          hide-right="{{!_isPreviewPaneToggledOn}}"
-        >
+          divider-position="{{_dividerPosition}}"
+          hide-right="{{!_isPreviewPaneToggledOn}}">
         <item-list id="files" slot="left" hide-header$="{{small}}" disable-selection$="{{small}}"></item-list>
         <preview-pane id="previewPane" slot="right" file="{{selectedFile}}" active="{{_isPreviewPaneEnabled}}"
                       class$="preview-pane-enabled--{{_isPreviewPaneEnabled}}">
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.html
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.html
@@ -253,12 +253,15 @@ the License.
       </paper-progress>
 
       <!--File listing-->
-      <div class="files-container">
-        <item-list id="files" hide-header$="{{small}}" disable-selection$="{{small}}"></item-list>
-        <preview-pane id="previewPane" file="{{selectedFile}}" active="{{_isPreviewPaneEnabled}}"
+      <resizable-divider id="filesContainer" class="files-container"
+          divider-position="{{_dividerPosition}}" minimum-width-px="0"
+          hide-right="{{!_isPreviewPaneToggledOn}}"
+        >
+        <item-list id="files" slot="left" hide-header$="{{small}}" disable-selection$="{{small}}"></item-list>
+        <preview-pane id="previewPane" slot="right" file="{{selectedFile}}" active="{{_isPreviewPaneEnabled}}"
                       class$="preview-pane-enabled--{{_isPreviewPaneEnabled}}">
         </preview-pane>
-      </div>
+      </resizable-divider>
     </div>
 
   </template>
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -68,6 +68,7 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
 
   private _addToolbarCollapseThreshold = 900;
   private _apiManager: ApiManager;
+  private _dividerPosition: number;
   private _previewPaneCollapseThreshold = 600;
   private _fetching: boolean;
   private _fileList: DatalabFile[];
@@ -85,6 +86,11 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
 
   static get properties() {
     return {
+      _dividerPosition: {
+        observer: '_dividerPositionChanged',
+        type: Number,
+        value: 70,
+      },
       _fetching: {
         type: Boolean,
         value: false,
@@ -291,7 +297,8 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
 
   /**
    * Called when the selection changes on the item list. If exactly one file
-   * is selected, sets the selectedFile property to the selected file object.
+   * is selected, sets the selectedFile property to the selected file object
+   * and opens the preview pane.
    */
   _handleSelectionChanged() {
     const selectedIndices = (this.$.files as ItemListElement).selectedIndices;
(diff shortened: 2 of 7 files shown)


sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
```diff
@@ -79,40 +99,77 @@ class ResizableDividerElement extends Polymer.Element {
...
+    this.dividerPosition =
+        (newPosition < 0) ? 0 : (newPosition > 100) ? 100 : newPosition;
+    // Let observer call _dividerMoved
+  }
+
+  /**
+   * Calculate the new divider position after hideRight changes.
+   */
+  _hideChanged() {
+    if (!this._readyDone) {
```
yebrahim: nit: Without using a variable, you could also check that the first argument is undefined, which is the case when the observer callback is called because of initial property setters.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
```diff
@@ -79,40 +99,77 @@ class ResizableDividerElement extends Polymer.Element {
...
+      return;   // Leave divider position unchanged on startup
+    } else if (this.hideRight) {
+      this.dividerPosition = 100;
+    } else {
+      // Make the right pane visible
+      if (this._lastMouseUpPosition > 0 &&
+          this._lastMouseUpPosition < 100) {
+        this.dividerPosition = this._lastMouseUpPosition;
+      } else if (this._lastMouseDownPosition > 0 &&
```
yebrahim: Curious when does this happen?

sources/web/datalab/polymer/components/file-browser/file-browser.html
```diff
@@ -253,12 +253,15 @@
       </paper-progress>
 
       <!--File listing-->
-      <div class="files-container">
-        <item-list id="files" hide-header$="{{small}}" disable-selection$="{{small}}"></item-list>
-        <preview-pane id="previewPane" file="{{selectedFile}}" active="{{_isPreviewPaneEnabled}}"
+      <resizable-divider id="filesContainer" class="files-container"
+          divider-position="{{_dividerPosition}}" minimum-width-px="0"
+          hide-right="{{!_isPreviewPaneToggledOn}}"
```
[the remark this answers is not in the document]
jimmc: Done.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -291,7 +297,8 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
 
   /**
    * Called when the selection changes on the item list. If exactly one file
-   * is selected, sets the selectedFile property to the selected file object.
+   * is selected, sets the selectedFile property to the selected file object
+   * and opens the preview pane.
```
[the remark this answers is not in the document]
jimmc: No; for a while I had it auto-opening, but I didn't like the effect, so I took that out.
Comment corrected.

sources/web/datalab/polymer/components/file-browser/file-browser.html
```diff
@@ -253,12 +253,15 @@
       </paper-progress>
 
       <!--File listing-->
-      <div class="files-container">
-        <item-list id="files" hide-header$="{{small}}" disable-selection$="{{small}}"></item-list>
-        <preview-pane id="previewPane" file="{{selectedFile}}" active="{{_isPreviewPaneEnabled}}"
+      <resizable-divider id="filesContainer" class="files-container"
+          divider-position="{{_dividerPosition}}" minimum-width-px="0"
```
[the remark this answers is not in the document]
jimmc: Done.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
```diff
@@ -29,9 +29,14 @@
       }
       .pane {
         overflow: auto;
-        width: 50%;
+        height: 100%;
+        width: calc(50% - var(--divider-half-width));
         user-select: none;
       }
+      .right {
+        position: absolute;
```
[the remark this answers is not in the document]
jimmc: It sets both the left and the width of the right pane. On resize, there is a resize handler that recomputes.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.html
```diff
@@ -51,24 +56,24 @@
         top: 0px;
         bottom: 0px;
         left: calc(50% - var(--divider-half-width));
-        right: calc(50% + var(--divider-half-width));
         width: calc(2 * var(--divider-half-width));
+        background-color: var(--selection-bg-color);
```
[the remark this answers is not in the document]
jimmc: I took out the minimum-width=0, so it now snaps between 0 and the minimum width.
I want the divider to appear to make it clear the other window is still there, just collapsed, which allows using just the divider bar without requiring an additional visibility toggle, which is not part of the resizable-divider component.
I can demo if you want, and discuss how it should work.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
```diff
@@ -24,20 +24,37 @@ class ResizableDividerElement extends Polymer.Element {
...
+  private _readyDone = false;
+
   static get is() { return 'resizable-divider'; }
 
   static get properties() {
     return {
-      initialDividerPosition: {
+      dividerPosition: {
+        notify: true,
+        observer: '_dividerMoved',
```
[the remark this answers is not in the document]
jimmc: Done.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
jimmc: Do you mean, add an argument such as _hideChanged(foo)? That doesn't appear to work, I never see an undefined value.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
jimmc: The mouse-up value is used as long as the user has not dragged the divider all the way over to close the pane; but if that did happen, then the mouse-down value is used, i.e. where the user dragged from the last time.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
```diff
@@ -79,40 +99,77 @@ class ResizableDividerElement extends Polymer.Element {
...
 
   /**
-   * Calculates and sets the new widths of the two panes.
+   * Calculates and sets the new widths of the two panes after the divider moved.
    */
-  _resizePanes(newPos: number) {
+  _dividerMoved() {
     const container = this.$.container as HTMLDivElement;
+    const containerRect = container.getBoundingClientRect();
+    const fullWidthLessDivider = containerRect.width - this._dividerWidth;
```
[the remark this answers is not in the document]
jimmc: Done.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
yebrahim: Sorry, I meant the second argument. See https://www.polymer-project.org/2.0/docs/devguide/observers#change-callbacks. The second argument is old value, and the first thing done on element init is to set the property to some default value, which will end up calling the observer with `(newValue = default, oldValue = undefined)`.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
yebrahim: Ah makes sense.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
```diff
@@ -76,12 +76,12 @@ class ResizableDividerElement extends Polymer.Element {
     this._dividerWidth = divider.getBoundingClientRect().width;
 
     // Initialize the divider position.
-    this._dividerMoved();
+    // this._dividerPositionChanged();
```
yebrahim: Remove. Also remove the comment on 78.

yebrahim:
I'm still a little skeptic about a collapsed pane that shows the divider, I think it should just disappear and wait for an attribute to be set / method called.

sources/web/datalab/polymer/components/resizable-divider/resizable-divider.ts
jimmc: Done.

jimmc:
Could certainly use more work, but yeah, I'd rather do that in a separate PR.

PR #1637, opened 2017-09-06
[WIP] Extending default datalab docker

pcejrowski:
This pull request contains additional extensions that we need to add to default datalab docker image to make it usable for our data science users.


pcejrowski:
Wrong target branch...

PR #1640, opened 2017-09-07
Make "file" query param work in file-browser

jimmc:
* Allow file query parameter to be entered in the URL
* Update the file query parameter in the URL while browsing

A "file" query parameter can be entered with the URL, and the file-browser and data-browser will go to that directory.
While browsing, when a directory is selected, the "file" query parameter in the URL is updated.
The Data and Files tabs each maintain their own "file" parameter, and update that into the URL when selected.

The "file" parameter is a file-manager type (jupyter, bigquery, or drive), followed by a colon, then a path, which is interpreted by the file-manager.
In HTML5, most punctuation characters are, strictly speaking, not allowed in query parameters and should be percent-encoded. With the current implementation, it works to enter colon and slashes in the file parameter. Slashes are preserved in the file parameter when browsing, but the colon gets converted to %3A.

Since the file id for a Drive object does not include a path, the "file" parameter in the URL does not include it either. When directly entering a URL with a "drive:<drive-id>" value for the "file" parameter, the current code does not attempt to recreate the path to that object.


sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -191,11 +219,15 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
...
-      this.readyPromise = this._loadStartupPath()
+      this.readyPromise = this._loadStartupPath(fileId)
           .then(() => this._finishLoadingFiles());
     }
 
-    return this.readyPromise;
+    return this.readyPromise.catch((e) => {
+      // TODO - present error info to user
+      this._fetching = false; // Stop looking busy
+      throw e;
```
yebrahim: If this should be a popup error, you can do `Utils.showErrorDialog(e.message)`.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -164,14 +173,33 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
...
 
-    // Allow forcing a file manager type.
+    const queryParams = (this.queryParams || {}) as {[key: string]: string};
+    const fileParamName = 'file';
+    const fileParam = queryParams[fileParamName] || '';
+    const filemanagerParamName = 'filemanager';
+    const filemanagerParam = queryParams[filemanagerParamName] || '';
+    let fileId: DatalabFileId|null = null;
+    // Ignore fileParam if we are not the visible tab
+    if (fileParam && this.offsetParent) {
```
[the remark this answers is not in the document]
yebrahim: Save below at line 378.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -803,8 +846,15 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
         }
       }, this._fileListRefreshInterval);
     }
-    // Now refresh the list once.
-    this._fetchFileList();
+    // Refresh the list and update the window location.
+    this._pathHistoryIndexChanged();
```
yebrahim: If you're calling this method to get the convenience of calling both `_fetchFileList` and `_setFileParamToCurrentFile`, I'd inline the two calls here, since that method has different semantics, and it could change at any point.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -818,28 +868,18 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
     }
   }
 
-  private async _loadStartupPath() {
+  private async _loadStartupPath(fileId: DatalabFileId|null) {
     // TODO - move this to SettingsManager and make it able to store startuppaths
```
yebrahim: Move down next to line 878 where the `SettingsManager` is used?

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -99,6 +99,32 @@ class BigQueryFileManager implements FileManager {
...
+        this._bqProjectIdToDatalabFile(pathParts[0])
+      ];
+    }
+    if (pathParts.length === 2) {
+      return [
+        this._bqProjectIdToDatalabFile(pathParts[0]),
+        this._bqProjectDatasetIdsToDatalabFile(pathParts[0], pathParts[1]),
+      ];
+    } else {
+      // We ignore any path parts past 3
```
yebrahim: Worth logging an error message for length > 3. Maybe even throwing an error?

sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
```diff
@@ -144,4 +144,18 @@ class DriveFileManager implements FileManager {
...
         '/notebook?file=' + fileId.toQueryString();
   }
+
+  public pathToPathHistory(path: string): DatalabFile[] {
+    if (path == '') {
+      return [];
+    } else {
+      // TODO - create the real path to this object, or figure out
+      // a better way to handle not having the full path in the breadcrumbs
+      const fileId = path;
```
yebrahim: Why this temp variable?

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -336,9 +371,17 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
...
 
     this.currentFile = this._pathHistory[this._pathHistoryIndex];
+    this._setFileParamToCurrentFile();
     this._fetchFileList();
   }
 
+  _setFileParamToCurrentFile() {
+    // Only update the location if we are currently visible
+    if (this.currentFile && this.offsetParent) {
+      this.set('queryParams.file', this.currentFile.id.toQueryString());
```
yebrahim: If we do something like `window.history.pushState` here, doesn't this already move all the history logic to the browser?

yebrahim: .. And then the back/forward buttons of course will need to do `window.history.{backward|forward}()`.

jimmc: Updating queryParams this way actually does add to the browser history. I added some code to observe changes to queryParams and update the page, which makes the browser's backwards and forwards buttons work properly - and it doesn't reload the page. There is some kind of problem when doing this with drive, though: it is sending a query to the drive API, but not updating the screen. I spent a bit of time trying to track that down, with no success yet. I don't know if it's a problem in drive-file-manager, file-browser, or how they connect.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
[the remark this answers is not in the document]
jimmc: Took another look, this is getting called multiple times on the same element, so perhaps my analysis was wrong, but it at least cuts out some unnecessary execution now.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
jimmc: Added some code that makes browser history work on drive also. So now the only oddity I am aware of is that the breadcrumbs in drive don't get set up on direct entry of the file parameter, and thus are not preserved browser forward/backwards.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -115,6 +134,19 @@ class DatalabAppElement extends Polymer.Element {
     }
   }
 
+  _queryParamsChanged() {
+    const queryParams = (this.queryParams || {}) as {[key: string]: string};
+    const fileParamName = 'file';
+    const fileParam = queryParams[fileParamName] || '';
+    if (fileParam !== this.fileId) {
+      this.fileId = fileParam;
```
yebrahim: Instead of this, why don't we set the active page element's `fileId`, and call this method on page change event? This saves the `file-browser` element from having to figure out if it's active/visible or now.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -915,7 +921,11 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
     // Always add the root file to the beginning.
     const root = await this._fileManager.getRootFile();
     this._pathHistory.unshift(root);
-    this._pathHistoryIndex = this._pathHistory.length - 1;
+    if (this._pathHistoryIndex === this._pathHistory.length - 1) {
+      this._pathHistoryIndexChanged();
```
yebrahim: I'm not sure I understand the logic here. Do we want the callback to fire all the time?

jimmc: Yes. For drive, the history length is always 1, but we're changing it and we want the code that reloads the file list to get executed. That code is triggered on a change to pathHistoryIndex, so even when pathHistoryIndex is not changing, we still want to call that method here.

sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
jimmc: I think the active vs visible are two separate questions, as it seems to get called when the page is active not not yet visible. We would want to call it from more than just the page-change event, since the fileId can change without the page changing. I think it may be possible to do it the way you are suggesting, but yes, I'd rather commit this now and maybe make that change later.

PR #1642, opened 2017-09-07
Fix height problem for divider in file-browser.

jimmc:
This makes the divider container in the file-browser full height.

Before:
divider-before

After:
divider-after


sources/web/datalab/polymer/components/file-browser/file-browser.html
```diff
@@ -102,6 +102,7 @@
         display: flex;
         flex-direction: column;
         flex: 1 1 auto;
```
yebrahim: Do we need this line if we have a 100% height?

jimmc: Apparently not. Removed.

PR #1643, opened 2017-09-07
Make /exp root path work for editor.

jimmc:


sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
```diff
@@ -136,8 +136,7 @@ class DriveFileManager implements FileManager {
   }
 
   public async getEditorUrl(fileId: DatalabFileId) {
-    return location.protocol + '//' + location.host +
-        '/editor?file=' + fileId.toQueryString();
+    return Utils.getHostRoot() + '/editor?file=' + fileId.toQueryString();
```
yebrahim: Why not do this for `getNotebookUrl` and other places too?

yebrahim: Actually I take that back. `Utils.getHostRoot` will not return the same thing if called outside of `datalab-app`, do we want to revisit this? What if we call it from the editor app or notebook, where the sidebar doesn't exist?

jimmc: /exp is not used for the notebook URL, as that has not changed.

PR #1658, opened 2017-09-12
Update the Datalab container to TensorFlow 1.3.0.

craigcitro:
PTAL @chmeyers @qimingj 

Last time I recall having issues running tests on Travis because I couldn't create a branch in the datalab repo. LMK if there's anything I should do to make it easier to test.


nikhilk:
Note that at this point we're staying with 1.2.x to match what we have on
the service in ML Engine.

PR #1663, opened 2017-09-14
Implement inline details showing BigQuery sample data in file browser

jimmc:
This PR implements a generic inline-details mechanism in the file browser, and a specific implementation that shows a small number of data rows when the file is a BigQuery file. This currently shows 5 rows of data. If the table is too wide, it will include a horizontal scrollbar.

At the moment, the display of the inline details is using the same algorithm as the sidebar preview, which is that it shows up if the row is the only selected row. It should be pretty easy to change this later once we figure out what UX we want to use.

inline-details

### DIFF
--- a/sources/web/datalab/polymer/components/inline-details-pane/inline-details-pane.html
+++ b/sources/web/datalab/polymer/components/inline-details-pane/inline-details-pane.html
@@ -35,7 +35,6 @@ the License.
     <div id="container" hidden$="{{!file}}">
       <iron-pages id="inline-details" selected="{{details}}" attr-for-selected="name">
         <table-inline-details class="inline-details" name="table" file="{{file}}" active="{{active}}"></table-inline-details>
-        <text-inline-details class="inline-details" name="text" file="{{file}}" active="{{active}}"></text-inline-details>
       </iron-pages>
     </div>
   </template>
--- a/sources/web/datalab/polymer/components/inline-details-pane/inline-details-pane.html
+++ b/sources/web/datalab/polymer/components/inline-details-pane/inline-details-pane.html
@@ -26,11 +26,6 @@ the License.
         display: flex;
         flex-direction: column;
       }
-      .placeholder {
-        padding: 15px;
-        text-align: center;
-        color: var(--neutral-fg-color);
-      }
     </style>
     <div id="container" hidden$="{{!file}}">
       <iron-pages id="inline-details" selected="{{details}}" attr-for-selected="name">
--- a/sources/web/datalab/polymer/components/inline-details-pane/inline-details-pane.html
+++ b/sources/web/datalab/polymer/components/inline-details-pane/inline-details-pane.html
@@ -1,0 +1,44 @@
+<!--
+Copyright 2017 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+in compliance with the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under the License
+is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+or implied. See the License for the specific language governing permissions and limitations under
+the License.
+-->
+
+<link rel="import" href="../../bower_components/iron-pages/iron-pages.html">
+
+<link rel="import" href="../../components/shared-styles/shared-styles.html">
+<link rel="import" href="../../modules/file-manager-factory/file-manager-factory.html">
+
+<dom-module id="inline-details-pane">
+  <template>
+    <style include="datalab-shared-styles">
+      #container {
+        height: 100%;
+        overflow-y: auto;
+        display: flex;
+        flex-direction: column;
+      }
+      .placeholder {
+        padding: 15px;
+        text-align: center;
+        color: var(--neutral-fg-color);
+      }
+    </style>
+    <div id="container" hidden$="{{!file}}">
+      <iron-pages id="inline-details" selected="{{details}}" attr-for-selected="name">
+        <table-inline-details class="inline-details" name="table" file="{{file}}" active="{{active}}"></table-inline-details>
+        <text-inline-details class="inline-details" name="text" file="{{file}}" active="{{active}}"></text-inline-details>
+      </iron-pages>
+    </div>
+  </template>
+</dom-module>
+
+<script src="inline-details-pane.js"></script>
--- a/sources/web/datalab/polymer/components/item-list/item-list.ts
+++ b/sources/web/datalab/polymer/components/item-list/item-list.ts
@@ -18,12 +18,17 @@
 class ItemListRow {
   public selected: boolean;
   public columns: string[];
+  public detailsName: string;
+  public showInlineDetails = false;
 
   private _icon: string;
 
-  constructor(columns: string[], icon: string, selected?: boolean) {
+  constructor({columns, icon, selected, detailsName}:
+      { columns: string[], icon: string,
+        selected?: boolean, detailsName?: string}) {
     this.columns = columns;
     this.selected = selected || false;
+    this.detailsName = detailsName || '';
     this._icon = icon;
   }
 
@@ -34,6 +39,21 @@ class ItemListRow {
    */
   get icon() { return this._hasLinkIcon() ? '' : this._icon; }
   get src() { return this._hasLinkIcon() ? this._icon : ''; }
+
+  /**
+   * Updates our showInlineDetails flag after a selection change.
+   */
+  _updateShowInlineDetails() {
+    if (!this.detailsName) {
+      this.showInlineDetails = false;
+      return;
+    }
+    if (!this.selected) {
+      this.showInlineDetails = false;
+      return;
+    }
+    this.showInlineDetails = true;
+  }
 
   private _hasLinkIcon() {
     return this._icon.startsWith('http://') || this._icon.startsWith('https://');
@@ -168,6 +188,9 @@ class ItemListElement extends Polymer.Element {
    */
   _selectItem(index: number) {
     this.set('rows.' + index + '.selected', true);
+    this.rows[index]._updateShowInlineDetails();
+    this.notifyPath('rows.' + index + '.showInlineDetails',
+        this.rows[index].showInlineDetails);
   }
 
   /**
@@ -176,6 +199,9 @@ class ItemListElement extends Polymer.Element {
    */
   _unselectItem(index: number) {
     this.set('rows.' + index + '.selected', false);
+    this.rows[index]._updateShowInlineDetails();
+    this.notifyPath('rows.' + index + '.showInlineDetails',
+        this.rows[index].showInlineDetails);
   }
 
   /**
--- a/sources/web/datalab/polymer/components/item-list/item-list.ts
+++ b/sources/web/datalab/polymer/components/item-list/item-list.ts
@@ -16,11 +16,24 @@
  * Mode definition for which items, out of all items that are capable
  * of showing details, actually show those details.
  */
-enum InlineDetailsDisplay {
+enum InlineDetailsDisplayMode {
   NONE,               // Don't show any inline details elements
   SINGLE_SELECT,      // Show details only when a single item is selected
   MULTIPLE_SELECT,    // Show details for all selected items
   ALL                 // Show details for all items
+}
+// TODO: the current behavior of ALL does not show all of the details when
+// the page first renders, but only displays the details when the
+// selection state for an item changes. We might want to change this
+// behavior so that it figures out in advance that it should show
+// all the details.
+
+/** Fields that can be passed to the ItemListRow constructor. */
+interface ItemListRowParameters {
+  columns: string[];
+  createDetailsElement?: () => HTMLElement;
+  icon: string;
+  selected?: boolean;
 }
 
 /**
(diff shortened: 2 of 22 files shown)


sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -167,15 +240,50 @@ class ItemListElement extends Polymer.Element {
...
   _unselectItem(index: number) {
-    this.set('rows.' + index + '.selected', false);
+    this._changeSelectItem(index, false);
+  }
+
+  /**
+   * Updates the show-details flag for a row after selection change.
+   */
+  _changeSelectItem(index: number, newValue: boolean) {
+    const previousSelectedCount =
```
yebrahim: nit: I'd move these three lines down when they're needed in the `if` condition, and separate that section by the comment down there.

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -167,15 +240,50 @@ class ItemListElement extends Polymer.Element {
...
+        this.selectedIndices.length + (newValue ? -1 : 1);
+    const previousMultiSelect = previousSelectedCount > 1;
+    this.set('rows.' + index + '.selected', newValue);
+    const multiSelect = this.selectedIndices.length > 1;
+    const nthDivSelector = 'div.row-details:nth-of-type(' + (index + 1) + ')';
```
yebrahim: Can we move these two lines to a `_getRowElement` method?

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -167,15 +240,50 @@ class ItemListElement extends Polymer.Element {
...
+    const rowDetailsElement = this.$.listContainer.querySelector(nthDivSelector);
+    this.rows[index]._updateShowInlineDetails(
+        this.inlineDetailsMode, multiSelect, rowDetailsElement);
+    this.notifyPath('rows.' + index + '.showInlineDetails',
```
yebrahim: Who's listening for these events?

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.html
```diff
@@ -0,0 +1,92 @@
...
+  <template>
+    <style include="datalab-shared-styles">
+      #placeholder {
+        height: 100%;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+      }
+      paper-spinner {
+        --paper-spinner-stroke-width: 2px;
```
yebrahim: Seems like a common rule, should we move it to `shared-styles`?

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.html
```diff
@@ -0,0 +1,92 @@
...
+      }
+      #container {
+        padding: 0px 20px;
+        height: 100%;
+        overflow-y: auto;
+        color: var(--primary-fg-color);
+      }
+      .strong {
+        font-weight: bold;
```
yebrahim: Also seems like a common rule.

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.html
```diff
@@ -0,0 +1,92 @@
...
+        overflow-wrap: break-word;
+      }
+      table {
+        border-spacing: 1px;
+        padding: 0 0 4px 0;
+        width: 100%;
+      }
+      th.field {
+        padding: 10px;
+        background-color: #eee;
```
yebrahim: Use `var(--secondary-bg-color)` here, same in `td.field`.

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.html
```diff
@@ -0,0 +1,92 @@
...
+      .separator {
+        margin: 20px 0px;
+      }
+    </style>
+    <div id="placeholder" hidden$={{_table}}>
+      <paper-spinner active hidden$={{!_busy}}></paper-spinner>
+    </div>
+    <div id="container" hidden$={{!_table}}>
+      <table id="tabledata">
+        <tr class="header">
```
yebrahim: Where's `header` defined?

yebrahim: Same for `field-name` and `field-type` below.

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.html
```diff
@@ -0,0 +1,92 @@
...
+      th.field {
+        padding: 10px;
+        background-color: #eee;
+        font-weight: bold;
+      }
+      td.field {
+        padding: 10px;
+        background-color: #eee;
+      }
+      .separator {
```
yebrahim: Where is this used?

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts
```diff
@@ -0,0 +1,215 @@
...
+        notify: true, // For unit tests
+        type: Object,
+        value: null,
+      },
+      file: {
+        observer: '_fileChanged',
+        type: Object,
+        value: {},
+      },
+      schemaFields: {
```
yebrahim: Should this be private (underscored)?

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  _fileChanged() {
+    if (this.file && this.file.id) {
+      // TODO(jimmc) - move this into BigQueryFile?
+      const path = this.file.id.path;
+      const parts = path.split('/');
+      this.tableId = parts[0] + ':' + parts[1] + '.' + parts[2];
+    } else {
+      this.tableId = '';
```
yebrahim: If this is the only spot where `this.tableId` is set, we can just move the `_tableIdChanged` implementation here, or just call it directly without setting up an event listener/handler loop.

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts
```diff
@@ -0,0 +1,215 @@
...
+    }
+  }
+
+  _tableIdChanged() {
+    const matches = this.tableId.match(/^(.*):(.*)\.(.*)$/);
+    if (matches && matches.length === 4) { // The whole string is matched as first result
```
yebrahim: If the above comment applies we can just pick these parts from the parsing in `_fileChanged`.

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts
```diff
@@ -0,0 +1,215 @@
...
+      this._busy = true;
+      const projectId = matches[1];
+      const datasetId = matches[2];
+      const tableId = matches[3];
+
+      GapiManager.bigquery.getTableDetails(projectId, datasetId, tableId)
+        .then((response: HttpResponse<gapi.client.bigquery.Table>) => {
+          this._table = response.result;
+        }, (errorResponse: any) =>
+            Utils.log.error('Failed to get table details: ' + errorResponse.body))
```
yebrahim: Should we write some user-friendly error message to the inline details element in this case? I'm assuming this is what `placeholder` is originally for. More importantly below when getting the rows.

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts
```diff
@@ -0,0 +1,215 @@
...
+        }, (errorResponse: any) =>
+            Utils.log.error('Failed to get table rows: ' + errorResponse.body))
+        .then(() => this._busy = false);
+    } else {
+      this._table = null;
+      this._rows = [];
+    }
+  }
+
+  _computeCreationTime(table: gapi.client.bigquery.Table | null) {
```
yebrahim: Where are these computed properties used?

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  _formatMode(mode: string) {
+    return mode || this.DEFAULT_MODE;
+  }
+
+  /**
+   * Opens the current table in the table schema template notebook.
+   */
+  async _openInNotebook() {
```
yebrahim: Same as above, we don't have this in the inline details right?

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts
```diff
@@ -0,0 +1,215 @@
...
+      const projectId = matches[1];
+      const datasetId = matches[2];
+      const tableId = matches[3];
+
+      GapiManager.bigquery.getTableDetails(projectId, datasetId, tableId)
+        .then((response: HttpResponse<gapi.client.bigquery.Table>) => {
+          this._table = response.result;
+        }, (errorResponse: any) =>
+            Utils.log.error('Failed to get table details: ' + errorResponse.body))
+        .then(() => GapiManager.bigquery.getTableRows(projectId, datasetId, tableId, 5))
```
yebrahim: Define a const for the 5.

sources/web/datalab/polymer/test/file-browser-test.ts
```diff
@@ -12,6 +12,8 @@
  * the License.
  */
 
+class MockDatalabFile extends DatalabFile {}
```
yebrahim: This can move into `test-utils.ts`.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -281,4 +311,113 @@ describe('<item-list>', () => {
     assert(!isSelected(0), 'first item should not be selected when selection is disabled');
   });
 
+  describe('inline details', () => {
+    it('should not create a details element for an item without details', () => {
+      testFixture.inlineDetailsMode = InlineDetailsDisplay.SINGLE_SELECT;
+      const firstRow = getRow(0);
+      const firstRowDetailsContainer = getRowDetailsContainer(0);
+      firstRow.click();
```
yebrahim: Don't we need to call `Polymer.dom.flush();` after the click to make sure templates have rendered, since we have some dom-repeats?

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -281,4 +311,113 @@ describe('<item-list>', () => {
...
+
+      assert(!fourthRowDetailsContainer.firstChild,
+          'fourth item should not create details element when clicked');
+      assert(fourthRowDetailsContainer.getAttribute('hidden') != null,
+          'fourth details container should not be visible');
+    });
+
+    it('should show details even when not clicked in ALL mode', () => {
+      testFixture.inlineDetailsMode = InlineDetailsDisplay.ALL;
+      // TODO: the current behavior of ALL does not show all of the details when
```
yebrahim: I'd put this comment into the element itself too.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -281,4 +311,113 @@ describe('<item-list>', () => {
...
+      // the page first renders, but only displays the details when the
+      // selection state for an item changes. We might want to change this
+      // behavior so that it figures out in advance that it should show
+      // all the details.
+      testFixture._unselectAll();   // Recalculate details display for all items
+      const fourthRowDetailsContainer = getRowDetailsContainer(3);
```
yebrahim: Should we make sure that all the items that can create a details element have one?

sources/web/datalab/polymer/components/inline-details-pane/inline-details-pane.html
```diff
@@ -0,0 +1,43 @@
...
+<dom-module id="inline-details-pane">
+  <template>
+    <style include="datalab-shared-styles">
+      #container {
+        height: 100%;
+        overflow-y: auto;
+        display: flex;
+        flex-direction: column;
+      }
+      .placeholder {
```
yebrahim: Where is this used?

sources/web/datalab/polymer/test/table-inline-details-test.ts
```diff
@@ -0,0 +1,159 @@
...
+          assert(field.v === dataFields[f].innerText, 'field data does not match');
+        });
+      });
+
+      done();
+    });
+
+    testFixture.tableId = 'pid:did.tid';
+  });
+
```
yebrahim: I'd add another test for when data fails to load.

jimmc: The code doesn't really do anything when the data fails to load, so I'm not sure what I would be testing for. I added a TODO to add some tests once the code handles errors better.

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts
```diff
@@ -0,0 +1,127 @@
...
+        .then(() => this._busy = false);
+    } else {
+      this._table = null;
+      this._rows = [];
+    }
+  }
+
+  // TODO: Consider adding expanders and nested tables to make the schema viewer
+  // narrower
+  _flattenFields(fields: gapi.client.bigquery.Field[]) {
```
yebrahim: Can we move this to `Utils` or find another common place? You could even make the one in `TablePreviewElement` static and reuse it here, just make sure you import that element's HTML.

sources/web/datalab/polymer/test/item-list-test.ts
yebrahim: I'm worried about flakiness, because this call makes sure the render completes before proceeding.

sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts
jimmc: Done.

jimmc:
Added calls to Polymer.dom.flush after each click in the item-list tests that might possibly change the DOM. The tests worked previously, and work with this change, but maybe if there could have been flakiness this will reduce that probability.

PR #1671, opened 2017-09-15
Implement github file manager.

jimmc:
This just implements the happy path for browsing github projects and reading files. It doesn't check for errors and it doesn't attempt to deal with paging. The default page size is 30 items, which can be bumped up to 100 with a query parameter. But as is, it can be used with the file browser to browse github projects and repositories and to open text files and notebooks.

### DIFF
--- a/sources/web/datalab/polymer/modules/github-file-manager/github-file-manager.ts
+++ b/sources/web/datalab/polymer/modules/github-file-manager/github-file-manager.ts
@@ -59,12 +59,10 @@ class GithubFileManager implements FileManager {
   private _githubApiManager: GithubApiManager;
 
   public get(fileId: DatalabFileId): Promise<DatalabFile> {
-    const pathParts = fileId.path.split('/').filter((part) => !!part);
-    if (pathParts.length === 0) {
+    if (fileId.path === '' || fileId.path === '/') {
       return Promise.resolve(this._ghRootDatalabFile());
     }
-    const githubPath = '/repos/' + pathParts.slice(0,2).join('/') +
-        '/contents/' + pathParts.slice(2).join('/');
+    const githubPath = this._githubPathForFileId(fileId, 'get');
     return this._githubApiPathRequest(githubPath)
         .then((response: GhFileResponse) => {
           return this._ghFileToDatalabFile(response);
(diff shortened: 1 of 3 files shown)


sources/web/datalab/polymer/modules/github-file-manager/github-file-manager.ts
```diff
@@ -0,0 +1,268 @@
...
+    const pathParts = path.split('/').filter((part) => !!part);
+    const files: DatalabFile[] = [];
+    for (let p = 0; p < pathParts.length; p++) {
+      files[p] = this._ghPathPartsToDatalabFile(pathParts.slice(0, p + 1));
+    }
+    return files;
+  }
+
+  // We don't know if the tyep of the item is actually a directory without
+  // querying the github API, so we assume every component is a dir.
```
yebrahim: This isn't always going to be correct though right? I'd add a TODO here at least.

sources/web/datalab/polymer/modules/github-file-manager/github-file-manager.ts
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  private _ghRootDatalabFile(): DatalabFile {
+    const path = '/';
+    return new GithubFile({
+      icon: '',
+      id: new DatalabFileId(path, FileManagerType.GITHUB),
+      name: '/',
+      status: DatalabFileStatus.IDLE,
+      type: DatalabFileType.FILE,
```
yebrahim: Why isn't this `DIRECTORY`?

sources/web/datalab/polymer/modules/github-file-manager/github-file-manager.ts
```diff
@@ -0,0 +1,268 @@
...
+    if (file.encoding != 'base64') {
+      throw new Error('github file encoding "' + file.encoding +
+        '" is not supported');
+    }
+    return atob(file.content);
+  }
+}
+
+// We just want the sendRequestAsync method of BaseApimanager.
+class GithubApiManager extends BaseApiManager {
```
yebrahim: Do we need this if we specify `false` for `prependBasepath` when calling `sendRequestAsync`?

sources/web/datalab/polymer/modules/github-file-manager/github-file-manager.ts
```diff
@@ -0,0 +1,268 @@
...
+/**
+ * A file manager that wraps the Github API so that we can browse github
+ * repositories.
+ */
+class GithubFileManager implements FileManager {
+
+  private _githubApiManager: GithubApiManager;
+
+  public get(fileId: DatalabFileId): Promise<DatalabFile> {
+    const pathParts = fileId.path.split('/').filter((part) => !!part);
```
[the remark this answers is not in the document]
jimmc: Done.

jimmc:
We don't need a client id to access public stuff. Github does request that we set User-Agent; I made a change to that code just now and added comments.
The only usage limits I see discussed on their API doc page is rate-limiting. I haven't tried to trigger that to see what happens.

PR #1680, opened 2017-09-18
Wait on uploading user creds

yebrahim:
Wait until the upload is complete and send an ack message back so the notebook editor can wait on the upload.

### DIFF
--- a/sources/web/datalab/polymer/notebook.ts
+++ b/sources/web/datalab/polymer/notebook.ts
@@ -31,7 +31,7 @@ const CommandId = {
 
 interface IframeMessage {
   command?: string;
-  arguments: any;
+  arguments?: any;
   guid?: string;
 }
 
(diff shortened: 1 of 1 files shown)


sources/web/datalab/polymer/notebook.ts
```diff
@@ -44,7 +44,13 @@ async function processMessageEvent(e: MessageEvent) {
...
 
   if (message.command === CommandId.UPLOAD_USER_CREDS) {
-    ApiManagerFactory.getInstance().uploadOauthAccessToken();
+    try {
+      await ApiManagerFactory.getInstance().uploadOauthAccessToken();
+    } catch (e) {
+      message.command = CommandId.ERROR;
+      message.arguments = e;
+    }
+    sendMessageToNotebookEditor(message);
```
chmeyers: Should this be inside the catch?

yebrahim: We want to send an ack always, the catch is just changing the command to error and sending the error along.

jimmc: You are essentially reusing 'message' for a completely different purpose. I think the intent of the code would be more clear if you did this:
try {
  await ApiManagerFactory...
  sendMessageToNotebookEditor(message);
} catch (e) {
  const errorMessage = {
    command: CommandId.ERROR;
    arguments: e
  };
  sendMessageToNotebookEditor(errorMessage);
  return;
}

yebrahim: Done. PTAL.

PR #1684, opened 2017-09-20
Add file source dropdown to file browser

yebrahim:
This PR adds the ability to switch the file browser's source using a dropdown. In order to achieve this, it does the following:
- Adds an app setting for the list of supported file sources, which gets read on startup.
- Adds icons for: local disk (jupyter), drive, bigquery, and github.
- Adds a dropdown, similar to the file browser's toolbar dropdowns that appear when the window is too narrow. In this it also fixes an issue where the dropdown doesn't dismiss after clicking a button, by listening on the blur event.
- Adds a `SharedDriveFileManager` that extends `DriveFileManager` and changes the gapi query appropriately.
- Cleans up the `FileManagerFactory`'s methods for getting and converting file manager types.

image

The dropdown appears as a single unselectable div when there's only one source:
image

### DIFF
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -302,7 +302,9 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
   async _fileManagerTypeListChanged() {
     if (!this.fileManagerTypeList) {
       const settings = await SettingsManager.getAppSettingsAsync();
-      this.fileManagerTypeList = settings.supportedFileBrowserSources.map((source) =>
+      // Fall back to Jupyter if settings are somehow outdated.
+      const types = settings.supportedFileBrowserSources || ['jupyter'];
+      this.fileManagerTypeList = types.map((source) =>
           FileManagerFactory.fileManagerNameToType(source));
     }
 
@@ -927,7 +929,10 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
   }
 
   _toggleFileSourceDropdown() {
-    if (this._hasMultipleFileSources) {
+    // Only toggle the dropdown open if there are multiple file sources. We do
+    // this so that it looks like a static div instead of a dropdown if only
+    // once source is supported.
+    if (this._hasMultipleFileSources || this.$.fileSourcesDropdown.opened) {
       this.$.fileSourcesDropdown.toggle();
     }
   }
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -313,15 +313,13 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
     Utils.deleteAllChildren(menu);
 
     this.fileManagerTypeList.forEach((type) => {
-      const config = FILE_MANAGER_CONFIG.get(type);
-      if (!config) {
-        throw new Error('Unknown FileManagerType: ' + type.toString());
-      }
+      const config = FileManagerFactory.getFileManagerConfig(type);
+      const strType = FileManagerFactory.fileManagerTypetoString(type);
       const btn = document.createElement('paper-button');
       btn.classList.add('toolbar-button');
       btn.addEventListener('click', () => {
-        if (this.fileManagerType !== type) {
-          this.fileManagerType = type;
+        if (this.fileManagerType !== strType) {
+          this.fileManagerType = strType;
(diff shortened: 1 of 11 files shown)


sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -285,6 +299,49 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
...
+      if (!config) {
+        throw new Error('Unknown FileManagerType: ' + type.toString());
+      }
+      const btn = document.createElement('paper-button');
+      btn.classList.add('toolbar-button');
+      btn.addEventListener('click', () => {
+        if (this.fileManagerType !== type) {
+          this.fileManagerType = type;
+          this._fileManager = FileManagerFactory.getInstanceForType(type);
+          this.fileId = '';
```
jimmc: Might be nice to store the last location used with this file manager, so that if you switch from, say My Drive to Shared Drive and back to My Drive, you will end up in the same place in My Drive as when you left it. But that could be in a future PR.

sources/web/datalab/polymer/modules/file-manager-factory/file-manager-factory.ts
```diff
@@ -12,38 +12,69 @@
...
+  BIG_QUERY = 'bigquery',
+  DRIVE = 'drive',
+  GITHUB = 'github',
+  JUPYTER = 'jupyter',
+  MOCK = 'mock',
+  SHARED_DRIVE = 'sharedDrive',
+}
+
+interface FileManagerConfig {
+  className: any;
```
jimmc: It looks like you are using this as a class rather than a name.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -285,6 +299,49 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
...
+      this.fileManagerTypeList = types.map((source) =>
+          FileManagerFactory.fileManagerNameToType(source));
+    }
+
+    const menu = this.$.fileSourcesDropdown as HTMLDivElement;
+    // Build and add buttons to the switcher dropdown menu
+    Utils.deleteAllChildren(menu);
+
+    this.fileManagerTypeList.forEach((type) => {
+      const config = FILE_MANAGER_CONFIG.get(type);
```
jimmc: This feels kind of like directly accessing the internals of a global. I think I might rather see this encapsulated in a method within FileManagerFactory.

yebrahim: Agreed. Done.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
yebrahim: Yes I thought about that. The `_loadStartupPath` method should take care of loading the last accessed path for the selected file manager.

sources/web/datalab/polymer/modules/file-manager-factory/file-manager-factory.ts
yebrahim: I hadn't finished my research on how class can be represented as type. Done.

sources/web/datalab/polymer/modules/file-manager-factory/file-manager-factory.ts
```diff
@@ -22,67 +22,65 @@ enum FileManagerType {
 }
 
 interface FileManagerConfig {
-  className: any;
+  className: new () => FileManager;
```
jimmc: This field is still not a name, but a class. Maybe call it managerClass?

yebrahim: Done, renamed to `class`.

sources/web/datalab/polymer/modules/file-manager-factory/file-manager-factory.ts
```diff
@@ -22,7 +22,7 @@ enum FileManagerType {
 }
 
 interface FileManagerConfig {
-  className: new () => FileManager;
+  class: new () => FileManager;
```
jimmc: I'm surprised this compiles, I expected that class is a reserved word. Apparently it is not, but in my mind it kind of is. For clarity, I think it should be something else. But if you don't share that concern, you can go ahead and submit as is.

yebrahim: Yea you can do that, but I mostly share your concern. Renamed to `typeClass`.

PR #1687, opened 2017-09-21
Get session file names from their file managers

yebrahim:
This change tries to get the file manager from the session path if it's a valid DatalabFileId, and get the file's name and icon from that, otherwise defaults to the old behavior.

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
+++ b/sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
@@ -34,7 +34,7 @@ class SessionsElement extends Polymer.Element implements DatalabPageElement {
 
   public resizeHandler = null;
 
-  private _sessionList: Session[];
+  private _sessionList: Session[] = [];
   private _fetching: boolean;
   private _sessionListRefreshInterval = 60 * 1000;
   private _sessionListRefreshIntervalHandle = 0;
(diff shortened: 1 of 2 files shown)


sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
```diff
@@ -74,17 +80,46 @@ class SessionsElement extends Polymer.Element implements DatalabPageElement {
...
+    const sessionDescriptions = this._sessionList.map((session) => {
+      // Try to parse the given path. If it's a valid DatalabId, extract the
+      // file manager from it, and get the file name and icon from that.
+      // Otherwise, assume it's a Jupyter file.
+      try {
+        const id = DatalabFileId.fromQueryString(session.notebook.path);
+        const type = FileManagerFactory.fileManagerNameToType(id.source);
+        const config = FileManagerFactory.getFileManagerConfig(type);
+        const fileManager = FileManagerFactory.getInstanceForType(type);
+        return fileManager.get(id)
```
jimmc: Since you are in an async function now, can you use await fileManager.get here?

sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
```diff
@@ -74,17 +80,46 @@ class SessionsElement extends Polymer.Element implements DatalabPageElement {
...
+            return description;
+          });
+      } catch (e) {
+        const config = FileManagerFactory.getFileManagerConfig(FileManagerType.JUPYTER);
+        const description: SessionDescription = {
+          icon: config.displayIcon,
+          kernel: session.kernel.name,
+          name: session.notebook.path,
+        };
+        return Promise.resolve(description);
```
jimmc: Does it work to just return description here?
Maybe refactor the body of the map function into something like "async sessionToSessionDescription".

sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
```diff
@@ -74,17 +80,46 @@ class SessionsElement extends Polymer.Element implements DatalabPageElement {
...
+        return fileManager.get(id)
+          .then((file) => {
+            const description: SessionDescription = {
+              icon: config.displayIcon,
+              kernel: session.kernel.name,
+              name: file.name,
+            };
+            return description;
+          });
+      } catch (e) {
```
jimmc: You are not intending to catch errors from the above promises, are you? Having so much stuff in the try clause makes it hard to figure out what you are actually trying to catch for.

sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
```diff
@@ -74,17 +80,46 @@ class SessionsElement extends Polymer.Element implements DatalabPageElement {
...
+        const config = FileManagerFactory.getFileManagerConfig(FileManagerType.JUPYTER);
+        const description: SessionDescription = {
+          icon: config.displayIcon,
+          kernel: session.kernel.name,
+          name: session.notebook.path,
+        };
+        return Promise.resolve(description);
+      }
+    });
+    const sessions = await Promise.all(sessionDescriptions);
```
jimmc: If I understand correctly, the naming of these variables would more accurately be something like
  const sessionDescriptions = await Promise.all(sessionDescriptionPromises)

yebrahim: Fixed.

PR #1691, opened 2017-09-22
Update to using ws@1.1.4 to avoid websockets/ws:778

fischman:
https://github.com/websockets/ws/issues/778 can cause ws to raise, terminating the python runtime.  Upgrading ws avoids this.

b/66429362

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -183,9 +183,8 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
 # Add some unchanging bits - specifically node modules (that need to be kept in sync
 # with packages.json manually, but help save build time, by preincluding them in an
 # earlier layer).
-# Note: ws is now over 1.0 but using that gives issues so leaving at 0.4.2 for now.
     /tools/node/bin/npm install \
-        ws@0.4.32 \
+        ws@1.1.4 \
         http-proxy@1.13.2 \
         mkdirp@0.5.1 \
         node-uuid@1.4.7 \
--- a/containers/base/Dockerfile.py2
+++ b/containers/base/Dockerfile.py2
@@ -110,9 +110,8 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
 # Add some unchanging bits - specifically node modules (that need to be kept in sync
 # with packages.json manually, but help save build time, by preincluding them in an
 # earlier layer).
-# Note: ws is now over 1.0 but using that gives issues so leaving at 0.4.2 for now.
     /tools/node/bin/npm install \
-        ws@0.4.32 \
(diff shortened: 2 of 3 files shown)


containers/base/Dockerfile
```diff
@@ -183,9 +183,8 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
 # Add some unchanging bits - specifically node modules (that need to be kept in sync
 # with packages.json manually, but help save build time, by preincluding them in an
 # earlier layer).
-# Note: ws is now over 1.0 but using that gives issues so leaving at 0.4.2 for now.
```
yebrahim: This is worrying. @qimingj or @chmeyers, do you guys have context about what sort of issues were seen with the update? If not we'll have to do some testing.

containers/base/Dockerfile
```diff
@@ -183,9 +183,8 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
...
     /tools/node/bin/npm install \
-        ws@0.4.32 \
+        ws@1.1.4 \
```
yebrahim: We'll also need to update the Dockerfile.py2 file.

containers/base/Dockerfile
qimingj: I think @nikhilk did that and added that comments. I am fine with it with some manual testing.

chmeyers: gramster@ originally added that comment in https://github.com/googledatalab/datalab/commit/e8659d9b6124f9a17c4ffedb4f749cbb8c257db2 (April of 2016)

fischman: FTR I did manually test with this change through the socket.io codepath and things seem to be more robust (see b/ link in original PR).

qimingj: Maybe that explains why we had some stability issues with socket io.

chmeyers: fischman@: Do you know if we are going to hit any of the "Backwards Incompatible Changes" listed in http://channels.readthedocs.io/en/latest/releases/1.0.0.html ?

If we're not sure, I'm okay with upgrading it and fixing any issues we find.

fischman: @chmeyers I don't _know_, but I expect this to be all fine as the motivating ws issue contains evidence of people using socket.io 1.7.x with ws 1.1.x (and we don't use ws directly).

chmeyers: Okay.  Update Dockerfile.py2 too, and I'll approve/merge.

containers/base/Dockerfile
fischman: Done.

containers/base/Dockerfile
fischman: Done.

PR #1694, opened 2017-09-22
Use datalab subdir as default for open-in-notebook when on jupyter:.

jimmc:


sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -95,9 +95,14 @@ class TemplateManager {
...
 
     // TODO(jimmc): Look for a user preference for baseDir
-    const baseDir = (appSettings.defaultFileManager || 'drive') + ':';
+    let baseDir = (appSettings.defaultFileManager || 'drive') + ':';
     // TODO(jimmc): Allow specifying a path with baseDir. For now, we are
     // just using the root of the filesystem as the default location.
+    if (baseDir === 'jupyter:') {
+      // Jupyter root directory does not persist between container restarts,
+      // so we use the datalab dir as the default in this case.
+      baseDir = baseDir + 'datalab/';
```
yebrahim: I think we want `/content` here right?

jimmc: There is no /content in the jupyter: hierarchy that is visible in the UI. The home folder only contains datalab, which contains docs and notebooks.

yebrahim: That's right. But `/content` is actually just `jupyter:`, you're saying this gets rewritten between sessions? This shouldn't happen unless you get a new disk right?

jimmc: For an instance on a VM, /content (and thus jupyter:) is preserved on the PD between stops and restarts of the VM. I ran into this problem doing local development, in which case only /content/datalab is preseved, the rest of /content is lost between restarts. But maybe we don't care about that.
What do you think, should I just close this pull request and leave it at just jupyter: ?

yebrahim: Yes I think we should write to the root. The user can theoretically delete `/content/datalab`.
For local development, you can change the run script to mount `/content` instead of `/content/datalab`.

jimmc: OK, I'll close this PR.

PR #1696, opened 2017-09-25
Refactor daas/jupyter api managers into one

yebrahim:
There's no longer need for having two different functionalities for how to get and parse the basepath. This PR moves to having a single `ApiManager` class with static methods that can be called directly.


sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.html
```diff
@@ -13,7 +13,7 @@
 -->
 
 <link rel="import" href="../../modules/utils/utils.html">
-<link rel="import" href="../api-manager-factory/api-manager-factory.html">
+<link rel="import" href="../api-manager/api-manager.html">
```
jimmc: Unused? ApiManager appears only in a comment in drive-file-manager.ts (which probably should be fixed).

yebrahim:
Comments done, PTAL.

PR #1697, opened 2017-09-26
Modify how notebook templating works.

jimmc:
This PR changes how notebook templating (for open-in-notebook command) works:

* Instead of doing string replacement when instantiating a template, we add an initial cell that defines a set of parameter values. This is cleaner, and doesn't require learning any new syntax to create a template notebook. It has the downside that variables can not be used in markdown cells; for now, we can generate markdown output in a code cell.
* There are now separate variable values defined for the three pieces of the full BigQuery table id, along with the fully qualified id.
* The template file is now specified with a fileId string, to make it easier to change that location. For debugging for now, that location can be overridden.
* The default name of the template file has been changed to BigQueryTableOverview.
* The template no longer needs to be on the same file manager as the created notebook.


sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -36,44 +34,47 @@ class NotebookTemplate {
...
-        const escapedPlaceholder = NotebookTemplate._regexEscapeAll(placeholder);
-        const regex = new RegExp(escapedPlaceholder);
-        const value = '\'' + parameter.value.toString() + '\'';
-        cell.source = cell.source.replace(regex, value);
-      });
-    });
-
+  public addParameterCell(notebook: NotebookContent) {
+    const definitionLines = this.parameters.map(parameter =>
+        parameter.placeholder + ' = ' + JSON.stringify(parameter.value));
```
yebrahim: Should we rename this to `parameter.name`?

sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -36,44 +34,47 @@ class NotebookTemplate {
...
+    const header = '# Parameter definitions generated by code';
```
yebrahim: Can we make the message a little clearer, maybe "Auto-generated parameter definitions"?

yebrahim: Also, do we want this to be a header? I would say this cell shouldn't stand out too much, so as not to interfere with the notebook's flow for the reader.

sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -36,44 +34,47 @@ class NotebookTemplate {
...
  * This template contains one cell that shows the given table's schema.
  */
-class TableSchemaTemplate extends NotebookTemplate {
-  constructor(tableName: string) {
-    const parameters = [{
-      placeholder: 'TABLE_NAME_PLACEHOLDER',
-      value: tableName,
-    }];
+class BigQueryTableOverviewTemplate extends NotebookTemplate {
+  constructor(dict: { [key: string]: string }) {
```
yebrahim: Are string parameters the only type we support? Perhaps we should make this `any`, or one of a defined set of types?

sources/web/datalab/polymer/modules/template-manager/template-manager.ts
jimmc: Done.

sources/web/datalab/polymer/modules/template-manager/template-manager.ts
jimmc: Message updated.
I thought having a comment indicating generated code was a good thing to have. What would you suggest?

sources/web/datalab/polymer/modules/template-manager/template-manager.ts
jimmc: Done.

sources/web/datalab/polymer/modules/template-manager/template-manager.ts
yebrahim: Oh, I mistook this for a separate markdown cell, so I thought this would be a level-0 header. Never mind my comment.

PR #1700, opened 2017-09-26
Use template placeholders if there instead of adding parameter cell.

jimmc:


sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -34,6 +43,39 @@ class NotebookTemplate {
   }
 
   /**
+   * Escapes all regex modifier characters in the given string.
+   * Inserts a new cell at position zero that defines python variables
```
yebrahim: Merge issue with comment?

jimmc: Yup. Fixed.

PR #1702, opened 2017-09-27
Load notebook template through our API.

jimmc:


sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -93,6 +100,15 @@ class NotebookTemplate {
   }
 }
 
+// A small hack to allow us to have a unified way of expressing
+// a file location on any of Drive, Jupyter, or our API.
+class ApiFileId extends DatalabFileId {
+  constructor(queryString: string) {
+    const tokens = queryString.split(':');
```
yebrahim: Throw for bad `queryString` format?

sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -93,6 +100,15 @@ class NotebookTemplate {
   }
 }
 
+// A small hack to allow us to have a unified way of expressing
+// a file location on any of Drive, Jupyter, or our API.
+class ApiFileId extends DatalabFileId {
```
yebrahim: I don't understand why we need this. Why not just use a `DatalabFileId`? It'll gladly accept `api` as the source, we just need to add it as a `FileManagerType`.

jimmc: Done.

sources/web/datalab/polymer/modules/template-manager/template-manager.ts
jimmc: Removed along with the previous change.

PR #1708, opened 2017-09-29
Rework notebook template load to use static mechanism rather than api.

jimmc:


sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -204,10 +209,12 @@ class TemplateManager {
     }
   }
 
-  public static async getTemplateStringContent(fileId: DatalabFileId) {
-    if (fileId.source === FileManagerType.API) {
+  public static async getTemplateStringContent(fileId: DatalabFileId,
+      resolver: Polymer.Element) {
+    if (fileId.source === FileManagerType.STATIC) {
+      const resolvedUrl = resolver.resolveUrl('../..' + fileId.path);
```
yebrahim: Can we use the static class method `Polymer.ResolveUrl.resolveUrl` instead, since this isn't really related to a Polymer element? This way we don't need to extend `Polymer.Element` or pass down the resolver instance.

jimmc: I spent some time looking into this. The problem with Polymer.ResolveUrl.resolveUrl is that it requires a baseUri, and I don't have a simple way of getting that. I think I can probably write a bunch of ugly code that gets it from the dom, but that doesn't seem like a good tradeoff versus this code. Maybe I am missing an easy way - if so, please let me know.

yebrahim: Sorry I didn't have time to take another look at this. We don't need a baseUrl here, so passing an empty string should work. It'll clean up the code quite a bit so I think it's worth fixing now. I can send another PR to fix this if you like.

yebrahim: Also, it seems the [baseUrl is optional](https://github.com/Polymer/polymer/blob/master/lib/utils/resolve-url.html#L26), which makes sense.

PR #1709, opened 2017-09-29
Allow datalab and jupyter to start up when missing a home dir.

chmeyers:
This will support attaching a home directory on another disk after
start.

### DIFF
--- a/sources/web/datalab/jupyter.ts
+++ b/sources/web/datalab/jupyter.ts
@@ -186,9 +186,17 @@ function createJupyterServerAtPort(port: number, userId: string, userDir: string
 function createJupyterServer(userId: string, remainingAttempts: number) {
   logging.getLogger().info('Checking user dir for %s exists', userId);
   var userDir = userManager.getUserDir(userId);
+  logging.getLogger().info('Checking dir %s exists', userDir);
   if (!fs.existsSync(userDir)) {
     logging.getLogger().info('Creating user dir %s', userDir);
-    fs.mkdirSync(userDir, parseInt('0755', 8));
+    try {
+      fs.mkdirSync(userDir, parseInt('0755', 8));
+    } catch (e) {
+      // This likely means the disk is not yet ready.
+      // We'll fall back to /content for now.
+      logging.getLogger().info('User dir %s does not exist', userDir);
+      userDir = '/content'
+    }
   }
 
   nextJupyterPort = appSettings.nextJupyterPort;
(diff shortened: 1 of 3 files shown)


sources/web/datalab/jupyter.ts
```diff
@@ -186,9 +186,17 @@ function createJupyterServerAtPort(port: number, userId: string, userDir: string
...
   if (!fs.existsSync(userDir)) {
     logging.getLogger().info('Creating user dir %s', userDir);
-    fs.mkdirSync(userDir, parseInt('0755', 8));
+    try {
+      fs.mkdirSync(userDir, parseInt('0755', 8));
+    } catch (e) {
+      // This likely means the disk is not yet ready.
+      // We'll fall back to /content for now.
+      logging.getLogger().info('User dir %s does not exist', userDir);
+      userDir = '/content'
```
yebrahim: This doesn't handle multi-user support like the rest of the code, but I guess we don't care about that now?

sources/web/datalab/settings.ts
```diff
@@ -116,16 +116,24 @@ export function getUserConfigDir(userId: string): string {
...
+  return mergedUserSettings;
+}
+
+/**
+ * Copies the default user settings into the user's directory.
+ */
+function copyDefaultUserSettings(userId: string) {
+  var userSettingsPath = path.join(getUserConfigDir(userId), SETTINGS_FILE);
+  _log('Copying default settings to: ' + userSettingsPath);
+  fs.writeFileSync(userSettingsPath, getDefaultUserSettings(userId));
```
yebrahim: If this can now throw, we should either handle it here or make sure all callers catch. For example, the update function also tries to copy the settings file to the user's directory, which will effectively crash the server if the disk isn't mounted yet.

Do we want to support the same functionality in the case where the disk is mounted at an arbitrary point in time, or do we just want the server not to fail, but it's ok if we have a more basic behavior?

chmeyers: It doesn't throw anything new.  We're just catching an existing failure below.

It's okay to have more basic behavior.

sources/web/datalab/jupyter.ts
chmeyers: Yes, I don't care about it in this case.

sources/web/datalab/settings.ts
chmeyers: Oh, I see the other use, catching there, too.

PR #1710, opened 2017-10-02
Fix showing preview placeholder when no items selected

yebrahim:
The placeholder element is common to all preview elements, so the container div containing both the elements and the placeholder shouldn't be hidden.

### DIFF
--- a/sources/web/datalab/polymer/test/table-preview-test.ts
+++ b/sources/web/datalab/polymer/test/table-preview-test.ts
@@ -74,13 +74,6 @@ describe('<table-preview>', () => {
     Polymer.dom.flush();
   });
 
-  it('displays an empty element if no table is provided', () => {
-    assert(!testFixture.$.placeholder.hidden, 'placeholder should be shown');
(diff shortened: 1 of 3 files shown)


sources/web/datalab/polymer/test/table-preview-test.ts
```diff
@@ -74,13 +74,6 @@ describe('<table-preview>', () => {
     Polymer.dom.flush();
   });
 
-  it('displays an empty element if no table is provided', () => {
-    assert(!testFixture.$.placeholder.hidden, 'placeholder should be shown');
```
jimmc: Was this test not working? Should it be fixed and moved somewhere else instead of being deleted?

yebrahim: I deleted it because I think it shouldn't be testing this; neither is there a placeholder in table-preview now, nor should the container be hidden any more.
The placeholder should ideally be tested in a preview-pane-test, which we don't have yet.

PR #1712, opened 2017-10-02
Bundling polymer resources by default

yebrahim:
This PR adds a few dependency behavior elements that only surfaced when bundling the resources, and adds bundling as the default behavior of the build system. A new `-d | --debug` option to the web sources build script disables that to help with debugging.

Note that the sources are still not minified. For some reason, this breaks the code, arbitrarily removing some class definitions. We can look into that later.

This helps bring down the number of requests to first meaningful paint significantly (201 -> 24). Load time on my local machine went down from 1.5s to 1.2s, should be better when server over network.


sources/web/build.sh
```diff
@@ -32,7 +44,12 @@ mkdir -p $WEB_DIR
 # Experimental UI build step
 cd datalab/polymer
 npm run build
-rsync -avpq ./build/experimental/ ../static/experimental
+if [[ $DEBUG == 1 ]]; then
+  rsync -avpq ./build/polymer_unbundled/ ../static/experimental
+else
+echo "Using bundled polymer resources.."
+  rsync -avpq ./build/polymer_bundled/ ../static/experimental
```
chmeyers: Just to double check, this means that everything is still getting placed into /datalab/web/static/experimental, right?  We have some references to that directory, so let me know if that is not true.

yebrahim: Yup, this doesn't change the end result, just the intermediate copy step.

PR #1713, opened 2017-10-03
Allow specifying the network name for Datalab instances.

### LINKED ISSUE #1535: Any way to prevent Datalab from connecting/creating its own separate [datalab] network, and force it to connect to the default vpc?
I've been looking at the docs, and there seems to be no way to create a datalab instance within the default VPC. It's really causing issues with interoperability

ojarjur:
This change also adds a warning when creating an instance if the
network has firewall rules other than the '...-allow-ssh' rule that
is added by the tool.

That new warning is meant to prevent users from accidentally creating
an instance in a network with too many sources/types of inbound traffic
allowed.

This fixes #1535


tools/cli/commands/create.py
```diff
@@ -36,8 +44,12 @@
 _DATALAB_NETWORK = 'datalab-network'
 _DATALAB_NETWORK_DESCRIPTION = 'Network for Google Cloud Datalab instances'
 
-_DATALAB_FIREWALL_RULE = 'datalab-network-allow-ssh'
+_DATALAB_FIREWALL_RULE_TEMPLATE = '{0}-allow-ssh'
 _DATALAB_FIREWALL_RULE_DESCRIPTION = 'Allow SSH access to Datalab instances'
+_DATALAB_FIREWALLS_WARNING_TEMPLATE = ('The network `{0}` has firewall rules '
```
jimmc: Should this be FIREWALL rather than FIREWALLS? I think there is still only one firewall.

tools/cli/commands/create.py
```diff
@@ -358,80 +383,108 @@ def flags(parser):
...
+    if has_non_standard_firewall_rules(args, gcloud_compute, network_name):
+        print(_DATALAB_FIREWALLS_WARNING_TEMPLATE.format(network_name))
+        resp = read_input('Do you still want to use this network? (y/[n]): ')
+        if len(resp) < 1 or (resp[0] != 'y' and resp[0] != 'Y'):
+            raise CancelledException()
+    return
+
+
+def ensure_firewall_rule_exists(args, gcloud_compute, network_name):
+    """Create the `{network_name}-allow-ssh` firewall rule if it necessary.
```
jimmc: Looks like a word is missing in this comment.

tools/cli/commands/create.py
```diff
@@ -358,80 +383,108 @@ def flags(parser):
...
     except subprocess.CalledProcessError:
-        create_network(args, gcloud_compute)
+        create_network(args, gcloud_compute, network_name)
     return
 
 
-def create_firewall_rule(args, gcloud_compute):
-    """Create the `datalab-network-allow-ssh` firewall rule.
+def create_firewall_rule(args, gcloud_compute, network_name, rule_name):
+    """Create the `{rule_name}` firewall rule.
```
jimmc: The previous comment included the rule name, from which one can understand in general what that rule does. SInce the new comment no longer includes that piece of information, it would be nice to add a word or two to the comment, such as 'Create the {rule_name} allow-ssh rule.'

tools/cli/commands/create.py
```diff
@@ -358,80 +383,108 @@ def flags(parser):
...
-        '--network', _DATALAB_NETWORK,
+        '--network', network_name,
         '--description', _DATALAB_FIREWALL_RULE_DESCRIPTION]
     utils.call_gcloud_quietly(args, gcloud_compute, create_cmd)
     return
 
 
-def ensure_firewall_rule_exists(args, gcloud_compute):
-    """Create the `datalab-network-allow-ssh` firewall rule if it necessary.
+def has_non_standard_firewall_rules(args, gcloud_compute, network_name):
```
jimmc: It is not obvious to me what "non-standard" means here. Maybe add a comment stating that this returns true if there are any firewall rules other than the one we expect.

tools/cli/commands/create.py
```diff
@@ -36,8 +44,12 @@
 _DATALAB_NETWORK = 'datalab-network'
 _DATALAB_NETWORK_DESCRIPTION = 'Network for Google Cloud Datalab instances'
 
-_DATALAB_FIREWALL_RULE = 'datalab-network-allow-ssh'
+_DATALAB_FIREWALL_RULE_TEMPLATE = '{0}-allow-ssh'
 _DATALAB_FIREWALL_RULE_DESCRIPTION = 'Allow SSH access to Datalab instances'
+_DATALAB_FIREWALLS_WARNING_TEMPLATE = (
```
jimmc: Should this be FIREWALL rather than FIREWALLS? I think there is still only one firewall.

tools/cli/commands/create.py
ojarjur: Changed to "unexpected". Is that more clear?

tools/cli/commands/create.py
ojarjur: Done.

tools/cli/commands/create.py
ojarjur: Done.

tools/cli/commands/create.py
ojarjur: No; this is warning about the unexpected firewalls, of which there could be multiple.

tools/cli/commands/create.py
jimmc: Yes, that's better. Thanks.

tools/cli/commands/create.py
jimmc: Hmmm, I would say it is multiple rules that make up one firewall, but not a big deal.

PR #1719, opened 2017-10-04
File browser busy loading fixes

yebrahim:
This PR checks and sets the fetching busy status of the file browser on all async operations.
It also removes loading sessions when in the small mode for the copy/move dialog.

### DIFF
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -202,9 +202,17 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
   }
 
   async _init() {
-    // Must set this to true before calling super.ready(), because the latter will cause
-    // property updates that will cause _fetchFileList to be called first, we don't want
-    // that. We want ready() to be the entry point so it gets the user's last saved path.
+    // TODO(yelsayed): We use this flag to prevent multiple async operations to
+    // step on each other. It basically acts as a semaphore; when an operation
+    // is called, it first checks if this flag is raised, and it stops if it is.
+    // Otherwise it raises it first. This result in operations stopping if
+    // others are executing, instead of queueing, which is not obviously a
+    // better experience. We might want to revisit this later.
+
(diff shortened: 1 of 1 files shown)


sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -323,6 +336,9 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
       const btn = document.createElement('paper-button');
       btn.classList.add('toolbar-button');
       btn.addEventListener('click', () => {
+        if (this._fetching) {
+          return;
```
jimmc: So if the page happens to be fetching when I click the switcher button, it just does nothing? This seems like a poor UX.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -449,6 +466,9 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
    * an effect, a directory will still navigate.
    */
   async _handleDoubleClicked(e: ItemClickEvent) {
+    if (this._fetching) {
+      return;
```
jimmc: Same concern as above. The page sometimes auto-refreshes, right? Seems like this will make the UI feel like it randomly doesn't work.
Likewise for all the other places where the UI is going to give me the cold shoulder and silently ignore what I asked it to do.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
yebrahim: Yes, it's an easy fix for the current state, which is inconsistent.
The other, more involved, option would be to queue all those requests and play them one by one,  but I'm not sure that's a great experience either, because there's no easy way to interrupt that queue.

jimmc: OK, please add a TODO.

yebrahim: Done.

PR #1721, opened 2017-10-05
Make the fake metadata address configurable.

craigcitro:
This makes it possible to configure the GCE metadata fake to serve on a
different host/port combination; the primary goal is to make it possible to
use the metadata service locally without the related Docker network
configs. (Without this, server startup will fail attempting to bind to
`metadata.google.internal`.)

PTAL @ojarjur

### DIFF
--- a/sources/web/datalab/metadata.ts
+++ b/sources/web/datalab/metadata.ts
@@ -46,9 +46,9 @@ const metadata: FakeMetadata = {
   },
 };
 
-function launchFakeServer(metadata: FakeMetadata): void {
-  const port = 80;
-  const host = 'metadata.google.internal';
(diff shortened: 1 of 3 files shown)


sources/web/datalab/metadata.ts
```diff
@@ -29,12 +29,18 @@ interface Creds {
 }
 
 interface FakeMetadata {
+  content_dir: string;
```
ojarjur: I don't think these should be added to the FakeMetadata interface, since they are attributes of the server rather than the metadata contents.

Instead, pass in the settings object to the `launchFakeServer` function, and then use the values in the settings object.

craigcitro: Done.

craigcitro:
PTAL

PR #1723, opened 2017-10-06
Use URL paths for file id, enable relative file paths

yebrahim:
This PR changes the way the new UI handles file paths in navigation and in the notebook editor, by making the file path part of the page URL, which allows relative links to work resolve properly.

### DIFF
--- a/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
+++ b/sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
@@ -43,7 +43,9 @@ class DatalabAppElement extends Polymer.Element {
    */
   public routeData: object;
 
-  // TODO: doc
+  /**
+   * Tail of the parsed route, which contains the file id in its path property.
+   */
   public routeTail: object;
 
   private _boundResizeHandler: EventListenerObject;
@@ -93,8 +95,9 @@ class DatalabAppElement extends Polymer.Element {
 
   static get observers() {
     return [
-      // We need a complex observer for changes to the routeData
-      // object's page property.
+      // We need a complex observer for changes to the routeData object's page
+      // property, and the path property of routeTail, which contains the file
+      // id.
       '_routePageChanged(routeData.page)',
       '_routeTailPathChanged(routeTail.path)',
     ];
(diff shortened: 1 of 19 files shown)


sources/web/datalab/polymer/components/datalab-app/datalab-app.ts
```diff
@@ -42,37 +37,21 @@ class DatalabAppElement extends Polymer.Element {
...
-   * Current matching result from the window.location against the
-   * root pattern. This gets re-evaluated every time the current page
-   * changes, and can be used to get the current active page's name.
+   * Current matching result from the window.location. This gets re-evaluated
+   * every time the current page changes, and can be used to get the current
+   * active page's name.
    */
   public routeData: object;
 
+  // TODO: doc
```
ojarjur: I don't think this should be left as a TODO

yebrahim: Weird! I was sure I fixed this before sending, looks like I accidentally reverted that.

sources/web/datalab/polymer/editor.ts
```diff
@@ -38,21 +38,15 @@ if (editorElement) {
...
-    const url = location.protocol + '//' + location.host + location.pathname +
-                '?' + decodeURIComponent(params.toString());
+    const url = location.protocol + '//' + location.host + '/editor/' + (newId || '');
     window.history.replaceState({}, '', url);
   });
 
   // Pass the file's path if it's specified in the location.
-  const params = new URLSearchParams(window.location.search);
-  if (params.has('file')) {
+  if (location.pathname.startsWith('/editor/')) {
```
jimmc: Might be worth adding something like "const prefix = '/editor/';" to avoid having the same string twice. Likewise for /notebook/ in the other file.

nikhilk: Can we have resource-based URIs -- so `/notebooks/` and `/files/`?

yebrahim: How does the user specify how to handle the resource though? Whether it should be downloaded, opened in the text editor, or opened in the notebook editor?

sources/web/datalab/polymer/modules/utils/utils.ts
```diff
@@ -17,6 +17,8 @@
  */
 class Utils {
 
+  static uiroot = '';   // Gets set to "/exp"
```
jimmc: Is this still being used anywhere?

yebrahim: Removed.

PR #1733, opened 2017-10-10
Set default Data tab start location to bigquery-public-data project.

jimmc:
This PR changes the default starting location of the Data tab to be the bigquery-public-data project. This is what it will display when the user first clicks the Data tab (or enters the /data URL).
start-bigquery-public-data

### DIFF
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -239,7 +239,7 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
       this._pathHistoryIndex = 0 + this.nLeadingBreadcrumbsToTrim;
     });
 
-    let fileId = this._getFileIdFromProperty();
+    const fileId = this._getFileIdFromProperty();
     if (fileId) {
       this.fileManagerType = FileManagerFactory.fileManagerTypetoString(fileId.source);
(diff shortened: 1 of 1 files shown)


sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -260,6 +260,12 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
       }
     }
 
+    if (!fileId && this.fileManagerType === 'bigquery') {
```
yebrahim: Is this the best place for this? Can we maybe set it in the data browser itself, or in the bigquery gapi module?
I'll approve to unblock in case you want to move forward with this.

jimmc: I admit this seems like a bit of a hack. I could instead add a method such as defaultStartLocation() to the FileManager interface, and make BigQueryFileManager return this location. Would that be better?

yebrahim: How about adding this to the `_loadStartupPath` instead? That method should change to accommodate things like this anyway, because if we couldn't get the user's last saved location, we'll need a default there.

jimmc: Done.

PR #1735, opened 2017-10-10
Parallel file fetch requests, unblock file browser

yebrahim:
This PR unblocks the file browser operations while a file list fetch is taking place. It does this by:
- Making opening file/notebook a sync operation.
- Making the `_fetchFileList` call non-blocking, so that it can be called while loading the file list. It follows from this that multiple calls to this method can be made, which is a nice addition.

This PR also cleans up the `FileManager` classes by making them extend a base class instead of implementing an interface. This way they can all use the same functions for opening in editors, except for the ones that need to override it.

Not this doesn't fix another bug with double clicking items, where the breadcrumbs can go out of sync with the file list.

### DIFF
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -487,10 +487,7 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
    * If this element is in "small" mode, double clicking a file does not have
    * an effect, a directory will still navigate.
    */
-  async _handleDoubleClicked(e: ItemClickEvent) {
-    if (this._fetching) {
-      return;
-    }
+  _handleDoubleClicked(e: ItemClickEvent) {
     const clickedItem = this._fileList[e.detail.index];
     if (this.small && clickedItem.type !== DatalabFileType.DIRECTORY) {
       return;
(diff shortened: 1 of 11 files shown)


sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -134,15 +134,15 @@ abstract class DatalabFile {
   }
 }
 
-interface FileManager {
+abstract class BaseFileManager {
```
jimmc: I think you should just call this FileManager. Having every reference need to include "Base" seems like exposing implementation detail. You could keep the FileManager interface and make BaseFileManager implement that, but that seems unnecessary.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -151,71 +151,76 @@ interface FileManager {
...
+  abstract getRootFile(): Promise<DatalabFile>;
 
   /**
    * Saves the given string as a file's content.
    * @param file object containing information about the destination file to
    *             save to.
    * @param content string to be saved in the file
    */
-  saveText(file: DatalabFile, content: string): Promise<DatalabFile>;
+  abstract saveText(file: DatalabFile, content: string): Promise<DatalabFile>;
```
jimmc: Maybe most of these functions should have a default implementation that throws UnsupportedMethod.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -452,9 +463,11 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
           Utils.log.error(msgPrefix, e);
         }
       })
+      // TODO: This is not very accurate if multiple fetch requests are running
```
jimmc: There are still a bunch of places in the code where it does not take action if this._fetching is set. I think it will be pretty easy to get into the state where this flag is cleared while something is still running, which means your checks won't work properly and things will execute while a fetch is running. If this is a not a problem, then you should remove those checks; if it is a problem, then you need to make _fetching more accurate.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
yebrahim: Calling it `FileManager` would indicate it's a full implementation, this is why I included the `Base` prefix, especially when incorporating the below comment too. I'm moving this to a separate PR so we can continue this discussion there.

yebrahim:
Closing this in favor of https://github.com/googledatalab/datalab/pull/1739 to focus on the parallel requests change.

PR #1736, opened 2017-10-11
Use BaseFileManager for common functionality

yebrahim:
(This was originally part of https://github.com/googledatalab/datalab/pull/1735)

This PR moves the common functionality of the `FileManager` implementations into a base class, `BaseFileManager`.

### DIFF
--- a/sources/web/datalab/polymer/modules/file-manager/file-manager.ts
+++ b/sources/web/datalab/polymer/modules/file-manager/file-manager.ts
@@ -134,7 +134,7 @@ abstract class DatalabFile {
   }
 }
 
-interface FileManager {
+class BaseFileManager {
   // TODO: Consider supporting getting both the file and content objects with
   // one call.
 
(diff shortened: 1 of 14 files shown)


sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -206,16 +208,73 @@ interface FileManager {
...
    * Creates a path history from a path string.
    */
   pathToPathHistory(path: string): DatalabFile[];
 }
+
+/**
+ * Base implementation of the FileManager interface that contains common
+ * functionality for the different FileManager classes.
+ */
+class BaseFileManager {
```
jimmc: Doesn't this need to have "implements FileManager"?

yebrahim: It does. Fixed.

PR #1739, opened 2017-10-12
Parallel file operations, unblock file browser

yebrahim:
[Replaces https://github.com/googledatalab/datalab/pull/1735]

This PR unblocks the file browser operations while a file list fetch is taking place. It does this by:

- Making opening file/notebook a sync operation.
- Making the _fetchFileList call non-blocking, so that it can be called while loading the file list. It follows from this that multiple calls to this method can be made, which is a nice addition.
- Not blocking all other file operations on the fetch call, and making them a different `_busy` flag to enable the progress bar.

Note this doesn't fix another bug with double clicking items, where the breadcrumbs can go out of sync with the file list.

### DIFF
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -89,7 +89,6 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
   private _previewPaneCollapseThreshold = 600;
   private _fetching: boolean;
   private _fileList: DatalabFile[];
-  private _fileListFetchPromise: Promise<any>;
   private _fileListRefreshInterval = 60 * 1000;
   private _fileListRefreshIntervalHandle = 0;
   private _fileManager: BaseFileManager;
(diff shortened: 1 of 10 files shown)


sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -401,44 +409,60 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
   /**
    * Calls the FileManager to get the list of files at the current path, and
    * updates the _fileList property.
+   * This method can be called multiple times, and it will ignore the fetch
+   * result if the currentFile object has changed after the request was made.
```
jimmc: This comment is not quite accurate. It will ignore the fetch result if the currentFile object does not match the file for the request. It could have changed after the request was made, and then changed back, and the request won't be ignored. But perhaps that's too much detail to worry about here.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -401,44 +409,60 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
...
+    }
+
+    // Now load the sessions and update the running status of each file
+    // whose id is in the session list.
+    // We do not need to load sessions when not displaying the Status column,
+    // or if the current fild id has changed since the beginning of this request.
+    if (!hideStatus && fetchFileId === this.currentFile.id) {
+      try {
+        const sessions = await SessionManager.listSessionPaths();
         const listElement = this.$.files as ItemListElement;
```
jimmc: I think you should check for fetchFieldId===this.currentFile.id here, right after getting back the session info, and return if not set, as you do on line 439 just after getting the file list. It looks like it probably won't cause a problem anyway, but I think it would be cleaner to add that bail-out clause.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -721,7 +737,7 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
     // Wait on all upload requests before declaring success or failure.
     try {
       await Promise.all(uploadPromises);
-      this._fetching = false;
+      this._busy = false;
```
jimmc: Given that you are allowing a second operation to start when a first is running, I think it's going to be pretty easy to get this busy-flag out of sync: start an async operation, then start another one just before the first one finishes. The busy-flag will get cleared when the first one finishes, and the second one will still be running, but the busy-flag will not be set. I'm fine with leaving this for another PR, since I think the current code is already problematic, but perhaps putting in a TODO at one of these places would be appropriate. Unless I misunderstand this flag, of course.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
yebrahim: Yea, I too didn't think it was worth explaining in more details, the word `changed` captures the common meaning.
I made it a little clearer.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
yebrahim: I initially thought this would actually be desirable to have the extra update, since by this point the hard work has already been done (sessions API call). But now I think it's safer to bail out earlier, instead of having multiple calls modifying the polymer properties at the same time.
Done.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
yebrahim: Your understanding is correct, I guess we could use the counter mechanism here. I'll put a TODO for now.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -459,10 +460,13 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
     // Now load the sessions and update the running status of each file
     // whose id is in the session list.
     // We do not need to load sessions when not displaying the Status column,
-    // or if the current fild id has changed since the beginning of this request.
-    if (!hideStatus && fetchFileId === this.currentFile.id) {
+    if (!hideStatus) {
```
jimmc: I think if the fileManager.list call on line 436 throws an error, you might get to here without executing the fetchFileId check on line 440, in which case you might still want that check here before calling listSessionsPath. You could just add another of those if-not-the-right-fileid-then-return statements before this line. But not a big deal, it would just save one API call and only in an edge case.

yebrahim: Yea I think I was mainly concerned about setting the `_fileList` incorrectly, the rest is just nice-to-haves. I'll keep this as is for now.

PR #1742, opened 2017-10-12
Add BigQueryPublicFileManager.

jimmc:
... and use it in the Data browser to allow seeing public dataset projects separately from the user's projects.

As in the Files browser, the file manager, which used to display only BigQuery, is now a dropdown list that includes BigQuery and Public Datasets.
bq-public-datasets

### DIFF
--- a/sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
+++ b/sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
@@ -41,10 +41,6 @@ class BigQueryFile extends DatalabFile {
  */
 class BigQueryFileManager extends BaseFileManager {
 
-  protected myFileManagerType() {
-    return FileManagerType.BIG_QUERY;
-  }
-
   public canHostNotebooks() {
     return false;
   }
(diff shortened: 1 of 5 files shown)


nikhilk:
Like the UI label, "Public Datasets", I would suggest having code not be coupled to BigQuery here.

sources/web/datalab/polymer/modules/bigquery-file-manager/bigquery-file-manager.ts
```diff
@@ -41,6 +41,10 @@ class BigQueryFile extends DatalabFile {
  */
 class BigQueryFileManager extends BaseFileManager {
 
+  protected myFileManagerType() {
```
yebrahim: nit: I thought protected methods should appear after public ones?

jimmc: Done.

nikhilk:
To try and make my confusing comment clearer.

The UI refers to this as "Public Datasets" (thanks for picking up my earlier suggestion when this was proposed), but the code has FileManagerType.BIG_QUERY_PUBLIC and the class is named BigQueryPublicFileManager ... it would be nice (as a suggestion, again) that this be called FileManagerType.PUBLIC_DATASETS/PublicDatasetsManager).

Personally I (in general) wish code identifiers matched the concept and design so it helps code understandability/navigation etc. In similar vein, I also recall suggested File -> Content when this abstraction was generalized beyond Files.

More offline.

yebrahim:
The concept here isn't any generic public datasets though, it's BigQuery specific. The icon next to the source in the dropdown menu conveys that too.

jimmc:
My concern about using the Public Datasets name for all the underlying code is that, as yebrahim said, these datasets are all specific to BigQuery. I would rather reserve the more generic names such as Public Datasets for a possible more generic implementation in the future that can return public datasets from more sources than just BigQuery.

nikhilk:
I tend to think in terms of user-facing concepts/intents and not current implementation, since implementation always changes, but anway, like I said on all this, including original comment -- it's a suggestion.

PR #1747, opened 2017-10-14
Make Data tab no longer be hidden.

jimmc:
Previously, the Data tab would only appear after the user manually visited the /data page. With this change, the Data tab is always visible.


jimmc:
Sounds like the powers don't care about the order. Easy enough to change later.

PR #1754, opened 2017-10-17
Remove file status

yebrahim:
With this change, the file browser does not need to look up backends to get the files' running status. The session status can now only be viewed in the Sessions page.

### DIFF
--- a/sources/web/datalab/polymer/modules/file-manager/file-manager.ts
+++ b/sources/web/datalab/polymer/modules/file-manager/file-manager.ts
@@ -37,11 +37,6 @@ enum DatalabFileType {
   NOTEBOOK,
 }
 
-enum DatalabFileStatus {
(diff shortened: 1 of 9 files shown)


sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -139,12 +132,6 @@ interface FileManager {
   // one call.
 
   /**
-   * Returns true if we can host notebooks on this filesystem, and thus should
-   * display the Status column in the file browser.
-   */
-  canHostNotebooks(): boolean;
```
jimmc: Also should be removed from bigquery-file-manager.

yebrahim: Done.

PR #1765, opened 2017-10-20
Enable binary transfer over wsHttpProxy.

blois:
This adds the ability to transfer binary data over the HTTP over WebSocket proxy. It also adds a few more fields to allow generating complete browser Response objects with the result.

Binary transfer is enabled by the requester for backwards compat.


blois:
Closing to try to fix the CLA issue.

PR #1767, opened 2017-10-20
Project picker

yebrahim:
This PR adds a new element for selecting project. It uses the cloud resource manager API to return all projects (with no specific order for now), and allows the user to select one from the list.

Currently, this element isn't used anywhere, but I also added a message handler to start it by sending a `pick-project` message to the top frame.

image


umang-sh:
Nice work @yebrahim  :)

PR #1768, opened 2017-10-23
Move build dependencies to be installed by npm instead of requiring global binaries

yebrahim:
This PR follows up on https://github.com/googledatalab/datalab/pull/1731 to move more build dependencies to `package.json` instead of requiring them to have been installed globally.

### DIFF
--- a/sources/build.sh
+++ b/sources/build.sh
@@ -31,21 +31,6 @@ function install_node() {
   export "PATH=${PATH}:/tools/node/bin"
 }
 
-function install_bower() {
-  echo "Installing Bower"
-  npm install -g bower
-}
-
-function install_polyer_cli() {
-  echo "Installing Polymer CLI"
-  npm install -g polymer-cli
-}
-
-function install_typescript() {
-  echo "Installing Typescript"
-  npm install -g typescript
-}
-
 function install_git() {
   echo "Updating apt repository"
   apt-get update -y -qq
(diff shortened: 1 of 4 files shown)


ojarjur:
The test environment doesn't have local versions of typescript, bower, and polymer-cli, so we need those in order for this to build.

I also disagree that requiring locally-installed versions would reduce the friction in building the repo. I think it increases the friction rather than lowering it.

I suspect the best way to lower any friction and to make the build easily reproducible on new machines, would be to run the build inside of a docker container. For instance, we could put together a script that runs the cloudbuild.yaml config locally. If we did that, then our local builds and our cloud builds would match exactly.

nikhilk:
It's customary to use locally installed dev dependencies.

On Oct 23, 2017 9:35 AM, "Omar Jarjur" <notifications@github.com> wrote:

The test environment doesn't have local versions of typescript, bower, and
polymer-cli, so we need those in order for this to build.

I also disagree that requiring locally-installed versions would reduce the
friction in building the repo. I think it increases the friction rather
than lowering it.

I suspect the best way to lower any friction and to make the build easily
reproducible on new machines, would be to run the build inside of a docker
container. For instance, we could put together a script that runs the
cloudbuild.yaml config locally. If we did that, then our local builds and
our cloud builds would match exactly.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/googledatalab/datalab/pull/1768#issuecomment-338718797>,
or mute the thread

sources/web/datalab/polymer/components/datalab-sessions/datalab-sessions.ts
```diff
@@ -217,7 +217,7 @@ class SessionsElement extends Polymer.Element implements DatalabPageElement {
     // notebook, then switches back to the sessions page. Killing a session also
     // triggers a refresh.
     if (!this._sessionListRefreshIntervalHandle) {
-      this._sessionListRefreshIntervalHandle = setInterval(() => {
+      this._sessionListRefreshIntervalHandle = window.setInterval(() => {
```
jimmc: Is this related to using locally installed build tools? If not, I would prefer to see it in a separate PR.

PR #1771, opened 2017-10-23
Long list

yebrahim:
This PR makes the item list render only a limited number of its items (for now 50), with a "Show more" experience when the user scrolls to the end. This avoids a long wait for the first render, and lets the user choose to wait on loading the entire list if desired. Ideally, we should also optimize the list itself to use virtual scrolling, this change is just a stop gap to improve the experience until we do.

image


jimmc:
As we discussed offline, let's hold off on this change and look into using iron-list instead.

yebrahim:
Agreed. Closing.

Issue #1443, opened 2017-06-20
Sidebar should preserve preformatting in Python help


IamJeffG:
Function and module docstrings in Python usually carefully use whitespace to aid the reader. In the screenshot below, see the output of `help(...)`.

Using the `?` syntax to read a function's help--as suggested in the datalab sidebar itself--renders the docstrings in a way that whitespace is discarded (in the screenshot, see the sidebar).

![screenshot from 2017-06-19 17-17-45](https://user-images.githubusercontent.com/2002703/27311426-a09fa90e-5515-11e7-9ec4-9cf2e26a7277.png)

API documentation in the sidebar should be as faithful as possible to the docstring being printed, which means preserving whitespace and maybe even using a monospace font.  In case that is not possible due to width limitations, is there another way to highlight changes in indentation that usually denote the next parameter?

nikhilk:
Agree - we should use a monospace font for help content (esp. for docstring content).

PR #1772, opened 2017-10-23
Use mono-space font in help pane

yebrahim:
Fixes https://github.com/googledatalab/datalab/issues/1443.

image


sources/web/datalab/static/datalab.css
```diff
@@ -575,6 +575,9 @@ div.selected .CodeMirror-linenumber, div#texteditor-container .CodeMirror-linenu
 div.modal-dialog pre, div.modal-dialog code, #help code, #help pre {
   font-size: 13px !important;
 }
+#help {
+  font-family: 'Source Code Pro';
```
nikhilk: Can you add monospace after the specific font, in case that fails to download or not available for whatever reason?

yebrahim: Done.

PR #1777, opened 2017-10-26
Disable user settings by default

yebrahim:
Until we have a good user settings story, this disables the option, unless it's explicitly enabled in the app settings. This change explicitly enables it to keep the current behavior.

This required the optional oauth2 client id to move from user settings to app settings in order not to break the local development story.

### DIFF
--- a/sources/web/datalab/polymer/test/datalab-toolbar-test.ts
+++ b/sources/web/datalab/polymer/test/datalab-toolbar-test.ts
@@ -21,13 +21,47 @@
 describe('<table-preview>', () => {
   let testFixture: ToolbarElement;
 
-  beforeEach(() => {
+  before(() => {
+    SettingsManager.getAppSettingsAsync = () => {
+      const mockSettings: common.AppSettings = {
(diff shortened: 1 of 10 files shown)


sources/web/datalab/polymer/test/datalab-toolbar-test.ts
```diff
@@ -21,13 +21,47 @@
 describe('<table-preview>', () => {
   let testFixture: ToolbarElement;
 
-  beforeEach(() => {
+  before(() => {
+    SettingsManager.getAppSettingsAsync = () => {
+      const mockSettings: common.AppSettings = {
+        allowHttpOverWebsocket: true, allowOriginOverrides: [], configUrl:
```
jimmc: This is pretty ugly. I suppose we can fix later if you want.

yebrahim: Missed this.
I think it's fine for settings, especially if most of them are just there for the build.
The fix is to make all settings definitions in common.d.ts optional, so we can just specify the ones we want here.

PR #1759, opened 2017-10-18
Addressed the flake changes

umang-sh:
Opening another pull request ,since this request :
https://github.com/googledatalab/datalab/pull/1755
got closed unexpectedly

Addressed the flake changes suggested by @chmeyers  
Also ran flake8 test , everything seems fine.

PR #1780, opened 2017-10-27
Extend the reconnection message with more details.

### LINKED ISSUE #1519: Include a message on connection telling how to reconnect if the connection is cancelled.
When this info is output on instance creation:

    The connection to Datalab is now open and will remain until this command is killed.
    You can connect to Datalab at http://localhost:8081/

... also include a message saying how to reconnect (i.e. to run `datalab connect ...`)

ojarjur:
This builds on the change from #1759 and extends it so that the printed command
can be directly copied and pasted in order to reconnect.

This introduces the potential to give the user a command to run that won't actually
work (e.g. if their connection failed due to the specified port being unavailable).

To prevent that risk, I've changed the reconnection message so that we only print
it if the initial connection succeeded.

This closes #1519

### DIFF
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -245,8 +245,8 @@ def connect(args, gcloud_compute, email, in_cloud_shell):
         Args:
           cancelled_event: A threading.Event instance that indicates we should
             give up on the instance becoming reachable.
-          healthy_event: A threading.Event instance that can be used to indicate
-            that the instance became reachable.
+          healthy_event: A threading.Event instance that can be used to
+            indicate that the instance became reachable.
         """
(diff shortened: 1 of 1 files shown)


tools/cli/commands/connect.py
```diff
@@ -67,6 +67,15 @@
     'select *Change port > Port {}*, and start using Datalab.')
 
 
+connection_closed_message = ("""
```
jimmc: Do you need these parentheses?

jimmc: Might want to call this something like "connection_closed_format", since it's not the actual message.

ojarjur: Done.

ojarjur: Done.

PR #1781, opened 2017-10-27
Preserve the config_local oauth client id on release.

chmeyers:


tools/release/config_local_template.js
```diff
@@ -1,6 +1,7 @@
 const vid = document.body.getAttribute("data-version-id");
 const GTM_ACCOUNT = {{GTM_ACCOUNT_PLACEHOLDER}};
 const LATEST_SEMVER = {{DATALAB_VERSION_PLACEHOLDER}};
+const OAUTH2_CLIENT_ID = {{DATALAB_LOCAL_CLIENT_ID}};
```
jimmc: Please also set window.datalab.oauth2ClientId = OAUTH2_CLIENT_ID, similar to GTM_ACCOUNT on line 18.

chmeyers: done

PR #1784, opened 2017-10-31
Fix race condition in gapi code.

jimmc:
This rolls back a piece of #1777. The key issue is that we need to use a 'then' on the return from api.auth2.init before we grab the currentUser to avoid sometimes getting a null result.
The user-visible symptom of this problem is an error dialog saying 'Cannot read property getEmail of undefined.'

I also added code to catch and display an auth error from this updated code. Here is what that looks like for a bad client ID:
datalab-auth-error

### DIFF
--- a/sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
+++ b/sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
@@ -440,21 +440,22 @@ class GapiManager {
   * Initialize the client with API key and People API, and initialize OAuth with an
   * OAuth 2.0 client ID and scopes (space delimited string) to request access.
   */
-  private static _initClient() {
+  private static _initClient(): Promise<void> {
     // TODO: Add state parameter to redirect the user back to the current URL
     // after the OAuth flow finishes.
-    try {
-      const auth2 = gapi.auth2.init({
-        client_id: GapiManager._clientId,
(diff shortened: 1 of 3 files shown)


sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -440,21 +440,22 @@ class GapiManager {
...
+      fetch_basic_profile: true,
+      redirect_uri: Utils.getHostRoot(),
+      scope: initialScopeString,
+      ux_mode: 'redirect',
+    })
+    // The return value of gapi.auth2.init is not a normal promise
+    .then(() => {
+      this._currentUser = gapi.auth2.getAuthInstance().currentUser.get();
+    }, (errorReason: any) => {
+      throw new Error('Error in gapi auth: ' + errorReason.details);
```
yebrahim: The reason behind this change in #1777 is that the error cases weren't being caught correctly. Can you try passing a bad/empty client id here and see if this error is thrown?

jimmc: I tested by manually entering a bad client id. A try/catch block is not effective here to catch that error. This code, which uses the error callback mechanism of the init call, catches that error and throws Error, but it was not being caught by the calling code. I added a catch in one place to display that code.

PR #1785, opened 2017-10-31
Make new notebooks use a template

chmeyers:
Also fixing all the typescript lint errors that appeared due to https://github.com/Microsoft/TypeScript/issues/19614

### DIFF
--- a/sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
+++ b/sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
@@ -19,8 +19,8 @@
  */
 class AuthPanel extends Polymer.Element {
 
-  private _signedIn: boolean;
-  private _userInfo: string;
+  protected _signedIn: boolean;
+  protected _userInfo: string;
   private _promptOnSignIn = false;
 
   static get is() { return 'auth-panel'; }
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -80,19 +80,20 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
    */
   public nLeadingBreadcrumbsToTrim: number;
 
-  private _addToolbarCollapseThreshold = 900;
   protected _busy: boolean; // Indicates an async file operation is taking place
   protected _canOpenInNotebook = false;
   protected _canPreview = false;
+  protected _fetching: boolean; // Indicates the file list is being fetched and updated
+  protected _fileManagerDisplayName: string;
+  protected _fileManagerDisplayIcon: string;
+
+  private _addToolbarCollapseThreshold = 900;
   private _dividerPosition: number;
   private _previewPaneCollapseThreshold = 600;
-  protected _fetching: boolean; // Indicates the file list is being fetched and updated
   private _fileList: DatalabFile[];
   private _fileListRefreshInterval = 60 * 1000;
   private _fileListRefreshIntervalHandle = 0;
   private _fileManager: FileManager;
-  protected _fileManagerDisplayName: string;
-  protected _fileManagerDisplayIcon: string;
   private _hasMultipleFileSources: boolean;
   private _ignoreFileIdChange = false;
   private _inlineDetailsOpenInNotebook: () => void | null;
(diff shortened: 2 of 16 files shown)


sources/web/datalab/polymer/components/auth-panel/auth-panel.ts
```diff
@@ -19,8 +19,8 @@
  */
 class AuthPanel extends Polymer.Element {
 
-  private _signedIn: boolean;
-  private _userInfo: string;
+  protected _signedIn: boolean;
```
yebrahim: The right fix here would be to remove these modifiers altogether, effectively making these public properties, which they actually are, since they're being accessed from the HTML. We were abusing the fact that typescript used to not flagging write-only accessed variables to treat them as private, but they're really not.

We could still name them with an underscore though, to signal that they're really just private to this web component and shouldn't be accessed directly from outside the component.

yebrahim: Additionally, can we please separate this cleanup from the actual change? It's not a small cleanup.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -583,15 +584,15 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
     }
   }
 
-  _createNewNotebook() {
+  async _createNewNotebook() {
```
yebrahim: These don't need to be `async` if they won't `await`.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -735,41 +736,41 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
...
+      try {
+        let newName = closeResult.userInput;
+        // Make sure the name ends with .ipynb for notebooks for convenience
+        if (itemType === DatalabFileType.NOTEBOOK && !newName.endsWith('.ipynb')) {
+          newName += '.ipynb';
         }
-      });
+
+        this._busy = true;
+        const newFile = await this._fileManager.create(itemType, this.currentFile.id, newName);
```
yebrahim: We can instead create the new file with the new notebook template and go to it directly, see: https://github.com/googledatalab/datalab/blob/47fa7fd3af7c3b3aa856961efb544af9bab1f866/sources/web/datalab/polymer/components/table-inline-details/table-inline-details.ts#L167

This way you'd also fix b/67907728.

sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -126,6 +127,24 @@ class BigQueryTableOverviewTemplate extends NotebookTemplate {
...
+ * This template contains intro cells so the notebook isn't empty on creation.
+ */
+class NewNotebookTemplate extends NotebookTemplate {
+  constructor(dict: { [key: string]: any }) {
+    const parameters = Object.keys(dict).map((k) => ({name: k, value: dict[k]}));
+
+    // Specify the default location of the template.
+    const defaultTemplateLocation = 'static/templates/NewNotebookTemplate.ipynb';
+
+    const windowDatalab = window.datalab || {};
```
yebrahim: Can we move this to the global scope in `template-manager.ts` and clean it up from all template constructors? It's really meant to be done once, and is there only for development.

sources/web/datalab/polymer/templates/NewNotebookTemplate.ipynb
```diff
@@ -0,0 +1,43 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#Place your code here.\n",
+    "\n"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
```
yebrahim: Do we really want to include a kernel spec here? I think python 2 is the default anyway, but we should consider either taking it out or making it a parameter placeholder.

sources/web/datalab/polymer/templates/NewNotebookTemplate.ipynb
```diff
@@ -0,0 +1,43 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
```
yebrahim: I thought we were going to have two markdown cells, one with the header title and one with a one line description?

yebrahim: Also, are we going to make the title a parameter?

chmeyers:
Closing in favor of https://github.com/googledatalab/datalab/pull/1788

PR #1793, opened 2017-11-03
Sanitize param values sent to the template creator.

chmeyers:
I'm taking a deliberatly aggressive approach here since we don't
currently have a use case where any special characters should be
needed in these strings.

### DIFF
--- a/sources/web/datalab/polymer/modules/template-manager/template-manager.ts
+++ b/sources/web/datalab/polymer/modules/template-manager/template-manager.ts
@@ -44,7 +44,20 @@ class NotebookTemplate {
 
   constructor(fileId: DatalabFileId, parameters: TemplateParameter[]) {
     this.fileId = fileId;
-    this.parameters = parameters;
+    this.parameters = NotebookTemplate._sanitizeParams(parameters);
+  }
+
+  /**
+   * Strictly sanitizes the parameters.
+   */
+  private static _sanitizeParams(params: TemplateParameter[]) {
+    params.forEach((p: TemplateParameter, i) => {
+      params[i].name = p.name.replace(/[^A-Za-z0-9._\-]/g, '');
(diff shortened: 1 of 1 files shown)


sources/web/datalab/polymer/modules/template-manager/template-manager.ts
```diff
@@ -44,7 +44,20 @@ class NotebookTemplate {
...
     this.fileId = fileId;
-    this.parameters = parameters;
+    this.parameters = NotebookTemplate._sanitizeParams(parameters);
+  }
+
+  /**
+   * Strictly sanitizes the parameters.
+   */
+  private static _sanitizeParams(params: TemplateParameter[]) {
+    params.forEach((p: TemplateParameter, i) => {
```
nikhilk: Isn't the notebook name part of the parameters? And that can be anything the user wants?

chmeyers: The filename isn't part of these params.  We currently allow anything for the name and sanitize those separately when displaying them, since we don't control all of the paths for creating a file (i.e. a user could create one named with a script tag themselves outside of Datalab)

nikhilk: Isn't the file name going to be used to populate a markdown cell within the template to serve as the starting title?

chmeyers: We are not yet doing that for any of our templates.  To do so, we'd have to pass the fileName as a parameter to the template manager, in which case it would go through this sanitization.

nikhilk: That _was_ the plan for the templates -- first cell contains the title based on the name.

PR #1795, opened 2017-11-03
Add redirect-to-notebook message.

jimmc:


sources/web/datalab/polymer/notebook.ts
```diff
@@ -29,6 +29,7 @@ const CommandId = {
   ERROR: 'error',
   LOAD_NOTEBOOK: 'load-notebook',
   PICK_PROJECT: 'pick-project',
+  REDIRECT_TO_NOTEBOOK: 'redirect-to-notebook',
```
yebrahim: Seems to me like this is a generic redirect functionality, should we call this just `REDIRECT` or `CHANGE_URL`?

sources/web/datalab/polymer/notebook.ts
```diff
@@ -83,6 +84,25 @@ async function processMessageEvent(e: MessageEvent) {
       sendMessageToNotebookEditor(outgoingMessage);
       break;
 
+    case CommandId.REDIRECT_TO_NOTEBOOK:
+      try {
+        const id = message.arguments;
+        const url = '/notebook/' + id;
+        window.location.href = url;
+        outgoingMessage = {
```
yebrahim: Does this work reliably? Doesn't the whole app stop executing after the browser switches away? Perhaps we should send the async reply message then set `location.href`?

sources/web/datalab/polymer/notebook.ts
jimmc: Yes, we could make it more generic. I wanted to restrict it just to notebooks for now, to make sure the notebook couldn't redirect to something else. We can consider the potential security of issues of that if that use case arises.

sources/web/datalab/polymer/notebook.ts
jimmc: It seems to take a while for the setting of href to take effect. I initially had a bug in some of the code after the href line, and I was getting an error dialog that blinked up before it switched to the other location. I assumed we needed not to have anything important running after setting href, so as long as it doesn't do anything bad, I'm not worried about whether it actually executes anything past that point.

yebrahim: I'm more concerned about that code potentially not getting executed at all. I think running any code after setting `location.href` is at least unpredictable.

PR #1802, opened 2017-11-07
Ask for multiple scopes up front, don't ask for others later.

jimmc:


sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -21,12 +21,14 @@ class MissingClientIdError extends Error {
 
 enum GapiScopes {
   BIGQUERY,
```
nikhilk: Could we do without needing to list BigQuery and GCS separately?

sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -21,12 +21,14 @@ class MissingClientIdError extends Error {
...
+  CLOUD,
   DRIVE,
   GCS,
 }
 
-// Authorize GCP scope by default
-const initialScopeString = 'profile email https://www.googleapis.com/auth/cloud-platform';
+// Ask for all necesary scopes up front so we don't need to ask again later.
+// Cloud-platform scope covers BigQuery and GCS but not Drive.
+const initialScopes = [ GapiScopes.CLOUD, GapiScopes.DRIVE ];
```
chmeyers: Do we need to include "profile" and "email" here?

jimmc: Yeah, I just noticed that, will put those back in.

sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
jimmc: Removed BIGQUERY and GCS scopes.

sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
jimmc: OK, done.

PR #1804, opened 2017-11-09
CreateNewItem tests

yebrahim:
This also adds a new handy test utility that waits until a given condition evaluates to true.

### DIFF
--- a/sources/web/datalab/polymer/test/file-browser-test.ts
+++ b/sources/web/datalab/polymer/test/file-browser-test.ts
@@ -197,7 +197,7 @@ window.addEventListener('WebComponentsReady', () => {
       multiSelectionEnabledButtonIds.forEach((id) => assertEnabledState(id, true));
     });
 
-    it('creates new notebook correctly', async () => {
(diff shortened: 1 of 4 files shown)


sources/web/datalab/polymer/test/file-browser-test.ts
```diff
@@ -194,5 +196,59 @@ window.addEventListener('WebComponentsReady', () => {
       singleSelectionEnabledButtonIds.forEach((id) => assertEnabledState(id, false));
       multiSelectionEnabledButtonIds.forEach((id) => assertEnabledState(id, true));
     });
+
+    it('creates new notebook correctly', async () => {
```
jimmc: This is perhaps too picky, but these three tests do not quite test what their descriptions say they do. You are testing to see if the file-browser correctly calls the underlying mechanism that performs these actions, but you are not actually testing to see that those actions were correctly done. But unless you can come up with a concise wording, then it's acceptable as is.

yebrahim: No I agree. Let me try to make it more correct.

PR #1805, opened 2017-11-09
Add filter to item-list

yebrahim:
This PR adds filtering capability to item list. The filter only works on the first column of the item list for now, which covers the common cases. We can expand on this later as needed.

In order to do this, I switched the way we get the selection index to use the clicked item's true order in the original list, as opposed to its order in the rendered (filtered) list.

This design isn't final, it's just a rough way to get filter functionality into the UI. It's a toggle that shows up next to the first column in the header, which on click shows an input box underneath the header. Items will change as the user types in that box, without needing to hit Enter. Filtering is case-insensitive for now to simplify the implementation and the UI.

image
[_Tooltip when hovering over filter icon_]

image
[_Clicking filter icon opens the filter box and focuses it_]

### DIFF
--- a/sources/web/datalab/polymer/components/item-list/item-list.ts
+++ b/sources/web/datalab/polymer/components/item-list/item-list.ts
@@ -186,6 +186,7 @@ class ItemListElement extends Polymer.Element {
    */
   public inlineDetailsMode: InlineDetailsDisplayMode;
 
+  _filterString: string;
   _showFilterBox: boolean;
 
   private _lastSelectedIndex = -1;
@@ -227,6 +228,7 @@ class ItemListElement extends Polymer.Element {
         value: false,
       },
       rows: {
+        observer: '_rowsChanged',
         type: Array,
         value: () => [],
       },
@@ -252,8 +254,28 @@ class ItemListElement extends Polymer.Element {
     });
   }
 
+  _rowsChanged() {
+    this._filterString = '';
+  }
+
   _toggleFilter() {
     this._showFilterBox = !this._showFilterBox;
+    if (this._showFilterBox) {
+      this.$.filterBox.focus();
+    } else {
+      this._filterString = '';
+    }
+  }
+
+  _computeFilter(filterString: string) {
+    if (!filterString) {
+      // set filter to null to disable filtering
+      return null;
+    } else {
+      // return a filter function for the current search string
+      filterString = filterString.toLowerCase();
+      return (item: ItemListRow) => item.columns[0].toLowerCase().indexOf(filterString) > -1;
+    }
   }
 
   /**
--- a/sources/web/datalab/polymer/components/item-list/item-list.ts
+++ b/sources/web/datalab/polymer/components/item-list/item-list.ts
@@ -228,7 +228,6 @@ class ItemListElement extends Polymer.Element {
         value: false,
       },
       rows: {
-        observer: '_rowsChanged',
         type: Array,
         value: () => [],
       },
@@ -254,8 +253,7 @@ class ItemListElement extends Polymer.Element {
     });
   }
 
-  _rowsChanged() {
-    // Reset the filter if the items change
+  resetFilter() {
     this._filterString = '';
   }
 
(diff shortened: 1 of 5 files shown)


sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -246,6 +254,34 @@ class ItemListElement extends Polymer.Element {
     });
   }
 
+  _rowsChanged() {
+    // Reset the filter if the items change
```
jimmc: We are auto-refreshing the file list, so if it happens to change while we are using filtering, this code will clear our filtering expression. That seems like it would be annoying. Maybe we can have a flag so we can tell whether this change to the file list is happening due to user action as opposed to auto-refresh, and only clear the filter when it is based on user action.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -436,5 +436,80 @@ describe('<item-list>', () => {
       assert(fifthDetailsContainer.getAttribute('hidden') != null,
           'fifth details container should be hidden');
     });
+
+    it('should hide the filter box by default', () => {
```
jimmc: Are we using 'should' or not in our test descriptions? I guess I would be happy with not using it so our descriptions are a bit shorter.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -436,5 +436,80 @@ describe('<item-list>', () => {
...
+      assert(rows.length === 0,
+          'should not show any rows, since no row has "second" in its first column');
+    });
+
+    it('ignores case when filtering', () => {
+      testFixture.$.filterToggle.click();
+      testFixture._filterString = 'FIRST';
+      Polymer.dom.flush();
+      const rows = testFixture.$.listContainer.querySelectorAll('.row');
+      assert(rows.length === 5, 'should show all rows, since filtering is case insensitive');
```
jimmc: Would be better to use a filter string that returns a subset of the list, to ensure that it is not simply ignoring all upper case in the filter string.

sources/web/datalab/polymer/components/item-list/item-list.ts
yebrahim: We only set `rows` if it will actually change to avoid changing current selection. This shouldn't have an effect.

sources/web/datalab/polymer/test/item-list-test.ts
yebrahim: Done.

sources/web/datalab/polymer/test/item-list-test.ts
yebrahim: Done.

sources/web/datalab/polymer/components/item-list/item-list.ts
jimmc: Yes, I am referring to the situation in which rows actually changes. The scenario is, I am looking at a shared folder with 300 files in it, and I have entered a filter string to get it down to 10 files, then someone else adds a new file to that folder, and my client auto-refreshes. In that case, although I have not done anything, I will suddenly lose my filter string and be back to looking at 301 files.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -482,10 +482,12 @@ describe('<item-list>', () => {
 
     it('ignores case when filtering', () => {
       testFixture.$.filterToggle.click();
-      testFixture._filterString = 'FIRST';
+      testFixture._filterString = 'COLUMN 4';
       Polymer.dom.flush();
       const rows = testFixture.$.listContainer.querySelectorAll('.row');
-      assert(rows.length === 5, 'should show all rows, since filtering is case insensitive');
+      assert(rows.length === 1, 'should show all rows, since filtering is case insensitive');
```
jimmc: comment needs to be updated

sources/web/datalab/polymer/components/item-list/item-list.ts
yebrahim: Good point. I'll move resetting the filter to be done from outside the component, and have the file browser component reset it only when changing directory.

PR #1807, opened 2017-11-10
Expose columns in file browser from file manager

yebrahim:
This PR allows each file manager to define its own set of columns to be displayed by a file-browser element. This enables the file browser to then display different columns per source, as a precursor to implementing sort over columns.

A significant portion of this PR is cleanup to the different DatalabFile extended class constructors to make things consistent and easy to extend.

Examples:
image
[_Drive view now shows last modified and owner columns_]

image
[_BigQuery view now changes the column name based on the browser level_]

### DIFF
--- a/sources/web/datalab/polymer/components/item-list/item-list.html
+++ b/sources/web/datalab/polymer/components/item-list/item-list.html
@@ -93,9 +93,14 @@ the License.
       .column {
         flex: 2;
       }
+      #header > .column {
+        color: var(--primary-fg-color);
+        font-size: 13px;
+      }
       .column + .column {
         flex: 1;
         max-width: 150px;
+        color: var(--neutral-fg-color);
       }
       #filterToggle {
         --paper-icon-button: {
@@ -153,7 +158,7 @@ the License.
         </div>
         <template is="dom-repeat" items="{{columns}}" as="column">
           <div class="column">
-            <strong>{{column}}</strong>
+            <span>{{column}}</span>
           </div>
         </template>
       </div>
--- a/sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
+++ b/sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
@@ -18,6 +18,12 @@
  */
 
 class DriveFile extends DatalabFile {
+  lastModified?: string;
+  owner?: string;
+
+  getColumnValues() {
+    return [this.name, this.lastModified || '', this.owner || ''];
+  }
 }
 
 /**
@@ -60,16 +66,20 @@ class DriveFileManager extends BaseFileManager {
       'mimeType',
       'modifiedTime',
       'name',
+      'owners',
       'parents',
     ];
-    const orderModifiers = [
-      'folder',
-      'modifiedTime desc',
-      'name',
-    ];
 
-    const upstreamFiles = await GapiManager.drive.listFiles(fileFields, queryPredicates, orderModifiers);
+    const upstreamFiles = await GapiManager.drive.listFiles(fileFields, queryPredicates);
     return upstreamFiles.map((file) => this._fromUpstreamFile(file));
+  }
+
+  public getColumns() {
+    return [
+      Utils.constants.columns.name,
+      Utils.constants.columns.lastModified,
+      Utils.constants.columns.owner,
+    ];
   }
 
   public async create(fileType: DatalabFileType, containerId?: DatalabFileId, name?: string)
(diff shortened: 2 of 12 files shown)


sources/web/datalab/polymer/components/item-list/item-list.html
```diff
@@ -93,9 +93,14 @@
...
         flex: 2;
       }
+      #header > .column {
+        color: var(--primary-fg-color);
+        font-size: 13px;
+      }
       .column + .column {
         flex: 1;
         max-width: 150px;
+        color: var(--neutral-fg-color);
```
jimmc: Making the first column be bolder than the others makes me think the data is sorted by the first column. I think I would be happier with all the data being the primary color.

sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
```diff
@@ -116,27 +126,33 @@ class DriveFileManager extends BaseFileManager {
...
-    } as DatalabFile);
-    if (datalabFile.name.endsWith('.ipynb')) {
-      datalabFile.type = DatalabFileType.NOTEBOOK;
+      file.mimeType === DriveFileManager._directoryMimeType ?
+                              file.iconLink : 'editor:insert-drive-file',
+    );
+    if (driveFile.type === DatalabFileType.FILE && driveFile.name.endsWith('.ipynb')) {
+      driveFile.type = DatalabFileType.NOTEBOOK;
+    }
+    driveFile.lastModified = new Date(file.modifiedTime).toDateString();
```
jimmc: Converting this date to a string here will make it difficult to sort on that column. Maybe you could do the string conversion in DatalabFile.getColumnValues instead. Or return a value here that has a toString() that generates the right display value, but that also can be sorted.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
```diff
@@ -239,6 +247,11 @@ class BaseFileManager implements FileManager {
   list(_containerId: DatalabFileId): Promise<DatalabFile[]> {
     throw new UnsupportedMethod('list', this);
   }
+
+  getColumns(_currentFileId?: DatalabFileId) {
```
jimmc: Maybe call this getColumnNames?

sources/web/datalab/polymer/modules/utils/utils.ts
```diff
@@ -32,6 +32,7 @@ class Utils {
     directory: 'folder',
     editorUrlComponent:   '/editor/',
     file: 'file',
+    me: 'Me',
```
jimmc: The Drive UI uses lower case 'me'. We could choose to be consistent with that.

sources/web/datalab/polymer/test/file-browser-test.ts
```diff
@@ -115,15 +118,18 @@ window.addEventListener('WebComponentsReady', () => {
       mockFiles.forEach((file: DatalabFile, i: number) => {
         assert(files.rows[i].columns[0] === file.name,
             'mock file ' + i + 'name not shown in first column');
+        assert(files.rows[i].columns[1] === file.type.toString(),
+            'mock file ' + i + 'type not shown in first column');
```
jimmc: 'in second column'

sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
jimmc: Using toDateString() here shows only the date and not the time (yes, I know the Drive UI doesn't show mod time, but we could). Maybe you could do it like <code>ls -l</code> does, showing the month/day/time when less than six months old, and showing the month/day/year when more than six months old. Or maybe use Date.toLocaleString().

sources/web/datalab/polymer/components/item-list/item-list.html
yebrahim: I'm not sure I see why the darker color implies sort order. It just makes this column more prominent, which is the original intent.
This is also inspired by the Drive UI, which I think makes sense.

sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
yebrahim: Good point.
I think we should do this more properly by defining the column type along with its name in the item-list component, so that ordering can use a custom sort function. I'd like to do that as a separate change.

yebrahim: Switched to using `toLocaleString`.

sources/web/datalab/polymer/modules/file-manager/file-manager.ts
yebrahim: Done.

sources/web/datalab/polymer/modules/utils/utils.ts
yebrahim: Done.

sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
jimmc: I'd ask you to add a TODO if it didn't require another round-trip for approval. Or maybe you are planning on working on that right away anyway.

yebrahim: Yup, it's next on my list. :)

PR #1809, opened 2017-11-10
Only allow valid names for new files.

jimmc:
If you try to create a file with name ending in .ipynb you will see this:

file-no-ipynb

If you try to create a file that does not include .txt in the name, you will see this:

file-txt

### DIFF
--- a/sources/web/datalab/polymer/components/file-browser/file-browser.ts
+++ b/sources/web/datalab/polymer/components/file-browser/file-browser.ts
@@ -766,6 +766,16 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
         // Make sure the name ends with .ipynb for notebooks for convenience
         if (itemType === DatalabFileType.NOTEBOOK && !newName.endsWith('.ipynb')) {
           newName += '.ipynb';
+        } else if (itemType === DatalabFileType.FILE) {
+          if (newName.endsWith('.ipynb')) {
+            Utils.showErrorDialog('Invalid filename', 'Only notebooks can end with .ipynb');
+            return;
+          }
+          const fileNameError = this._fileManager.newFileNameError(newName);
+          if (fileNameError) {
+            Utils.showErrorDialog('Invalid filename', fileNameError);
+            return;
+          }
         }
 
         this._busy = true;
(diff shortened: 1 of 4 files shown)


sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
```diff
@@ -52,6 +52,15 @@ class DriveFileManager extends BaseFileManager {
     return this._fromUpstreamFile(upstreamFile);
   }
 
+  public newFileNameError(fileName: string): string | null {
+    // Must match _getWhitelistFilePredicates()
+    if (fileName.indexOf('.ipynb') > -1 || fileName.indexOf('.txt') > -1) {
```
chmeyers: Shouldn't this be "endsWith" of something instead of indexOf?

Also, I feel like we should have a function or something in one place that controls the list of acceptable file extensions.

jimmc: I was trying to make it match the search criteria used in _getWhitelistFilePredicates, but I guess we can be more strict here and require the .txt to be at the end of the name. Done.
Re having a function in one place, at present the filename filtering is on a per-file-manager basis; for example, the jupyter file manager shows all files. We can add a master type list later once we figure out what it should control.

nikhilk:
What about .py, .yaml, .csv, .json, .md etc etc.

chmeyers:
Discussed .py, .yaml, .csv, .json, .md, etc with Jim in person.  Short answer: we don't currently show any of those in the list for these file managers.  When we update the list of file types we show in the list, we should also allow you to create those types.

sources/web/datalab/polymer/components/file-browser/file-browser.ts
```diff
@@ -766,6 +766,16 @@ class FileBrowserElement extends Polymer.Element implements DatalabPageElement {
         // Make sure the name ends with .ipynb for notebooks for convenience
         if (itemType === DatalabFileType.NOTEBOOK && !newName.endsWith('.ipynb')) {
           newName += '.ipynb';
+        } else if (itemType === DatalabFileType.FILE) {
+          if (newName.endsWith('.ipynb')) {
+            Utils.showErrorDialog('Invalid filename', 'Only notebooks can end with .ipynb');
```
yebrahim: Why not switch to creating a notebook instead of blocking the user operation?

sources/web/datalab/polymer/modules/drive-file-manager/drive-file-manager.ts
```diff
@@ -157,6 +166,7 @@ class DriveFileManager extends BaseFileManager {
 
   protected _getWhitelistFilePredicates() {
     return [
+      // Must match newFileNameIsValid()
```
yebrahim: What's `newFileNameIsValid`?

PR #1812, opened 2017-11-12
Add sort to item-list

yebrahim:
This PR adds sorting capability to the item-list element. It does this by adding a new `Column` class that the users of `item-list` use to specify each column's name and type. Currently only 'string' and 'date' types are supported, but this list can easily extend to numbers for example.

The UI adds a down/up arrow icon next to the currently sorted column, and turns columns into clickable buttons. Clicking the currently sorted column again reverses the sort order.

image

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,4 +40,4 @@ script:
 - for i in `ls ui/broken/`; do echo ui/broken/$i; base64 ui/broken/$i; echo \"-----\"; done
 - cd ../sources/web/datalab/polymer/test
 - export DISPLAY=:99.0
-- npm test -- --verbose
+- npm test -- --verbose -l chrome
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,4 +40,4 @@ script:
 - for i in `ls ui/broken/`; do echo ui/broken/$i; base64 ui/broken/$i; echo \"-----\"; done
 - cd ../sources/web/datalab/polymer/test
 - export DISPLAY=:99.0
-- npm test -- --verbose -l chrome
+- npm test -- --verbose
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,4 +40,5 @@ script:
 - for i in `ls ui/broken/`; do echo ui/broken/$i; base64 ui/broken/$i; echo \"-----\"; done
 - cd ../sources/web/datalab/polymer/test
 - export DISPLAY=:99.0
-- npm test -- --verbose
+- export TEST_LOG="${HOME}/${TRAVIS_JOB_NUMBER}_npm_test.log"
+- timeout 2m bash -c "npm test -- --verbose | tee '${TEST_LOG}'" || grep "Test run ended with great success" "${TEST_LOG}"
--- a/sources/web/datalab/polymer/components/item-list/item-list.html
+++ b/sources/web/datalab/polymer/components/item-list/item-list.html
@@ -201,7 +201,7 @@ the License.
           </div>
           <template is="dom-repeat" items="{{row.columns}}" as="column">
             <div class="column">
-              <span>[[_formatColumnValue(column, index)]]</span>
+              <span>[[_formatColumnValue(column, index, columns)]]</span>
             </div>
           </template>
         </paper-item>
--- a/sources/web/datalab/polymer/components/item-list/item-list.html
+++ b/sources/web/datalab/polymer/components/item-list/item-list.html
@@ -16,9 +16,11 @@ the License.
 
 <link rel="import" href="../../bower_components/iron-icons/editor-icons.html">
 <link rel="import" href="../../bower_components/iron-icons/hardware-icons.html">
+<link rel="import" href="../../bower_components/iron-icons/iron-icons.html">
+<link rel="import" href="../../bower_components/paper-button/paper-button.html">
 <link rel="import" href="../../bower_components/paper-checkbox/paper-checkbox.html">
-<link rel="import" href="../../bower_components/paper-item/paper-item.html">
 <link rel="import" href="../../bower_components/paper-icon-button/paper-icon-button.html">
+<link rel="import" href="../../bower_components/paper-item/paper-item.html">
 
 <dom-module id="item-list">
   <template>
@@ -97,6 +99,19 @@ the License.
         color: var(--primary-fg-color);
         font-size: 13px;
       }
+      .column-button {
+        display: flex;
+        justify-content: flex-start;
+        width: 100%;
+        padding: 0 10px;
+        margin: 0;
+        border-radius: 0;
+        text-transform: none;
+      }
+      .column-button > iron-icon {
+        width: 15px;
+        padding-left: 3px;
+      }
       .column + .column {
         flex: 1;
         max-width: 150px;
@@ -158,7 +173,10 @@ the License.
         </div>
         <template is="dom-repeat" items="{{columns}}" as="column">
           <div class="column">
-            <span>{{column.name}}</span>
+            <paper-button class="column-button" on-click="_columnButtonClicked">
+              {{column.name}}
+              <iron-icon class="sort-icon"></iron-icon>
+            </paper-button>
           </div>
         </template>
       </div>
--- a/sources/web/datalab/polymer/components/item-list/item-list.html
+++ b/sources/web/datalab/polymer/components/item-list/item-list.html
@@ -93,7 +93,7 @@ the License.
         line-height: 26px;
       }
       .column {
-        flex: 2;
+        flex: 1.5;
       }
       #header > .column {
         color: var(--primary-fg-color);
@@ -116,6 +116,7 @@ the License.
         flex: 1;
         max-width: 150px;
         color: var(--neutral-fg-color);
+        padding-left: 10px;
       }
       #filterToggle {
         --paper-icon-button: {
@@ -200,7 +201,7 @@ the License.
           </div>
           <template is="dom-repeat" items="{{row.columns}}" as="column">
             <div class="column">
-              <span>{{column}}</span>
+              <span>[[_formatColumnValue(column, index)]]</span>
             </div>
           </template>
         </paper-item>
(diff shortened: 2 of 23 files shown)


sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -251,12 +257,60 @@ class ItemListElement extends Polymer.Element {
...
+    this.$.list.sort = (a: ItemListRow, b: ItemListRow) => {
+      // Bail out of sort if no columns have been set yet.
+      if (!this.columns.length) {
+        return;
+      }
+      let compResult = -1;
+      if ((this.columns[column].type === 'string' &&
+          a.columns[column].toLowerCase() > b.columns[column].toLowerCase()) ||
+          (this.columns[column].type === 'date' &&
+          new Date(a.columns[column]) > new Date(b.columns[column]))) {
```
jimmc: Creating Date objects from a string is "[strongly discouraged](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse)" due to differences in parser implementation. Plus, isn't this now a locale-dependent string?
Now that item-list knows the type of each column, you could make DatalabFile.getColumnValues return the value as a date object and do the conversion to a string when displaying the column value in item-list. That would mean that a.columns[column] at this point would be a date object, so you can just directly compare. It also would centralize the date-to-string conversion code in one place in item-list rather than in every file-manager that returns a date column.

.travis.yml
```diff
@@ -40,4 +40,4 @@ script:
 - for i in `ls ui/broken/`; do echo ui/broken/$i; base64 ui/broken/$i; echo \"-----\"; done
 - cd ../sources/web/datalab/polymer/test
 - export DISPLAY=:99.0
-- npm test -- --verbose
+- npm test -- --verbose -l chrome
```
jimmc: Why change this back? Does the sorting not work on Firefox?

sources/web/datalab/polymer/components/item-list/item-list.ts
yebrahim: According to that link, this is only pre-ES5?
Aren't we doing the conversion from string to Date anyway, because we get the value as a string from the upstream service?

.travis.yml
yebrahim: This was added to the commit by mistake while locally testing. Will remove.

sources/web/datalab/polymer/components/item-list/item-list.ts
jimmc: The date strings coming from Drive are in ISO8601 format, which is safe to parse. As long as Americans use mm/dd/yyyy and Europeans use dd/mm/yyyy, you won't be able to unambiguously parse localized date strings without knowing _a priori_ what format they are in. Date parsing of anything other than a known fixed format is brittle.
Sorting on the actual date here rather than trying to parse the string will make it possible to display more sophisticated date strings, such as the year-or-time format of <code>ls -l</code>.

yebrahim: Currently, the item-list's template prints the value of the column as specified in the `rows` array, which means a value of type `Date` will be converted to string by calling its `toString` method. We have no control over this as far as I can see. The only way out of this is by keeping two versions of the `rows` data, one with `Date` objects and one with their respective `string` versions in whichever format we want, but this complicates things quite a bit.

What if we convert all date strings into Date objects, then back to ISO-format strings before passing them to the item-list? This way the item-list itself only parses standard formats.

jimmc: I think you should be able to do something like this in item-list.html:
```
  <template is="dom-repeat" items="{{row.columns}}" as="column" indexAs="columnIndex">
    <div class="column">
      <span>[[_computeColumnDisplay(column, columnIndex)]]</span>
    </div>
  </template>
```
_computeColumnDisplay can then use this.columns[columnIndex].type to determine how to format the column value.

yebrahim: Ok I modified the item list to support Date, string, and number types, and do proper sorting for these. We can think about the `ls -l` semantics separately. PTAL.

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -12,6 +12,19 @@
  * the License.
  */
 
+type ColumnTypeList = Date|number|string;
```
jimmc: This type name sounds to me like it is a list of the ColumnType types defined just below. How about something like "ColumnJavascriptType"?

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -251,12 +270,80 @@ class ItemListElement extends Polymer.Element {
...
+        column,
+      };
+    }
+
+    this.$.list.sort = (a: ItemListRow, b: ItemListRow) => {
+      // Bail out of sort if no columns have been set yet.
+      if (!this.columns.length) {
+        return;
+      }
+      let compResult = -1;
```
jimmc: Shouldn't you be returning 0 if the two items are equal? Otherwise, it seems possible you will have problems when sorting lists with equal elements.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -500,6 +515,145 @@ describe('<item-list>', () => {
       assert(testFixture.$.listContainer.querySelectorAll('.row').length === 5,
           'all rows should show after closing filter box');
     });
+  });
+
+  describe('sorting', () => {
+    const rows = [
+      new ItemListRow({columns: ['item c*', 'Sat Nov 11 2017 18:58:42 GMT+0200 (EET)']}),
```
jimmc: Looks like these date strings should now be Date values instead, and change the type on line 539 to DATE.

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -500,6 +515,145 @@ describe('<item-list>', () => {
...
+      columns0 = renderedRows[0].querySelectorAll('.column');
+      assert(columns0[0].innerText === testFixture.rows[0].columns[0]);
+      assert(columns0[1].innerText === testFixture.rows[0].columns[1]);
+      // row 1
+      columns1 = renderedRows[1].querySelectorAll('.column');
+      assert(columns1[0].innerText === testFixture.rows[1].columns[0]);
+      assert(columns1[1].innerText === testFixture.rows[1].columns[1]);
+    });
+
+    it('sorts numbers correctly', () => {
```
jimmc: Could add a test to see what happens when there are a few values that are equal.

sources/web/datalab/polymer/components/item-list/item-list.html
```diff
@@ -182,7 +201,7 @@
           </div>
           <template is="dom-repeat" items="{{row.columns}}" as="column">
             <div class="column">
-              <span>{{column}}</span>
+              <span>[[_formatColumnValue(column, index, columns)]]</span>
```
jimmc: You are inside a nested dom-repeat here. I think it would add clarity and robustness to be explicit about which index you are using by adding index-as to the columns dom-repeat.

sources/web/datalab/polymer/components/item-list/item-list.ts
yebrahim: Is there any advantage to return 0 as opposed to -1 if two items are equal?

sources/web/datalab/polymer/test/item-list-test.ts
yebrahim: Done.

sources/web/datalab/polymer/test/item-list-test.ts
yebrahim: Done.

sources/web/datalab/polymer/components/item-list/item-list.ts
jimmc: That's probably OK.

sources/web/datalab/polymer/components/item-list/item-list.ts
jimmc: The contract for a javascript comparator function is to return 0 if the two values are equal. Broken comparator functions can cause incorrect sorting. Even if it works today, the sort function might change in a future version in a way that reveals the broken comparator and causes the sort not to order correctly.

yebrahim: Done.

sources/web/datalab/polymer/components/item-list/item-list.ts
yebrahim: Done.

sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -251,12 +270,83 @@ class ItemListElement extends Polymer.Element {
...
+      this._currentSort = {
+        asc: true,
+        column,
+      };
+    }
+
+    this.$.list.sort = (a: ItemListRow, b: ItemListRow) => {
+      // Bail out of sort if no columns have been set yet.
+      if (!this.columns.length) {
+        return;
```
jimmc: Doesn't typescript complain about have a no-value return in a function that returns a value?

yebrahim: Unless you define a return type, the compiler will compute the return type for this function to be `number|void`, which is reasonable to me since at this point there's missing information, and the function shouldn't return a number.

jimmc: What behavior do we want or expect from the caller in this case? It seems unlikely the caller is expecting to get back a void value, so it could end of doing something strange. Perhaps this function should return 0 in this case (we define all rows as being equal, meaning we don't care how it gets sorted), or throw an exception.

yebrahim: I missed this reply. Sure this sounds reasonable. I can include it in the next change to item-list, which can be your other suggestion of optionally using relative time for Date columns.

PR #1826, opened 2017-11-17
Timeout NPM tests after 2 minutes

ojarjur:


yebrahim:
Are we hitting cases where Travis is timing out the job because of tests?

ojarjur:
@yebrahim Yes; more specifically, the tests pass but the test command never exists.

PR #1830, opened 2017-11-20
Notebook preview fixes and tests

yebrahim:
This PR fixes a few things with the behavior of the `notebook-preview` element:
- Adds a spinner while notebook is loading.
- Only shows the preview div if some markdown will actually render.
- Only says 'Showing the first n cells' if some markdown will actually render.

Also adds tests.

image

### DIFF
--- a/sources/web/datalab/polymer/test/file-browser-test.ts
+++ b/sources/web/datalab/polymer/test/file-browser-test.ts
@@ -13,31 +13,6 @@
  */
 
 window.addEventListener('WebComponentsReady', () => {
-  class MockFile extends DatalabFile {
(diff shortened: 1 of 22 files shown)


sources/web/datalab/polymer/test/file-browser-test.ts
```diff
@@ -100,6 +100,7 @@ window.addEventListener('WebComponentsReady', () => {
     });
 
     it('shows column names returned from the file manager in header', () => {
+      debugger;
```
jimmc: Remove

yebrahim: Done.

PR #1832, opened 2017-11-20
Don't delete /var/lib/dpkg/info/*

haavardw:
Removing these files prevents installation of other packages (e.g r-base).


chmeyers:
Any idea how big this directory is?  I'm just wondering how much this will increase the container size by.

chmeyers:
I'll check the size difference once the automatic build completes.

haavardw:
Thanks. To me it looks like the size of the directory is approx 9 MB:
```
$ docker run -ti 9841e8d65f0d du --max-depth=1 /var/lib/dpkg/info 
9636	/var/lib/dpkg/info
```

PR #1833, opened 2017-11-20
Use moment.js for relative date in item-list

yebrahim:
This PR adds using moment.js to display relative dates in item-list optionally. This is enabled by default, and can be turned off using an attribute.

image

This also addresses a comment from the last PR, to return 0 in the sort function if columns have not been defined yet.


nikhilk:
I tend to think this sort of coarser timestamps is useful for grouping but not for actual column values or sorting.

Strictly from a prioritization perspective, can we please just sort by type (folders first, files next) and by name within for now?

yebrahim:
Closing this for now, will open a separate issue to track sorting by type.

Issue #1839, opened 2017-12-06
Swap file initialization during start takes too long


chmeyers:
Especially evident on large memory instances, as the swap file size is based on the amount of memory.

Any way we can skip initializing the swap file with zeroes like that?

ojarjur:
I'm thinking this might be resolvable by making two changes:

1. Add a flag to the `create` command that disables the swapfile management on the instance. Users of high memory instances may not want swap enabled, and this would give them the option to turn it off.
2. Automatically disable the swapfile management if the size of memory on the instance is large relative to the disk size. For instance, if the swapfile would take more tha 10% of the disk, then we probably should not be configuring swap.

alexgorbachev:
Thanks @ojarjur

ojarjur:
Note that the fix will not be included in the Cloud SDK until the next Cloud SDK release (i.e. until version 184.0.0)

PR #1850, opened 2017-12-13
Don't create a swapfile if there is not enough disk space.

ojarjur:
This change adds a check to the Datalab instance startup script where
the swapfile creation/usage is prevented if the free disk space is not at
least 10x the size of memory.

This is the first step towards addressing #1839, but the full issue
won't be fixed until we also provide a way to turn off swap entirely
for users who do not want it.

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -131,7 +131,7 @@ configure_swap() {{
   # Specifically, if the free space on disk is not N times the
   # size of memory, then enabling swap makes no sense.
   #
-  # Arbitrarily choosing a value of N=10 
+  # Arbitrarily choosing a value of N=10
   disk_kb_cutoff=`expr 10 "*" ${{memory_kb}}`
   disk_kb_available=`df --output=avail ${{MOUNT_DIR}} | tail -n 1`
   if [ "${{disk_kb_available}}" -lt "${{disk_kb_cutoff}}" ]; then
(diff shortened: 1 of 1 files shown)


tools/cli/commands/create.py
```diff
@@ -126,6 +126,13 @@
   mem_total_line=`cat /proc/meminfo | grep MemTotal`
   mem_total_value=`echo "${{mem_total_line}}" | cut -d ':' -f 2`
   memory_kb=`echo "${{mem_total_value}}" | cut -d 'k' -f 1 | tr -d '[:space:]'`
+
+  disk_kb_cutoff=`expr 10 "*" ${{memory_kb}}`
```
jimmc: If I understand your calculations correctly, this value is 10 not because it means 10%, but because 10 happens to go into 100% 10 times. I think it would be more clear to calculate the disk size * 10 / 100 and compare that against memory size. And maybe call the variable swap_size instead of memory_size, or set another variable swap_size to make it clear that you are creating a swap file that happens to be the same size as the amount of memory.

ojarjur: No; the calculation is not trying to figure out if the swapfile size is < 10% of the disk size. Instead, it is trying to figure out if the disk size is > 10x the memory size.

The multiplier was chosen arbitrarily, but the check is about ensuring that the disk size is greater than some multiple of the memory size.

ojarjur: I've added a comment inline in the code to try to make the logic easier to follow.

PTAL

Issue #1857, opened 2017-12-16
item-list returns incorrect selected indices


yebrahim:
This happens when the items are sorted in a way that they're out of sync with the actual order they were given to the item-list element. The showed index vs real index calculation isn't correct.

PR #1859, opened 2017-12-18
Fix item-list index issues while sorting/filtering

yebrahim:
When sorting/filtering, the selected items' real indices should be distinguished from their render index. This PR fixes a bug where this wasn't the element's behavior, and the element returned the render indices of the items selected.

Additionally, this PR also fixes the behavior of Select/Unselect All to only work on the currently visible (filtered) items.

Fixes https://github.com/googledatalab/datalab/issues/1857.

### DIFF
--- a/sources/web/datalab/polymer/components/item-list/item-list.ts
+++ b/sources/web/datalab/polymer/components/item-list/item-list.ts
@@ -378,10 +378,13 @@ class ItemListElement extends Polymer.Element {
    * of indices of the currently selected items.
    */
   _computeSelectedIndices() {
-    const allElements = this.$.listContainer.querySelectorAll('paper-item') as NodeList;
-    return Array.from(allElements)
-        .filter((el: HTMLElement) => el.hasAttribute('selected'))
-        .map((el: HTMLElement) => this.$.list.modelForElement(el).itemsIndex);
+    const selected: number[] = [];
+    this.rows.forEach((row, i) => {
+      if (row.selected) {
+        selected.push(i);
+      }
+    });
+    return selected;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


sources/web/datalab/polymer/components/item-list/item-list.ts
```diff
@@ -451,63 +469,84 @@ class ItemListElement extends Polymer.Element {
...
+      this._unselectAllDisplayedItems();
     }
   }
 
   /**
    * On row click, checks the click target, if it's the checkbox, adds it to
    * the selected rows, otherwise selects it only.
+   * Note the distinction between displayIndex, which is the index of the item
+   * in the rendered list, and realIndex, which is the index of the item in the
+   * original list that was submitte to the item-list element. These might be
```
jimmc: typo: submitted

sources/web/datalab/polymer/test/item-list-test.ts
```diff
@@ -683,6 +699,26 @@ describe('<item-list>', () => {
         assert(columns[0].innerText === testFixture.rows[sortedOrder[i]].columns[0].toString());
       }
     });
+
+    it('returns the correct selectedIndices result matching clicked items', () => {
```
jimmc: Should there be a test that includes sorting?

yebrahim: This one does. It sorts the list first, reversing the order of the items 3 -> 2 -> 1, then selects the first one and makes sure its selected index is 2 not 0.

sources/web/datalab/polymer/components/item-list/item-list.ts
yebrahim: Done.

sources/web/datalab/polymer/test/item-list-test.ts
yebrahim: I can add one more for filtering.

PR #1863, opened 2017-12-19
Silence a misleading prompt from gcloud.

ojarjur:
This change prevents an issue where gcloud would issue a misleading
prompt when we ran `datalab delete`.

The issue was that gcloud would unconditionally issue a prompt
telling the user about its `--keep-disks` flag, which does not
appear directly in the `datalab delete` command (we instead have
a `--keep-disk` flag).

This change silences that prompt from gcloud and replaces it with
a custom prompt that we issue.

Since this is the second instance of the tool issuing a confirmation
prompt, I've moved the logic for issuing those into the utils package.

I've also made a modification to the `maybe_prompt_for_zone` method
so that the check for the `quiet` flag is nested deeper inside of
it. That allows the `--quiet` flag to work with instances that have
a single, obvious zone even if the `--zone` flag is not specified.


ojarjur:
@jimmc I caught a minor bug in the formatting of the error message for the missing-zone exception. PTAL

PR #1864, opened 2017-12-19
Fix notebook-app race condition

yebrahim:
This race condition showed up in long notebooks that had a lot of front-end resources to load, where by the time the `notebook-app.postLoad` executes, the `main.min.js` script had already tried to call the original `add_kernel_help_links`, which hasn't been overridden yet. This caused the UI to freeze completely.


yebrahim:
I moved the override code to the `notebook-app.preLoad` block, since it's specific to the notebook app only. PTAL.

PR #1865, opened 2017-12-20
Disable file search by default, except on GCP VMs

yebrahim:
We've seen a few users hit slowdown issues with a MacOS bug with the Chokidar package we're using for indexing, so we're disabling indexing by default unless the host is a GCP VM, which don't have the slowdown issue.

### DIFF
--- a/sources/web/datalab/fileSearch.ts
+++ b/sources/web/datalab/fileSearch.ts
@@ -16,12 +16,13 @@
 /// <reference path="../../../third_party/externs/ts/chokidar/chokidar.d.ts" />
 /// <reference path="common.d.ts" />
 
-import http = require('http');
-import url = require('url');
-import fs = require('fs');
-import path = require('path');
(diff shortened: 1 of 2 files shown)


sources/web/datalab/fileSearch.ts
```diff
@@ -131,7 +132,8 @@ function filter(pattern: string, data: string[]): string[] {
  */
 export function createHandler(settings: common.AppSettings): http.RequestHandler {
   appSettings = settings;
-  if (appSettings.enableFilesystemIndex) {
+  // Always enable the filesystem index on GCP VMs
```
ojarjur: I think we still want to be able to disable this on a VM.

I'd rather that the default be unspecified and we pick a different value on a VM vs. off of one. Then, if the setting has been explicitly set then we honor that setting.

yebrahim: Done.

PR #1866, opened 2017-12-22
Repackage some auth code into GapiManager.authManager.

jimmc:


sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -326,6 +326,78 @@ class GapiManager {
 
   };
 
+  public static authManager = class {
```
yebrahim: As discussed offline, if we rename this to just `auth` then it sounds more like the auth part of `GapiManager` than a separate authentication class.

PR #1871, opened 2017-12-27
Move gapi auth to separate class and define an interface for it.

jimmc:


jimmc:
You are correct, the changes you noted plus adding the interface are the only changes in this PR.

PR #1872, opened 2017-12-27
Add some cookie support.

jimmc:


sources/web/datalab/polymer/modules/utils/utils.ts
```diff
@@ -206,6 +208,26 @@ class Utils {
     return location.protocol + '//' + location.host;
   }
 
+  /**
+   * Returns the value for the named cookie.
+   */
+  public static readCookie(name: string) {
+    if (!Utils.cookies) {
```
yebrahim: Should the parsing be re-done every time a cookie's value is read instead of once? When do we expect this value to be set/used?

jimmc: Cookie values are normally set on a page load. I suppose it would be possible to set a cookie in an AJAX response. I don't think we are currently doing that, so it doesn't seem necessary.

yebrahim: SGTM.

PR #1873, opened 2017-12-28
Add support for server-side oauth (not yet enabled).

jimmc:


sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -277,6 +277,165 @@ class ClientAuth implements GapiAuth {
...
+  /**
+   * Signs the user out.
+   */
+  public signOut(): Promise<void> {
+    // We sign out by deleting our cookies.
+    Utils.deleteCookie('Datalab-access-token');
+    Utils.deleteCookie('Datalab-access-token-expiry');
+    Utils.deleteCookie('Datalab-refresh-token');
+    Utils.deleteCookie('Datalab-token-type');
+    this.setIsSignedIn(false);
```
yebrahim: We should reload or redirect the user to another page here.

sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -569,12 +728,12 @@ class GapiManager {
...
     }
-
   };
 
   /**
    * The auth class is responsible for initializing authorization, including
    * loading the gapi code and any data needed for oauth.
    */
   public static auth = new ClientAuth();
+  // public static auth = new ServerAuth();
```
yebrahim: Remove?

sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
yebrahim: Maybe we can make that an optional parameter, since I see `signOut` being called for another reason below other than the user initiating an explicit sign out.

jimmc: This method is called from only one place outside this file, from auth-panel, and that location calls reload right after calling signOut, so I don't think we need it here.

sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
jimmc: I added this comment to make it clear what needs to be done to switch from client-side auth flow to server-side auth flow. I could change it to a prose comment instead if that would be better.

yebrahim: I'll approve since I don't have a strong opinion on this.

PR #1875, opened 2017-12-29
Only propagate list item changes when selection state actually changes.

jimmc:
This avoids an n^2 problem in data browser on unselect-all.


jimmc:
Yes, in the Data Browser when there are a large number of projects, that n^2 problem was essentially making it hang when I would select something.

PR #1878, opened 2017-12-29
Don't attempt to load a preview for Views.

### LINKED ISSUE #1764: BQ Table preview doesn't work on views
Clicking a view in the Data tab gives this error: "Cannot list a table of type VIEW."
```
Failed to get table rows: {
  "error": {
    "code": 400,
    "message": "Cannot list a table of type VIEW.",
    "errors": [
      {
        "message": "Cannot list a table of type VIEW.",
        "domain": "global",
        "reason": "invalid"
      }
    ],
    "status": "INVALID_ARGUMENT"
  }
}
```

jimmc:
Fixes #1764.

### DIFF
--- a/sources/web/datalab/polymer/components/table-inline-details/table-inline-details.html
+++ b/sources/web/datalab/polymer/components/table-inline-details/table-inline-details.html
@@ -53,6 +53,7 @@ the License.
       <paper-spinner active hidden$={{!_busy}}></paper-spinner>
     </div>
     <div id="container" hidden$={{!_table}}>
+      <span hidden$=[[!_tableMessage]]>[[_tableMessage]]</span>
       <table id="tabledata">
         <tr class="header">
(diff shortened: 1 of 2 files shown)


sources/web/datalab/polymer/components/table-inline-details/table-inline-details.html
```diff
@@ -53,6 +53,7 @@
       <paper-spinner active hidden$={{!_busy}}></paper-spinner>
     </div>
     <div id="container" hidden$={{!_table}}>
+      <span hidden$=[[!_tableMessage]]>[[_tableMessage]]</span>
```
yebrahim: Do we actually need the `hidden` attribute here? If there's no message, it's an empty span, which is inlined anyway right?

jimmc: I think the intent is more clear this way.

PR #1888, opened 2018-01-04
Switch GPU instances over to using COS for the host OS instead of Ubuntu

ojarjur:


tools/cli/commands/creategpu.py
```diff
@@ -31,79 +31,121 @@
...
+    After=network-online.target gcr-online.target
+
+    [Service]
+    User=root
+    Type=oneshot
+    RemainAfterExit=true
+    EnvironmentFile=/etc/nvidia-installer-env
+    ExecStartPre=/bin/bash -c 'mkdir -p "${{NVIDIA_INSTALL_DIR_HOST}}" && mount --bind "${{NVIDIA_INSTALL_DIR_HOST}}" "${{NVIDIA_INSTALL_DIR_HOST}}" && mount -o remount,exec "${{NVIDIA_INSTALL_DIR_HOST}}"'
+    ExecStart=/usr/bin/docker run --privileged --net=host --pid=host --volume "${{NVIDIA_INSTALL_DIR_HOST}}":"${{NVIDIA_INSTALL_DIR_CONTAINER}}" --volume /dev:/dev --volume "/":"${{ROOT_MOUNT_DIR}}" --env-file /etc/nvidia-installer-env "${{COS_NVIDIA_INSTALLER_CONTAINER}}"
+    ExecStartPost=/bin/bash -c "${{NVIDIA_INSTALL_DIR_HOST}}/bin/nvidia-smi"
```
chmeyers: Why do we need to run this?  Is this just so the GPU info is in the logs?

tools/cli/commands/creategpu.py
```diff
@@ -169,43 +211,53 @@ def run(args, gcloud_beta_compute, gcloud_repos,
     if (not args.zone) and (not args.quiet):
         args.zone = utils.prompt_for_zone(args, gcloud_beta_compute)
     disk_cfg = create.prepare(args, gcloud_beta_compute, gcloud_repos)
+
     print('Creating the instance {0}'.format(args.instance))
     print('\n\nDue to GPU Driver installation, please note that '
```
chmeyers: With the new approach, how long does creation take?  Do we still need this admittedly vague message?

ojarjur: It still takes a long time; that's a fundamental consequence of downloading/installing the drivers after startup.

tools/cli/commands/creategpu.py
ojarjur: Removed.

PR #1889, opened 2018-01-05
Fix a bug where the repo cloning was being interrupted.

ojarjur:
This fixes an issue that appears to have only occurred with the
COS-based GPU instances, where the instance would fail to clone
the 'datalab-notebooks' repository.

The root cause was that the instance would reboot while the startup
script was running. That would interrupt it after the disk was
formatted, but before the repository was cloned. This, in turn
caused the repo to never be cloned, as subsequent startups would
see that the disk was already formatted.

This change tries to mitigate the likelihood of that happening
by moving the long running operation (downloading the Datalab
image to run the 'git clone' command) to before the disk is
formatted.

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -94,6 +94,8 @@ populate_repo() {{
 
 format_disk() {{
   echo "Formatting the persistent disk"
+  docker-credential-gcr configure-docker
+  docker pull {0}
   mkfs.ext4 -F \
     -E lazy_itable_init=0,lazy_journal_init=0,discard \
     ${{PERSISTENT_DISK_DEV}}
--- a/tools/cli/commands/creategpu.py
+++ b/tools/cli/commands/creategpu.py
@@ -148,11 +148,11 @@ write_files:
     RestartSec=1
 
 runcmd:
+- ['while', '[', '!', '-e', '/mnt/disks/datalab-pd/tmp', ']', ';',
+   'do', 'sleep', '1', ';', 'done']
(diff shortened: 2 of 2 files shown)


nikhilk:
Is there a way to avoid rebooting (which likely adds significantly to the startup latency)?

I saw some details to the effect of trying to avoid reboots in the Dataproc docs for their GPU support, and wonder if it might be applicable/interesting. See the shell script in https://cloud.google.com/dataproc/docs/concepts/compute/gpus

tools/cli/commands/create.py
```diff
@@ -94,6 +94,8 @@
 
 format_disk() {{
   echo "Formatting the persistent disk"
+  docker-credential-gcr configure-docker
+  docker pull {0}
```
jimmc: Are you doing the pull twice now? It looks like creategpu.py also still does the pull.

ojarjur: I don't see any explicit pulls in creategpu.py, but it would be fine even if it does.

Running `docker pull` is idempotent; as long as the image hasn't changed it won't pull it down again.

ojarjur:
@nikhilk I agree that avoiding reboots would be nice if possible, but as long as we are using the COS driver installer, we don't have any control over that.

Additionally, this fix would still be a good idea even if we could avoid the reboot, as it is still possible for the instance to reboot for other reasons, and this fix makes that less likely to cause issues.

ojarjur:
Added #1890 to track the "avoid-the-reboot" work.

PR #1897, opened 2018-01-09
Auto-refresh the access token.

jimmc:


sources/web/datalab/polymer/modules/gapi-manager/gapi-manager.ts
```diff
@@ -414,6 +428,74 @@ class ServerAuth implements GapiAuth {
...
+
+  /**
+   * Sends a request to the service to refresh our access token.
+   */
+  private async refreshToken(): Promise<void> {
+    const xhrOptions: XhrOptions = {
+      failureCodes: [401],
+      successCodes: [200],
+    };
+    const refreshPath = '/_auth/refresh';
```
yebrahim: I'd prefer we add these to the `ServiceId` enum instead of using the string. We kind of missed them in the last PR when refactoring `GapiManager` too.

jimmc: At the moment, I think of this as an internal implementation detail of the server-side auth code, so it seems cleaner to keep it inside ServerAuth. But I can see an argument for putting it in ServiceId.

PR #1900, opened 2018-01-10
Add debug log method and a few debug log messages.

jimmc:


sources/web/datalab/polymer/modules/utils/utils.ts
```diff
@@ -24,10 +24,21 @@ class Utils {
       // tslint:disable-next-line:no-console
       console.log(...args);
     }
+
     public static error(...args: any[]) {
       // tslint:disable-next-line:no-console
       console.error(...args);
     }
+
+    // To enable debug messages, set DATALAB_LOG_DEBUG=1 in your browser console
```
yebrahim: Can we make this comment a little clearer? We're setting `DATALAB_LOG_DEBUG=1` as a cookie not a `window` property right?

yebrahim: Also, why don't we make this change in `Utils.log.verbose`? I always thought `debug` and `verbose` log levels are the same?

jimmc: Are you saying that, instead of adding a debug method, I should make the verbose method not print out anything unless the cookie is set? That sounds reasonable to me. Do you think the cookie name should be COOKIE_LOG_VERBOSE, or should it just be something like COOKIE_LOG_MODE and we look for the value "verbose" to set verbose logging?

yebrahim: Yes, I think the verbose method should check the cookie.
Putting the log level in the cookie sounds like a good idea, but I don't see a lot of need for it with our limited logging, so it's up to you.

jimmc: Done.

PR #1901, opened 2018-01-11
Stop caching cookies so AJAX cookie updates are seen.

jimmc:


sources/web/datalab/polymer/modules/utils/utils.ts
```diff
@@ -220,9 +220,7 @@ class Utils {
    * Returns the value for the named cookie.
    */
   public static readCookie(name: string) {
-    if (!Utils.cookies) {
-      Utils.parseCookies();
-    }
+    Utils.parseCookies();
```
yebrahim: Can this cause performance issues?
We'll be calling this very frequently, for example because of the logging changes we just made. I remember logging was very noisy for some features, and now every log call will read and parse cookies, even in a deployment where logging is disabled.

jimmc: Good point. I changed it to allow passing a reload flag for cookies that might change.

Issue #1898, opened 2018-01-10
External PD not attached/identified


VikramTiwari:
At times while connecting to datalab, the external persistent disk is not identified properly. It resolves after a reset of the machine though but is concerning when encountering for the first time. I think that this might be happening when the automatic timeout (1h30m) happens and then I retry to connect to the instance.

chmeyers:
By "identified" you mean that you can't access files on that PD?

This might be an the mount command having an issue.  A little concerning since if it fails we might attempt to format the disk: https://github.com/googledatalab/datalab/blob/master/tools/cli/commands/create.py#L112, but as it works after a reset, we obviously aren't doing a format.

VikramTiwari:
Yes, the datalab page opens up but missing all the files on PD. Seems like it could be an issue with the mount part. Also, just straight up formatting PD doesn't seem like a great idea. Any destructive action should require user consent.

ojarjur:
> Also, just straight up formatting PD doesn't seem like a great idea.

I agree, we can do better here. The format is only meant to happen when attaching an unformatted disk for the first time.

Unfortunately, we don't have a way of checking whether or not this is the first time the disk was attached, but it seems we can explicitly check if it is formatted using the `blkid` command.

I'm going to send out a PR to add that explicit check prior to formatting the disk. I'll also see if I can add some check to verify the disk has been mounted before we proceed with starting the Datalab instance.

chmeyers:
Maybe we should touch a hidden file after mounting it for the first time and check and see if that file exists?

ojarjur:
@chmeyers I think the `blkid` check will be better because it will check exactly what we want to check; whether or not the disk is formatted.

ojarjur:
It looks like the release we included in version 185.0.0 of the Cloud SDK changed this issue from being a rare occurrence to always happening.

Fortunately, we already had the fix for it; but that will not go out until version 186.0.0 of the Cloud SDK, which is expected later this week.

In the mean time, if you are hit by this issue (i.e. you only see an empty "notebooks" directory in your instance and no "docs" directory), you can fix it by restarting your instance.

jimmiebtlr:
Maybe worth it to publicize this a bit more or something?  We definitely saw this and were looking at recovery options for a while before coming across this ticket.

chmeyers:
This is fixed with gcloud 186.0.0 (went out Wednesday)

PR #1903, opened 2018-01-11
Improvements to startup reliability.

ojarjur:
This tries to make the startup logic for mounting (and, if necessary,
formatting) the persistent disk more reliable. It does that by
making two changes:

1. Verify that the disk has not already been formatted before
   proceeding with the format command.
2. Configure the Datalab systemd service to not start until after
   we have verified that the startup script has finished creating
   the directories that Datalab needs.

The goal of this change is to prevent the sporadic failures reported
in #1898

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -111,7 +111,7 @@ checked_format_disk() {{
   if [ -z "$(blkid ${{PERSISTENT_DISK_DEV}})" ]; then
     format_disk
   else
-    echo "Disk already formatted, but mounting failed"
+    echo "Disk already formatted, but mounting failed; rebooting..."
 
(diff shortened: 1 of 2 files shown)


tools/cli/commands/create.py
```diff
@@ -106,10 +106,29 @@
   fi
 }}
 
+checked_format_disk() {{
+  echo "Checking if the persistent disk needs to be formatted"
+  if [ -z "$(blkid ${{PERSISTENT_DISK_DEV}})" ]; then
+    format_disk
+  else
+    echo "Disk already formatted, but mounting failed"
```
jimmc: I think when you reboot, it might be helpful to output a message to that effect.

ojarjur: Done.

PR #1916, opened 2018-01-17
Verification whether firewall exists added

lkuligin:


ojarjur:
Thanks for contributing and putting in the effort to try to improve the tool!

I need a bit more context so I can better understand the intended effects of this change. Can you give some background on what your underlying goal is?

In particular, what is the scenario that currently fails that you want us to support?

lkuligin:
Sometimes the _...-allow-ssh_ firewall rule can be deleted (e.g., by the audit rule that deletes the unused firewall rules that are too wide). In this case the `datalab connect` command could not establish the connection, and the description error is not useful.

ojarjur:
To close the loop on this; I chatted with @lkuligin out of band, and we're not going to move forward with this exact change.

Checking if the network and firewall rules exist is good, but if they do not then we should at most be reporting an error rather than creating them; as the `datalab connect` command is not expected to have a side effect of changing the project's network setup.

PR #1919, opened 2018-01-18
h5py added for non-gpu image

lkuligin:
Otherwise tensorflow.keras throws an error "ImportError: `save_model` requires h5py." and even prototyping is not possible.


yebrahim:
With this change, does `import h5py` work in a notebook? There was an open issue about this in the past. See https://stackoverflow.com/a/43490376/1768702.

nikhilk:
Going off the title, we should make sure the GPU image is also kept in sync.

lkuligin:
1. yes, running `!pip install h5py`and then restarting the kernel and redoing imports helps
2. gpu image already has h5py

chmeyers:
Re Nikhilk's comment: The GPU image already had h5py, it was one of 3 extra packages that we installed there, along with tensorflow-gpu and tflearn.  As the GPU image inherits from the non-GPU image, it will always get anything that the non-GPU image has.

PR #1941, opened 2018-02-05
Switching to Tensorflow 1.5

chmeyers:
Switching to Tensorflow 1.5

Updating to CUDA 9 and cuDNN v7 for GPU images as that version is
required by TF 1.5, this has the additional benefit of adding support
for V100 (Volta) GPUs.


chmeyers:
@rileyjbauer, please take another look. (It's actually the same as the first commit again.)

PR #1942, opened 2018-02-05
Fix build break

yebrahim:
For some reason, npm decided it didn't want to install one of the nested dependencies: @types/assert. This is a direct dependency of polyserve, which is a dependency of polymer-cli, which in turn is a dependency of Datalab. This reproduces locally and on travis, and I'm not sure why this only happens with the latest polymer-cli (1.6.0).

I've also updated a few other dependencies, and put back the good old `package-lock.json` into git so it can help make these investigations easier in the future. (There's no need to review this file.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 *.ipynb_checkpoints
 .DS_Store
-package-lock.json
 build
 staging
 *.gradle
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,6 +1,5 @@
 *.ipynb_checkpoints
 .DS_Store
-package-lock.json
```
chmeyers: Double checking: This was on purpose?

yebrahim: Yup.

PR #1952, opened 2018-02-12
Always print out the nested command on failure.

chmeyers:
This way users will always have at least something to look at when we report that a gcloud command fails.

### DIFF
--- a/tools/cli/datalab.py
+++ b/tools/cli/datalab.py
@@ -357,8 +357,10 @@ def run():
             email=get_email_address(),
             in_cloud_shell=('DEVSHELL_CLIENT_PORT' in os.environ),
             gcloud_zone=gcloud_zone)
-    except subprocess.CalledProcessError:
+    except subprocess.CalledProcessError as e:
         print('A nested call to gcloud failed.')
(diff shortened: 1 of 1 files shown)


tools/cli/datalab.py
```diff
@@ -357,8 +357,10 @@ def run():
             email=get_email_address(),
             in_cloud_shell=('DEVSHELL_CLIENT_PORT' in os.environ),
             gcloud_zone=gcloud_zone)
-    except subprocess.CalledProcessError:
+    except subprocess.CalledProcessError as e:
         print('A nested call to gcloud failed.')
+        print('Command: ["' + '","'.join(e.cmd) + '"]')
```
ojarjur: I feel like this should be under an `if utils.print_debug_messages(args):` condition, so that we don't doubly spam users when the cause of the failure has already been reported.

I also think we should probably make the `get_gcloud_zone` and `get_email_address` methods print their stdout/stderr if they fail.

chmeyers: Done & Done.

chmeyers:
Fixed flake8 errors.  PTAL

PR #1956, opened 2018-02-13
Add some basic version information reporting:

ojarjur:
1. Add the version of the Cloud SDK and datalab component used to
   create an instance to that instance's metadata.
2. When connecting to an instance using the `--diagnose-me` flag,
   report the version info from that instance's metadata entries.

### DIFF
--- a/tools/cli/commands/connect.py
+++ b/tools/cli/commands/connect.py
@@ -363,9 +363,19 @@ def run(args, gcloud_compute, email='', in_cloud_shell=False, **unused_kwargs):
     status, metadata_items = utils.describe_instance(
         args, gcloud_compute, instance)
     for_user = metadata_items.get('for-user', '')
+    sdk_version = metadata_items.get('created-with-sdk-version', 'UNKNOWN')
+    datalab_version = metadata_items.get(
+        'created-with-datalab-version', 'UNKNOWN')
     if (not args.no_user_checking) and for_user and (for_user != email):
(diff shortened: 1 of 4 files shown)


tools/cli/commands/connect.py
```diff
@@ -363,10 +363,20 @@ def run(args, gcloud_compute, email='', in_cloud_shell=False, **unused_kwargs):
...
     if (not args.no_user_checking) and for_user and (for_user != email):
         print(wrong_user_message_template.format(for_user, email))
         return
 
+    if args.diagnose_me:
+        print('Instance {} was created with the following '
+              'Cloud SDK component versions:'
+              '\n\tCloud SDK: {}'
+              '\n\tDatalab: {}'.format(
+                  instance, sdk_version, datalab_version))
```
chmeyers: Should we also print the current datalab/gcloud version, to help diagnose whether connect problems are due to a bad local version?

ojarjur: That's already done in the `datalab.py` file.

An example output of running `datalab connect --diagnose-me ...` is:

```
Running with diagnostic messages enabled
Using the command "gcloud" to invoke gcloud
The installed gcloud version is:
        Cloud SDK: 188.0.0
        Datalab: 20180119
Instance ... was created with the following Cloud SDK component versions:
        Cloud SDK: 188.0.0
        Datalab: 20180119
Connecting to ...
```

PR #1958, opened 2018-02-13
Add a facility to warn users of versions with known issues.

ojarjur:
This change was prompted by the recent issue we had with version
20180105, where an issue was discovered, but we did not have a
good means of notifying users about the issue.

With this change, the CLI will download a versions info file from
https://storage.googleapis.com/cloud-datalab/version-issues.js
that contains a JSON object reporting known issues for any
affected versions of the CLI. It will then check the current
version against that object and report any matching warnings.

### DIFF
--- a/tools/cli/datalab.py
+++ b/tools/cli/datalab.py
@@ -26,6 +26,7 @@ import json
 import os
 import subprocess
 import traceback
+import urllib2
 
 
 _SUBCOMMANDS = {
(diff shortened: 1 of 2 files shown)


tools/cli/datalab.py
```diff
@@ -113,6 +114,11 @@
 datalab_component = 'datalab'
 
 
+# Public file reporting all known version issues.
+version_issues_url = (
+    'https://storage.googleapis.com/cloud-datalab/version-issues.js')
```
chmeyers: Does this file already exist?

ojarjur: Yes, it does. I populated it with information about the 20180105 release so that I could verify the warnings appear correctly.

For example, if you run `gcloud components update --version 185.0.0`, and then run the version of this code from this branch, you will get the following result:

```
$ ./tools/cli/datalab.py list
You are using the Datalab CLI version "20180105", which has the following known issues:
        Instances fail to mount their notebooks disk on first boot. The root cause is a race condition first reported here: https://github.com/googledatalab/datalab/issues/1898. The issue is fixed in versions 20180119 and later. As a workaround, failing instances will correcly mount their notebooks disk after being reset.
Listed 0 items.
```

PR #1961, opened 2018-02-14
Fix a leak in the GCS Backup utility.

### LINKED ISSUE #1960: GCS Backup Issue
We noticed our /tmp directory inside the datalab docker instance filled up our persistent disk with gigs of "tmp.k3j4589/archive.zip" files. I tracked it down to GCSBackup.sh, which was successfully making a 40MB zip file each time it ran, but still output an error, causing the backup script to think it wasn't successful, and therefore it would run again, in a vicious cycle:

```
$ head /datalab/.backup_log.txt
20180211233656: Running GCS backup tool..
20180211233656: Running GCS backup tool..
20180211233656: Running GCS backup tool..
Failed creating the backup archive
Failed creating the backup archive
Failed creating the backup archive
20180211234655: Running GCS backup tool..
20180211234655: Running GCS backup tool..
20180211234655: Running GCS backup tool..
Failed creating the backup archive
Failed creating the backup archive
Failed creating the backup archive
20180211235655: Running GCS backup tool..
20180211235655: Running GCS backup tool..
20180211235655: Running GCS backup tool..
Failed creating the backup archive
Failed creating the backup archive
Failed creating the backup archive
20180212000655: Running GCS backup tool..
20180212000655: Running GCS backup tool..
20180212000655: Running GCS backup tool..
Failed creating the backup archive
```

I ran the zip command in verbose mode to see what was going on, and it seems to be the "forever" command's sock file that was throwing an error when trying to be read:

```
$ cd /content
$ zip -rv /tmp/wow.zip .
...
updating: .jupyter/	(in=0) (out=0) (stored 0%)
updating: .jupyter/nbconfig/	(in=0) (out=0) (stored 0%)
updating: .jupyter/nbconfig/notebook.json	(in=71) (out=61) (deflated 14%)
updating: .viminfo	(in=658) (out=308) (deflated 53%)
updating: .gitconfig	(in=120) (out=72) (deflated 40%)
  adding: .forever/sock/worker.15183916117837xa.sock
zip warning: No such device or address
	zip warning: could not open for reading: .forever/sock/worker.15183916117837xa.sock

zip warning: Not all files were readable
  files/entries read:  423 (133M bytes)  skipped:  1 (0 bytes)
total bytes=140321049, compressed=44816814 -> 68% savings
```

I'm going to disable auto-backups for now, but I thought I'd bring this to your attention.

Here's the culprit line: https://github.com/googledatalab/datalab/blob/master/containers/datalab/content/GCSbackup.sh#L144

Maybe exclude forever folder via:

```
zip -rq ${archive_name} "${backup_path}" -x *.forever*
```

ojarjur:
This fixes #1960

That bug appears to be caused by the zip command encountering what
looks like a socket file under the `/content/.forever` directory,
trying to add it to the archive, and then failing. Since the
zip command reported a failure, the GCSBackup.sh script then
exits without deleting the locally-created archive.

As such, there seem to be two bugs at play here:

1. The `zip` command fails when processing the `.forever` dir.
2. If the `zip` command reports an error, we do not delete the
   temporary archive even if it was partially created.

This change fixes that by making two changes:

1. Excluding the `.forever` directory from the zip file
2. Setting up a `trap` command to delete the temp directory as
   soon as that directory has been created.

### DIFF
--- a/containers/datalab/content/GCSbackup.sh
+++ b/containers/datalab/content/GCSbackup.sh
@@ -139,9 +139,11 @@ echo
 echo "${timestamp}: Running GCS backup tool.." | tee -a ${log_file}
 
 # create an archive of the backup path
-archive_name=$(mktemp -d)"/archive.zip"
+tempdir="$(mktemp -d)"
+archive_name="${tempdir}/archive.zip"
+trap "rm -rf ${tempdir}" EXIT
 echo "Creating archive: $archive_name"
-zip -rq ${archive_name} "${backup_path}" || {
+zip -rq ${archive_name} "${backup_path}" -x '*/.forever/*' || {
   echo "Failed creating the backup archive" | tee -a ${log_file}
   exit 1
 }
@@ -171,13 +173,11 @@ echo "New archive md5 hash: ${new_backup_hash}"
 echo "Last backup md5 hash: ${last_backup_hash}"
 if [[ $new_backup_hash == $last_backup_hash ]]; then
   echo "Hash not different from last backup. Skipping this backup round." | tee -a $log_file
-  rm -f "${archive_name}"
   exit 0
 fi
 
 # copying backup to GCS
 gsutil cp ${archive_name} "gs://${backup_id}"
-rm -f "${archive_name}"
 
 # remove excessive backups
 all_backups=($(gsutil ls "gs://${gcs_bucket}/datalab-backups/${zone}/${machine_name}${backup_path}/${tag}-*"))


containers/datalab/content/GCSbackup.sh
```diff
@@ -139,9 +139,11 @@ echo
 echo "${timestamp}: Running GCS backup tool.." | tee -a ${log_file}
 
 # create an archive of the backup path
-archive_name=$(mktemp -d)"/archive.zip"
+tempdir="$(mktemp -d)"
+archive_name="${tempdir}/archive.zip"
+trap "rm -rf ${tempdir}" EXIT
 echo "Creating archive: $archive_name"
-zip -rq ${archive_name} "${backup_path}" || {
+zip -rq ${archive_name} "${backup_path}" -x '*/.forever/*' || {
```
harmon: I think you have an extra "/" here at the beginning that you don't want. Did you mean "\"?

ojarjur: I did intend for the slash to be there because I wanted to exclude the "/content/.forever" dir, but not anything like "/content/this.directory.should.be.backed.up.forever".

Hence, the leading slash.

I tested this locally with a ".forever" directory and confirmed that adding/removing/changing files under that directory did not alter the contents of the backup.

Is there an issue you think the leading slash will cause?

harmon: Sorry, I was testing this when actually in the directory, like:

$ cd /content
$ zip -rq archive.zip . -x '*/.forever/*'

In which case it would show up as ".forever" and wouldn't match, but here we don't cd to /content first, so it would be "/content/.forever". So your way is all good. Carry on!

PR #1970, opened 2018-02-23
 Use the pip version of h5py instead of the conda version

chmeyers:
Also aliasing pip3 to pip for the python 3 environment.

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -110,7 +110,7 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
         ggplot==0.6.8 \
         google-cloud-dataflow==2.0.0 \
         lime==0.1.1.23 \
-        tensorflow==1.5.0 && \
+        tensorflow==1.5.0 \
         h5py==2.7.1 && \
     source deactivate && \
 # Clean up before setting up the Python3 env.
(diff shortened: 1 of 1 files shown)


chmeyers:
Note: This switch is because the Conda version of h5py doesn't appear to install cleanly for us and gives this error when imported:
/usr/local/envs/py3env/lib/python3.5/site-packages/h5py/__init__.py in <module>()
     60 from .version import version as __version__
     61 
---> 62 from .tests import run_tests
     63 
     64 if version.hdf5_version_tuple != version.hdf5_built_version_tuple:

ImportError: No module named 'h5py.tests'

gupta17anukriti:
I am facing a similar issue with h5py 

```
> --------------------------------------------------------------------------
> ImportError                               Traceback (most recent call last)
> <ipython-input-1-e81388eeaef8> in <module>()
> ----> 1 import h5py
> 
> C:\Users\agupta82\AppData\Local\conda\conda\envs\my_root\lib\site-packages\h5py\__init__.py in <module>()
>      24 # We tried working around this by using "package_dir" but that breaks Cython.
>      25 try:
> ---> 26     from . import _errors
>      27 except ImportError:
>      28     import os.path as _op
> 
> ImportError: DLL load failed: %1 is not a valid Win32 application.
```

PR #1983, opened 2018-04-06
Use Anaconda base packages (with MKL support) by default

mvanwyk:
Changed the command adding the conda-forge channel from an  --add to an --append so that it doesn't override the default conda channel. This will install base packages before conda-forge ones, the opposite of which is happening now. Performance wise this means MKL pacakges will be installed rather than openblas, which should make for a nice speed boost.

For more info see - https://conda.io/docs/user-guide/tasks/manage-channels.html#after-conda-4-1-0


mvanwyk:
Hmm.. not sure why that's failing.. a quick test show it working on my end.. additionally I see I commited with the wrong user.. I'll redo and use travis-ci to check

PR #1985, opened 2018-04-10
Use packages from conda defaults channel first

mvanwyk:
The current installation of conda uses --add to add the conda-forge channel which results in the demotion of packages from the defaults channel. Since conda-forge packages use open_blas (non-mkl) versions of packages such as numpy are installed affecting performance. 

By first updating conda to a version post 4.1 we can use --append to add the conda-forge channel so that default packages still take precedence. Then mkl versions of packages will be installed by default. This should result in a nice performance boost for many users.


rileyjbauer:
Thanks for the tip here!

Just for the record, after this change when one uses the python2 kernel and runs:

    !jupyter kernelspec list

the result is:

    Available kernels:
      python2    /usr/local/envs/py2env/lib/python2.7/site-packages/ipykernel/resources

whereas before this change, the same command in the python2 kernel resulted in:

    Available kernels:
      python2    /usr/local/envs/py2env/share/jupyter/kernels/python2

The above change doesn't seem to have negatively affected anything though (kernel switching still changes both python runtime and conda env, and conda installs still work)

PR #1994, opened 2018-04-25
Update nvidia driver version

Plazmius:
Hi, I have tried to create instance of gpu-based datalab, but unfortunately instance start is failed, because link for NVIDIA drivers no longer works. But, there is 390.46 version available, so if change it, instance is working fine.


yebrahim:
Thanks for the fix @Plazmius, I'll merge this.

PR #1997, opened 2018-05-01
Upgrade the GPU images to use version 9.1 of the CUDA toolkit.

ojarjur:
PR #1994 upgraded GPU instances to use version 390.46 of the
NVIDIA drivers, but that version is meant to be used against
version 9.1 of the CUDA toolkit and we were using version
9.0 of the toolkit.

This change fixes this by upgrading to the base image that
includes the newer version of the CUDA toolkit.

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -87,6 +87,7 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
         pandas==0.22.0 \
         pandocfilters==1.4.2 \
         pillow==5.0.0 \
+        pip==9.0.3 \
         plotly==1.12.5 \
         psutil==4.3.0 \
         pygments==2.1.3 \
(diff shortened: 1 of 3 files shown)


containers/base/Dockerfile
```diff
@@ -87,6 +87,7 @@ RUN echo "deb-src http://ftp.us.debian.org/debian testing main" >> /etc/apt/sour
         pandas==0.22.0 \
         pandocfilters==1.4.2 \
         pillow==5.0.0 \
+        pip==9.0.3 \
```
yebrahim: Just curious, newer pip versions cause issues or is this just speculative?

ojarjur: The newer pip version broke the build.

There is a warning during our build about uninstalling a distutils package (it upgrades "futures" from 3.0.3 to 3.0.5) being only a partial uninstall.

With pip 10, that warning was converted to an error.

PR #2003, opened 2018-05-07
Make the end-to-end CLI test use a free port for connections.

ojarjur:
This change extends the end-to-end test for the CLI so that instead
of always using port 8081 (the default), it picks a port that was
free at the time the connection was being created.

This is meant to fix a recent regression in the tests where it would
fail in our Cloud Builder runs due to the port not being available.

There is still a race condition where the port could be grabbed by
another process between the time that it is picked and the time that
the connection is created, but that is not worth worrying about.

### DIFF
--- a/tools/cli/tests/end-to-end.py
+++ b/tools/cli/tests/end-to-end.py
@@ -18,6 +18,7 @@
 # of the bundled CLI tool. This requires a GCP project in which the
 # test will create, connect to, and delete Datalab instances.
 
+import socket
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


tools/cli/tests/end-to-end.py
```diff
@@ -171,8 +185,8 @@ def test_connect(self):
             self.assertIn('TERMINATED', instance.status())
             with tempfile.NamedTemporaryFile() as tmp:
                 with DatalabConnection(self.project, self.zone,
-                                       instance.name, tmp):
-                    readme = urlopen(readme_url)
+                                       instance.name, tmp) as conn:
+                    readme = urlopen(conn.readme_url)
```
yebrahim: It seems there are some cases where this wouldn't be defined?

ojarjur: Yes; if the connection fails. In that case we want to raise an exception, so I figured it was fine to just let it raise the exception here.

yebrahim: I see, lgtm then.

PR #2004, opened 2018-05-08
Pick a random zone for CLI tests.

ojarjur:
This changes the CLI end-to-end test so that it defaults to
picking a random zone in the region 'us-west' instead of being
hard-coded to 'us-west1-a'.

The test will still honor the 'zone' config setting in gcloud if
it is set; this only changes the behavior when that is not set.

### DIFF
--- a/tools/cli/tests/end-to-end.py
+++ b/tools/cli/tests/end-to-end.py
@@ -18,6 +18,7 @@
 # of the bundled CLI tool. This requires a GCP project in which the
 # test will create, connect to, and delete Datalab instances.
 
+import random
 import socket
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


tools/cli/tests/end-to-end.py
```diff
@@ -58,6 +59,15 @@ def free_port():
     return port_number
 
 
+def random_zone():
+    zones_list = subprocess.check_output([
+        'gcloud', 'compute', 'zones', 'list',
+        '--filter=region~us-west', '--format=value(name)']).decode(
```
yebrahim: Should this be `--format=value("name")`?

ojarjur: No; the string is already in quotes, so nested quotes would be passed directly to the format handler, which would not know what to do with them.

yebrahim: Ah, what threw me off is I didn't know this actually executed as 'gcloud' 'compute'... etc (with the quotes). It fails for me without the quotes.

PR #2011, opened 2018-05-23
Reject CORS websocket requests

yebrahim:


sources/web/datalab/server.ts
```diff
@@ -299,6 +299,15 @@ function stopVmHandler(request: http.ServerRequest, response: http.ServerRespons
 }
 
 function socketHandler(request: http.ServerRequest, socket: net.Socket, head: Buffer) {
+  // Websocket requests aren't CORS-checked by the browser. Reject any CORS
+  // requests here by checking their host vs origin request headers.
+  if (request.headers.origin &&
+    request.headers.origin.replace(/^.*\/\//, '') !== request.headers.host) {
```
ojarjur: I can't understand what this replacement is doing. Can you either document this or rework it to be easier to interpret?

yebrahim: Ah, forgot to add the comment. Done.

sources/web/datalab/server.ts
```diff
@@ -299,6 +299,17 @@ function stopVmHandler(request: http.ServerRequest, response: http.ServerRespons
 }
 
 function socketHandler(request: http.ServerRequest, socket: net.Socket, head: Buffer) {
+  // Websocket requests aren't CORS-checked by the browser. Reject any CORS
+  // requests here by checking their host vs origin request headers.
+  // Origin headers contain the protocol, while host headers do not, so we strip
+  // it off the origin header before matching.
+  if (request.headers.origin &&
+    request.headers.origin.replace(/^.*\/\//, '') !== request.headers.host) {
```
ojarjur: Can we reduce this check to trimming the literal prefix of `ws://`?

I would rather err on the side of being overly restrictive rather than overly permissive.

yebrahim: Sure.
Do we see any instances where this can be `ws://`?

PR #2013, opened 2018-05-25
Compare against referer header if no origin header present

yebrahim:


sources/web/datalab/server.ts
```diff
@@ -303,13 +303,24 @@ function trimProtocol(requestPath: string) {
   return start > -1 ? requestPath.slice(start + '://'.length) : requestPath;
 }
 
-function socketHandler(request: http.ServerRequest, socket: net.Socket, head: Buffer) {
-  // Websocket requests aren't CORS-checked by the browser. Reject any CORS
-  // requests here by checking their host vs origin request headers.
+function requestIsCrossOrigin(request: http.ServerRequest) {
```
ojarjur: The name is inverted from the behavior. This is checking that the request is not cross origin rather than that it is.

I would suggest keeping the name and inverting the behavior (since it would be confusing for the check to be `!requestIsNotCrossOrigin(...)`)

sources/web/datalab/server.ts
```diff
@@ -303,13 +303,24 @@ function trimProtocol(requestPath: string) {
...
   // Origin headers contain the protocol, while host headers do not, so we strip
   // it off the origin header before matching.
-  if (request.headers.origin &&
-    trimProtocol(request.headers.origin) !== request.headers.host) {
+  // Referer headers also contain the protocol, but they also contain the full
+  // pathname, so we extract the origin.
+  if (request.headers.origin) {
+    return trimProtocol(request.headers.origin) === request.headers.host;
+  } else if (request.headers.referer) {
+    return trimProtocol(new URL(request.headers.referer).origin) === request.headers.host;
```
ojarjur: Can we reduce this to `new URL(request.headers.referer).host` instead of getting the origin and then trimming the protocol.

Similarly, could we use the URL constructor to replace the `trimProtocol` logic, if it is already capable of parsing URLs?

sources/web/datalab/server.ts
yebrahim: Woops, fixing.

sources/web/datalab/server.ts
yebrahim: Ah sweet, didn't know that existed. Thanks.

Issue #1924, opened 2018-01-23
Feature request: Add --subnet flag in datalab create


charlesverdad:
I want to attach my datalab instance into a subnetwork I created but I can't find a way to do it. The `gcloud compute instances create` command has a `--subnet` flag. It would be great if that same flag can be used in the `datalab create` command.

> --subnet=SUBNET
> Specifies the subnet that the instances will be part of. If --network is also specified subnet must be a subnetwork of network specified by --network.
> 

**EDIT**: added additional info below
If the VPC network is in custom mode, there's currently no way to create a datalab instance into that network because it gives this error:
```
Creating the instance my-datalab
ERROR: (gcloud.compute.instances.create) Could not fetch resource:
 - Invalid value for field 'resource.networkInterfaces[0]': ''. Subnetwork should be specified for custom subnetmode network
A nested call to gcloud failed.
```

chmeyers:
This is basically https://github.com/googledatalab/datalab/issues/1535 but "even more options".  I think this is more evidence that we should just allow arbitrary pass-through arguments that go straight to 'gcloud compute instances create'

nikhilk:
Would we need some options for creating an instance, vs. some for connecting, vs. some for creating the repo, or any of the N steps we do etc? At the risk of complicating things with generalization, maybe we need a --create-flags and a --connect-flags, and so on?

charlesverdad:
I like the --create-flags and --connect-flags options.

Update:
Because I converted my main VPC network into custom subnet mode (irreversible), I can no longer create datalab instances that are connected to my main instances. I can't even make a separate VPC network set to auto and use VPC peering because their subnets overlap.

chmeyers:
FYI charlesverdad if you are blocked while we are debating this, you can grab a copy of the create command from https://github.com/googledatalab/datalab/blob/master/tools/cli/commands/create.py#L746 and edit it to specify a subnet (at around line 746).

ojarjur:
This is now fixed with version 214.0.0 of the Cloud SDK

HeshamMahrous:
I updated gcloud componets to the latest 214 and it still get Creating the instance hesh
ERROR: (gcloud.compute.instances.create) Could not fetch resource:
 - Invalid value for field 'resource.networkInterfaces[0]': ''. Subnetwork should be specified for custom subnetmode network

A nested call to gcloud failed, use --verbosity=debug for more info.

PR #2015, opened 2018-06-01
Implementation for Feature Request #1924 - Add --subnet flag

bgood:
I needed to target subnets for my datalab create commands so I did a quick implementation of the feature request.


ojarjur:
Thanks for taking the time and putting in the work to improve the project!

This has been previously requested (see #1924), so we know this will help others too.

Looking through the code, I am worried about how this will behave when the `--subnet-name` flag is not specified. I'll leave more detailed comments inline.

tools/cli/commands/create.py
```diff
@@ -373,6 +388,11 @@ def flags(parser):
         dest='network_name',
         default=_DATALAB_NETWORK,
         help='name of the network to which the instance will be attached.')
+    parser.add_argument(
+        '--subnet-name',
+        dest='subnet_name',
+        default=_DATALAB_SUBNET,
```
ojarjur: I would recommend not providing a default value, so that later on we can check whether or not any value was provided with the conditional:

```py
if args.subnet_name is not None:
    ...
```

tools/cli/commands/create.py
```diff
@@ -670,7 +711,9 @@ def prepare(args, gcloud_compute, gcloud_repos):
       subprocess.CalledProcessError: If a nested `gcloud` calls fails
     """
     network_name = args.network_name
+    subnet_name = args.subnet_name
     ensure_network_exists(args, gcloud_compute, network_name)
+    ensure_subnet_exists(args, gcloud_compute, subnet_name)
```
ojarjur: We should only run this check if the `--subnet-name` flag was provided.

tools/cli/commands/create.py
```diff
@@ -773,6 +816,7 @@ def run(args, gcloud_compute, gcloud_repos,
                 '--format=none',
                 '--boot-disk-size=20GB',
                 '--network', args.network_name,
+                '--subnet', args.subnet_name,
```
ojarjur: It's important that this be conditional on the `--subnet-name` flag being provided.

For instance, if someone wants to create a Datalab instance connected to a legacy network, then they can't have any value for the `--subnet` flag.

bgood:
The PR has been updated to only run the subnet check method and include the subnet option on commands when the `--subnet-name` is provided.

rtadewald:
How can I add this parameter to my already installed "datalab"? 

After ``gcloud components update``, `using datalab create --subnet ...` keeps me returning "unknow method"...

ojarjur:
@rtadewald This change has been submitted, but is not yet included in a Cloud SDK release.

If you want to use it before it gets included in a Cloud SDK release, then clone this repository and run the python code from your locally cloned repository.

E.G.

```sh
git clone https://github.com/googledatalab/datalab.git
cd datalab
./tools/cli/datalab.py create --subnet ...
```

PR #2016, opened 2018-06-02
redirect stderr to /dev/null for local "docker tag -f" failures

jramos:
The Datalab Jenkins build still uses an old Docker version with incompatible flags; this causes confusing output during local builds. E.g.

```bash
$ cd containers/datalab && ./build.sh
...
unknown shorthand flag: 'f' in -f
See 'docker tag --help'.
....
```

I've updated both build scripts to redirect these errors to `/dev/null`.


jramos:
I signed it!

PR #2018, opened 2018-06-05
Try to make the CLI integration tests more reliable.

ojarjur:
This change modifies the CLI integration test in two ways
to try to reduce test flakiness:

1. If picking a random zone, use a different zone for each test. This
   means that if a particular zone has an issue, we do not limit
   ourselves to that zone that we've picked.
2. Retry the test suite up to 3 times in case it fails.

### DIFF
--- a/tools/cli/tests/end-to-end.py
+++ b/tools/cli/tests/end-to-end.py
@@ -34,6 +34,7 @@ except ImportError:
 import uuid
 
 
+retry_count = 3
 python_executable = sys.executable
 connection_msg = (
     'The connection to Datalab is now open and will '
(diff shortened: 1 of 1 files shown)


tools/cli/tests/end-to-end.py
```diff
@@ -211,4 +216,9 @@ def test_connect(self):
 
 
 if __name__ == '__main__':
-    unittest.main()
+    for _ in range(retry_count):
+        runner = unittest.main(exit=False)
+        if runner.result.wasSuccessful():
+            sys.exit(0)
+    print('Test run failed {} times'.format(retry_count))
```
yebrahim: Can we get the last error to print it here too?

tools/cli/tests/end-to-end.py
```diff
@@ -211,4 +216,9 @@ def test_connect(self):
 
 
 if __name__ == '__main__':
-    unittest.main()
+    for _ in range(retry_count):
```
qimingj: Does it make sense to move the retry into the test itself, so we can still rely on unittest.main() to exit with a proper error code instead of us calling sys.exit? It might require a code refactoring so a TODO is fine.

ojarjur: Done.

PR #2028, opened 2018-06-29
Create a mock Datalab image that can be used when testing changes to the CLI

ojarjur:
This is meant to greatly reduce the turn-around time for debugging changes to the creation/connection logic of the CLI.

By having a much smaller mock image we should be able to get a much lower latency between making a change and seeing whether or not the connection still works.

### DIFF
--- a/containers/mock/Dockerfile
+++ b/containers/mock/Dockerfile
@@ -1,0 +1,32 @@
+# Copyright 2018 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


yebrahim:
Why not use the nginx image (or any other hello-world server) instead of building a similar one?

ojarjur:
@yebrahim I don't understand the question; we *are* using the nginx image.

ojarjur:
Testing this change, this cuts the time between creating a Datalab instance and getting the `You can connect to Datalab ...` message in half (going from 6 minutes to 3).

PR #2033, opened 2018-07-01
Added word highlighter jupyter in the DockerFile to highlight the selected word

umang-sh:
@yebrahim @nikhilk @ojarjur 
Added Jupyter Code highlighter extension to highlight the selected word
https://github.com/jcb91/jupyter_highlight_selected_word

It highlights the current selected word.
I have added and enabled the extension in DockerFile.I also added the BSD license for the same in license.txt

I also built the image for the same and checked how its working 
Its working fine as seen in the below screenshot
Let me know if anything needs to be done :) 
screen shot 2018-07-01 at 5 56 02 pm

### DIFF
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -261,7 +261,7 @@ RUN if [ -d /datalab/lib/pydatalab/.git ]; then \
     jupyter nbextension install --py datalab.notebook && \
     jupyter nbextension install --py google.datalab.notebook && \
     jupyter nbextension install --py jupyter_highlight_selected_word && \
-    jupyter nbextension enable --py jupyter_highlight_selected_word && \
+    jupyter nbextension enable --sys-prefix --py jupyter_highlight_selected_word && \
     jupyter nbextension enable --py widgetsnbextension && \
     source deactivate && \
 # Clean up
--- a/containers/base/Dockerfile
+++ b/containers/base/Dockerfile
@@ -260,6 +260,7 @@ RUN if [ -d /datalab/lib/pydatalab/.git ]; then \
     pip install --upgrade-strategy only-if-needed jupyter_highlight_selected_word && \
     jupyter nbextension install --py datalab.notebook && \
     jupyter nbextension install --py google.datalab.notebook && \
+    jupyter nbextension install --py jupyter_highlight_selected_word && \
     jupyter nbextension enable --py jupyter_highlight_selected_word && \
     jupyter nbextension enable --py widgetsnbextension && \
     source deactivate && \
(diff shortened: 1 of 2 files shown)


containers/base/Dockerfile
```diff
@@ -257,8 +257,11 @@ RUN if [ -d /datalab/lib/pydatalab/.git ]; then \
     pip install --upgrade-strategy only-if-needed --no-cache-dir . && \
     pip install --upgrade-strategy only-if-needed /datalab/lib/pydatalab/solutionbox/image_classification/. && \
     pip install --upgrade-strategy only-if-needed /datalab/lib/pydatalab/solutionbox/structured_data/. && \
+    pip install --upgrade-strategy only-if-needed jupyter_highlight_selected_word && \
     jupyter nbextension install --py datalab.notebook && \
     jupyter nbextension install --py google.datalab.notebook && \
+    jupyter nbextension install --py jupyter_highlight_selected_word && \
+    jupyter nbextension enable --py jupyter_highlight_selected_word && \
```
ojarjur: I built an image from this and deployed it from the CLI.

It looks like the change does not work in that scenario, as the extension is enabled in the user's home directory, which then gets shadowed by a directory on the VM.

I.E. this change works if you test the image locally, but not if you use the command line tool to deploy it to a VM.

The fix is to add a `--sys-prefix` flag to the enable command:

```sh
jupyter nbextension enable --sys-prefix --py jupyter_highlight_selected_word && \
```

This was not necessary for the other extension which we enable (the ipywidgets one) because that one is manually re-enabled at container startup [here](https://github.com/googledatalab/datalab/blob/master/containers/datalab/content/run.sh#L98).

Arguably, that may have been the wrong solution for ipywidgets, and both should be installed with the --sys-prefix option, but I am not asking you to do that as part of this PR.

containers/base/Dockerfile
```diff
@@ -257,8 +257,11 @@ RUN if [ -d /datalab/lib/pydatalab/.git ]; then \
     pip install --upgrade-strategy only-if-needed --no-cache-dir . && \
     pip install --upgrade-strategy only-if-needed /datalab/lib/pydatalab/solutionbox/image_classification/. && \
     pip install --upgrade-strategy only-if-needed /datalab/lib/pydatalab/solutionbox/structured_data/. && \
+    pip install --upgrade-strategy only-if-needed jupyter_highlight_selected_word && \
```
ojarjur: This can apparently be installed with conda, but since we are only installing it in one environment I don't know if that provides any benefit.

I.E. I'm not asking you to change that; just noting that it is an option.

containers/base/Dockerfile
umang-sh: @ojarjur 
Thanks! 
Added the sys prefix flag in my latest commit :)

PR #2037, opened 2018-07-03
Add support for instances without an external IP address.

### LINKED ISSUE #1844: Support connecting to a Datalab instance's internal IP address rather than its external address
Right now the `gcloud compute ssh` command (which `datalab connect` invokes) only supports connecting to a VM's external IP address.

That is a big limitation since external IP addresses are optional; not every VM has one.

However, there is an `--internal-ip` flag in the ["gcloud alpha compute ssh" and "gcloud beta compute ssh" commands](https://cloud.google.com/sdk/gcloud/reference/alpha/compute/ssh).

Once that flag is available in the main `gcloud compute ssh` command, we should add similar support to the Datalab command line tool.

This would require two things:

1. A flag to `datalab create` that tells it to not create an external IP address for the instance.
2. A `--internal-ip` flag for the `datalab connect` command that tells it to connect to the internal IP address rather than an external one.

ojarjur:
This change adds two flags:

1. A `beta-no-external-ip` flag for the `create` and `beta create-gpu`
   commands.
2. A `beta-internal-ip` flag to the `connect` command.

These tell the tool to create an instance with no external IP address
and connect to the instance's internal IP addres, respectively.

Due to the nature of Datalab, instances still need to be able to contact Google services in order to work. In particular, they need to be able to download the Docker images from GCR. As such, creating an instance without an external IP address requires that the corresponding subnet allow private IP addresses to access Google services.

This change also includes a check to make sure the subnet is properly configured to allow that.

This fixes #1844


ojarjur:
> LGTM but can we have some tests covering this? At least to check we're building the right gcloud commands?

@yebrahim Done. PTAL

PR #2042, opened 2018-07-13
Add Anchor style to Logo that will take to home

aayusharyan:
The home logo image did not had anchor style cursor which was difficult to comprehend as end user... So added style to improve that.


aayusharyan:
I signed it!

sources/web/datalab/static/appbar.html
```diff
@@ -5,7 +5,7 @@
 </head>
 <body>
   <span id="logo" class="nav navbar-brand pull-left">
-    <a id="homeLink"><img id="logoImg" alt="Google Cloud DataLab" style="display:none" /></a>
+    <a id="homeLink"><img id="logoImg" alt="Google Cloud DataLab" style="display:none; cursor: pointer;" /></a>
```
yebrahim: Can we add the style to the [CSS file](https://github.com/googledatalab/datalab/blob/74ed59030fa1ae3379ec4c5cc5d71a9b4067d761/sources/web/datalab/static/datalab.css#L277) instead.

aayusharyan: Sure, I will do that

aayusharyan: @yebrahim Thanks for review. But, in the CSS, there is no Style for id `logoImg`. so should I create one?

yebrahim: I put a link in my first comment, the line is here: https://github.com/googledatalab/datalab/blob/74ed59030fa1ae3379ec4c5cc5d71a9b4067d761/sources/web/datalab/static/datalab.css#L277
:)

yebrahim: The link was for the image. If you add the cursor style there, it'll be applied to the `img` element under `#logo`, which is the `logoImg`, just a different CSS selector.

aayusharyan:
@yebrahim Updated the PR, please review, thanks. :)

PR #2052, opened 2018-08-08
Upgrade to Jupyter Notebook 5.6.0

rileyjbauer:
This change upgrades the Jupyter Notebook version from `4.4.1` to `5.6.0`

Other upgrades include:
```
ipykernel: 4.5.2 --> 4.8.2
pip: 9.0.3 --> 18.0
pyzmq: 16.0.2 --> 17.1.0
```

The main changes center around the import of various Jupyter javascript files within `tree.html` and `nb.html` as well as changes to events as described in https://github.com/jupyter/notebook/issues/2499

There are few small API changes like `CodeCell`'s `config_defaults` becoming `options_default`.

The changes from using the function `require` to `requirejs` is based on the recommendation here:
https://github.com/jupyter/notebook/issues/155

The changes to `sources/web/datalab/static.ts` are all clean-up, and **can be ignored in terms of the review**. They include general clean-up, as well as avoiding some shadowed variable name collisions and mistakes caused by using `var` rather than `const`/`let`

`third_party/externs/ts/require/require.d.ts` was removed as it was not being used.


sources/web/datalab/static/datalab.js
```diff
@@ -35,29 +35,29 @@ function initializeDataLab(
...
         notebooklist.postLoad(ipy.notebook_list, ipy.new_notebook_widget, dialog);
         window.datalab.loaded = true;
       }
     });
-  }
-  else if (pageClass.indexOf('session_list') >= 0) {
+  } else if (pageClass.indexOf('session_list') >= 0) {
     // The page is finished loading after the notebook list is drawn for the first
     // time. This event is used also after loading the terminal list. These lists are
     // refreshed periodically though, so we need to only capture the first occurrence
```
yebrahim: No promise alternative for the line below? It'd be nice to get rid of `events` entirely as a dependency.

rileyjbauer: I think `promises.app_initialized` should work, but even still we would need events for the savewidget in the editor

Issue #2054, opened 2018-08-09
Unable to select existing subnet using the latest merge that includes the --subnet-name flag


jmmaldonado:
As of today the merge that includes the --subnet-name flag is not surfaced in the SDK
I have cloned the repo and executed the following command:

`./datalab.py create --network-name eci-vpc --subnet-name default --no-connect --verbosity debug --zone europe-west1-b datalabvm01`

The `default` subnet exists as can be seen by executing manually the gcloud command:
`gcloud compute networks subnets describe default --region europe-west1`

However the datalab create command fails as apparently it does not pass through the `--region` flag to the gcloud compute command, whether it is specified as the `--zone` flag in the datalab command, or it is selected via the interactive prompt of the datalab create command

DEBUG: (gcloud.compute.networks.subnets.describe) Underspecified resource [default]. Specify the [--region] flag.
Traceback (most recent call last):
  File "/google/google-cloud-sdk/lib/googlecloudsdk/calliope/cli.py", line 848, in Execute
    resources = calliope_command.Run(cli=self, args=args)
  File "/google/google-cloud-sdk/lib/googlecloudsdk/calliope/backend.py", line 770, in Run
    resources = command_instance.Run(args)
  File "/google/google-cloud-sdk/lib/surface/compute/networks/subnets/describe.py", line 49, in Run
    scope_lister=compute_flags.GetDefaultScopeLister(client))
  File "/google/google-cloud-sdk/lib/googlecloudsdk/command_lib/compute/flags.py", line 755, in ResolveAsResource
    default_scope, scope_lister)
  File "/google/google-cloud-sdk/lib/googlecloudsdk/command_lib/compute/flags.py", line 510, in ResolveResources
    api_resource_registry)
  File "/google/google-cloud-sdk/lib/googlecloudsdk/command_lib/compute/flags.py", line 421, in _ResolveUnderspecifiedNames
    raise UnderSpecifiedResourceError(names, [s.flag for s in self.scopes])
UnderSpecifiedResourceError: Underspecified resource [default]. Specify the [--region] flag.
**ERROR: (gcloud.compute.networks.subnets.describe) Underspecified resource [default]. Specify the [--region] flag.**
Traceback (most recent call last):
  File "./datalab.py", line 420, in run
    sdk_version=sdk_version, datalab_version=datalab_version)
  File "/home/jmmaldonado/datalabcli/tools/cli/commands/create.py", line 802, in run
[...]

PR #2055, opened 2018-08-09
Added --region flag to ensure_subnet_exists

jmmaldonado:
Added required --region flag to the ensure_subnet_exists procedure.
Region is calculated based on the zone selected by the user when issuing the datalab create command, either by argument or by interactive selection

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -565,19 +565,20 @@ def ensure_network_exists(args, gcloud_compute, network_name):
     return
 
 
-def ensure_subnet_exists(args, gcloud_compute, subnet_name):
+def ensure_subnet_exists(args, gcloud_compute, subnet_region, subnet_name):
     """Check the specified subnet if it does not exit with error.
 
     Args:
       args: The Namespace returned by argparse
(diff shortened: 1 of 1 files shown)


ojarjur:
Thanks for catching this and putting in the effort to fix it.

In #2037, I am adding a [`get_region_name` method](https://github.com/googledatalab/datalab/pull/2037/files#diff-ed4811dbebc089623dd82cca13888c40R578). I think that might be a good resource to use in your fix (once that PR gets merged).

Also, `flake8` is complaining that line 773 is now too long. Can you split it across multiple lines? We can't merge until the `flake8` checks pass.

Thanks again for the help here!

jmmaldonado:
Thanks Omar! I think I messed up pushing the wrong button before making the change to the file to pass the line length test. I am submitting the change as is, and once your PR is merged we can update the region calculation with yours ... it will be more resilient if there are changes to the zone / region naming convention

jmmaldonado:
This fixes Issue #2054

ojarjur:
@jmmaldonado 

> I am submitting the change as is, and once your PR is merged we can update the region calculation with yours

That's perfectly reasonable. Thanks for putting this together.

PR #2056, opened 2018-08-10
Timeout Datalab connections if they never become healthy

### LINKED ISSUE #2051: Creating the SSH tunnel can silently fail.
It looks like there is a scenario where the SSH tunnel created by the `datalab create` and `datalab connect` commands can silently fail.

This leads to a situation where the Datalab instance is running properly, but the command to create or connect to it hangs forever with a message saying `Waiting for Datalab to be reachable at http://localhost:8081/`.

I encountered this with my own instance where the connection waited for far longer than expected, so I manually SSH'ed in to the VM and confirmed that Datalab was running on it. I then confirmed that the SSH tunnel was not running.

Killing the command and then running `datalab connect` worked, so it seems to have been flakiness on the part of the SSH tunnel setup.

ojarjur:
This change adds a timeout to the health check, after which the
SSH tunnel will be killed (triggering it to be recreated), if
the health check never succeeds before that timeout.

This adds an important check to catch issues where the SSH tunnel
itself is unhealthy as opposed to the connecting instance being
unhealthy.

This fixes #2051

### DIFF
--- a/tools/cli/datalab.py
+++ b/tools/cli/datalab.py
@@ -183,7 +183,7 @@ def gcloud_compute(
       stderr: The 'stderr' argument for the subprocess call
       wait: Whether or not to wait for the command to complete
     Returns:
-      A subprocess.Popen object iff `wait` is True
+      A subprocess.Popen object iff `wait` is falsy
     Raises:
       KeyboardInterrupt: If the user kills the command
       subprocess.CalledProcessError: If the command dies on its own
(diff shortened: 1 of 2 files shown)


tools/cli/datalab.py
```diff
@@ -181,6 +181,9 @@ def gcloud_compute(
       stdin: The 'stdin' argument for the subprocess call
       stdout: The 'stdout' argument for the subprocess call
       stderr: The 'stderr' argument for the subprocess call
+      wait: Whether or not to wait for the command to complete
+    Returns:
+      A subprocess.Popen object iff `wait` is True
```
rileyjbauer: This seems like it's backwards; the Popen object is returned iff wait is false. Am I misreading? Same below

ojarjur: You're right; thanks for catching that.

PR #2070, opened 2018-08-28
 Ensure that websockets are shimmed before any Jupyter client code runs.

### LINKED ISSUE #2069: Race condition when setting up a shim for socket.io
There is a race condition bug in the way we setup socket.io when we are replacing the native WebSocket implementation (which is necessary when connecting from Cloud Shell).

The issue is that our websocket.js code, which replaces the native `window.WebSocket` with our shim, runs *after* the main.min.js code, which creates the `Kernel` object.

The `Kernel` code does not directly call `window.WebSocket`. Instead, it copies the value of `window.WebSocket` at the time it was created into a field of the object, and then calls the value of that field.

This means that if we replace the `window.WebSocket` value before the `Kernel` object gets created, then our shim works as expected. However, if that replacement does not happen until the `Kernel` object is created, then the shim will not work.

We have two options to fix this:

1. Ensure the `websocket.js` code runs before the `main.min.js` code
or
2. Make the `websocket.js` code modify the field in the `Kernel` object.

I'm not sure if the second option is even possible, so I think the first is by far the best choice.

ojarjur:
This fixes #2069

### DIFF
--- a/sources/web/datalab/templates/edit.html
+++ b/sources/web/datalab/templates/edit.html
@@ -88,7 +88,7 @@
     window.datalab = {};
 
     $("#appBar").load("<%baseUrl%>static/appbar.html", function() {
-        require(['websocket'], (websocket) => {
+        requirejs(['websocket'], (websocket) => {
             requirejs(['edit/js/main.min']);
         });
     });
--- a/sources/web/datalab/templates/edit.html
+++ b/sources/web/datalab/templates/edit.html
@@ -88,7 +88,9 @@
     window.datalab = {};
 
     $("#appBar").load("<%baseUrl%>static/appbar.html", function() {
-      requirejs(['edit/js/main.min']);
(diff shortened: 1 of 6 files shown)


sources/web/datalab/templates/edit.html
```diff
@@ -88,7 +88,9 @@
     window.datalab = {};
 
     $("#appBar").load("<%baseUrl%>static/appbar.html", function() {
-      requirejs(['edit/js/main.min']);
+        requirejs(['websocket'], (websocket) => {
```
yebrahim: Why don't we add `'edit/js/main.min'` to the list of dependencies of the `websocket` module? Is there a scenario where that doesn't work?

yebrahim: I see. LGTM then.

PR #2075, opened 2018-08-31
Enable Datalab to run in a read-only mode.

ojarjur:
In this mode, Datalab will refuse all websocket requests, and any
HTTP requests other than 'HEAD' or 'GET' requests.


sources/web/datalab/config/settings.json
```diff
@@ -47,5 +47,6 @@
   "socketioPort": 8084,
   "supportedFileBrowserSources": ["jupyter"],
   "supportUserOverride": false,
-  "useWorkspace": false
+  "useWorkspace": false,
+  "readOnly": false
```
yebrahim: nit: keep sort order?

Issue #1984, opened 2018-04-10
google-cloud-dataflow and apach-beam libraries


eilalan:
Hello,

Is it possible to have datalab with the latest versions of dataflow and apache-beam.
The new dataflow has a direct runner that is 15 times faster and additional new features.
when I run pip update command, the libraries modules for apache-beam are not recognized by the import command.

Please advice.

Many thanks in advance,
Eila

rileyjbauer:
Hello!

This should be fixed as of #2083 and will be in the next release.

Thank you for raising this!

PR #2083, opened 2018-10-01
Update pytz and replace google-cloud-dataflow lib with apache-beam[gcp]

rileyjbauer:
We are replacing the `google-cloud-dataflow` package with `apache-beam` as, according to the Google Cloud Dataflow project's release notes, "The Cloud Dataflow SDK 2.5.0 is the last Cloud Dataflow SDK release that is separate from the Apache Beam SDK releases."
https://cloud.google.com/dataflow/release-notes/release-notes-python


rileyjbauer:
This should fix #1984

PR #2093, opened 2018-11-21
Add third party license files to container image.

qimingj:


containers/base/license.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash -e
```
yebrahim: Seems to me most this file can be replaced with a one-liner bash command?
Maybe something like `comm -23 <(cat test.txt | sed s/,.*// | sort) <(pip freeze | sed s/=.*// | sort)`? Might not be very readable, but I'd say it's more reliable than writing new bash scripts.

To make it more readable, maybe store each content in a variable, then compare using `comm`, if all you need is to subtract the two lists.

qimingj: That looks cool. In this case I need to read a csv file and get the first column, not simply (cat test.txt). Also I need to dump the diff if there is any. Does comm work for that?

yebrahim: The command above doesn’t just cat, it extracts the first column. 

Do you want the diff or subtraction? If you also want lines that are in either file and not the other, use the “diff” command. “comm” gives you a way to do subtraction.

PR #2110, opened 2019-01-06
Add argument to bypass firewall rule creation during when creating Datalab instance

3thanZ:
The `--network-name` and `--subnet-name` allows for an instance to be created in a shared VPC network from a different host project. However, the firewall rule checks and creation will fail. Therefore, an argument to skip the firewall rule check and creation would enable the `datalab create` command to succeed.

### DIFF
--- a/tools/cli/commands/create.py
+++ b/tools/cli/commands/create.py
@@ -522,6 +522,13 @@ def flags(parser):
         help='do not automatically backup the disk contents to GCS')
 
     parser.add_argument(
+        '--no-firewall-rule',
+        dest='no_firewall_rule',
+        action='store_true',
+        default=False,
+        help='do not check and automatically create SSH firewall rule')
+
+    parser.add_argument(
         '--beta-no-external-ip',
         dest='no_external_ip',
         action='store_true',
@@ -919,8 +926,11 @@ def prepare(args, gcloud_compute, gcloud_repos):
     """
     network_name = args.network_name
     ensure_network_exists(args, gcloud_compute, network_name)
-    prompt_on_unexpected_firewall_rules(args, gcloud_compute, network_name)
-    ensure_firewall_rule_exists(args, gcloud_compute, network_name)
+    if not args.no_firewall_rule:
+        prompt_on_unexpected_firewall_rules(args, gcloud_compute, network_name)
+        ensure_firewall_rule_exists(args, gcloud_compute, network_name)
+    else:
+        print('Bypassing firewall rule check and creation')
 
     disk_name = args.disk_name or '{0}-pd'.format(args.instance)
     ensure_disk_exists(args, gcloud_compute, disk_name)


yebrahim:
I think you need to sign the CLA.

ojarjur:
I don't think skipping the firewall checking is the best solution to the described issue. In particular, the firewall rule is still required, so if it is missing the attempts to connect to the created instance will fail.

Instead, I would suggest making the existing check work when the network is in another project.

That could be done by getting the project name for the expected firewall from the self link of the network, and then using that in the nested `gcloud` calls to ensure the firewall rule exists.

3thanZ:
Hi Omar. That is a good suggestion. I think that allowing the user to
bypass firewall rule creation is still a useful option. Enterprises
normally allow users to create compute instances, but reserve firewall rule
creations for the network team.

I will submit a different PR for the firewall rule on host VPC.

3thanZ:
> I think you need to sign the CLA.

Hi @yebrahim 
I'm submitting this on behalf of the company I work for. The person in charge to review and submit the CLA is currently still away. Will let you know when CLA is submitted. Thanks.

3thanZ:
Hi @yebrahim. CLA submitted.

yebrahim:
Hi @3thanZ, I think you might need to reply here saying "I signed it." just like the bot suggests above (not sure what the bot looks for in your comment exactly).

3thanZ:
I signed it!

yebrahim:
@3thanZ I think it's safer to follow @ojarjur's suggestion. Skipping the check and relying on the user to make sure it's been created correctly can easily lead to obscure bugs. I'd rather we generalize the check and creation code path.

3thanZ:
Hi @yebrahim 
How would the firewall rule check be helpful in organisations where those who are creating Datalab instances have no permissions to list or create firewall rules? Users in these organisations will be blocked from creating Datalab instances.

ojarjur:
@3thanZ Here's my take on the overall user experience:

Even if a user does not have permission to *create* firewall rules, the check is still valuable as it will prevent them from accidentally creating an instance which is unusable, and it will present an error message to them saying what they need to do to fix the issue (i.e. have a network administrator add the firewall rule).

If the user does not have permission to even *list* firewall rules, then I think that would be a setup that we should not support, and instead the organization needs to give the firewall-listing permission out to anyone who needs to create Datalab instances. Otherwise, we would have the situation where the user creating a Datalab instance has no way of knowing whether or not the instance is actually usable, and if it is not usable, they would have no way of debugging the issue.

If giving permission to list firewall rules to a Datalab user is an issue, that organization can still proceed by having an administrator create the Datalab instance on the user's behalf by passing in the `--for-user` flag.

mdhedley:
Just added Issue #2126 related to this. I think there is value in this flag. It doesn't make sense to create a firewall rule when there isn't any intention for external access.

rileyjbauer:
Closing this since #2135 was merged

PR #2112, opened 2019-01-12
Support datalab connect() to connect a GCE instance created from Deep…

yixinshi:
…learning images

### DIFF
--- a/tools/cli/commands/utils.py
+++ b/tools/cli/commands/utils.py
@@ -70,7 +70,7 @@ class InvalidInstanceException(Exception):
         'The specified instance, {}, does not appear '
         'to have been created by the `datalab` tool, or '
         'from any GCE Deeplearning images. Therefore it '
-        'so cannot be managed by `datalab` tool.')
+        'cannot be managed by `datalab` tool.')
 
     def __init__(self, instance_name):
         super(InvalidInstanceException, self).__init__(
--- a/tools/cli/commands/utils.py
+++ b/tools/cli/commands/utils.py
@@ -68,7 +68,8 @@ class InvalidInstanceException(Exception):
 
     _MESSAGE = (
         'The specified instance, {}, does not appear '
-        'to have been created by the `datalab` tool, and '
+        'to have been created by the `datalab` tool, or '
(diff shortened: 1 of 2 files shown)


tools/cli/commands/utils.py
```diff
@@ -68,7 +68,8 @@ class InvalidInstanceException(Exception):
 
     _MESSAGE = (
         'The specified instance, {}, does not appear '
-        'to have been created by the `datalab` tool, and '
+        'to have been created by the `datalab` tool, or '
+        'from any GCE Deeplearning images, and '
```
qimingj: Do we need the "and"?

tools/cli/commands/utils.py
```diff
@@ -240,22 +241,35 @@ def flatten_metadata(metadata):
     return result
 
 
-def _check_datalab_tag(instance, tags):
+def _check_datalab_tag(instance, status_tags_and_metadata):
```
qimingj: Also rename the function?

tools/cli/commands/utils.py
yixinshi: I rewrote it.

tools/cli/commands/utils.py
yixinshi: Done

yixinshi:
Thanks, qimingj. PTAL

qimingj:
There is a test failure. I just reran it and see if it is just flaky.

tools/cli/commands/utils.py
```diff
@@ -68,8 +68,9 @@ class InvalidInstanceException(Exception):
 
     _MESSAGE = (
         'The specified instance, {}, does not appear '
-        'to have been created by the `datalab` tool, and '
-        'so cannot be managed by it.')
+        'to have been created by the `datalab` tool, or '
+        'from any GCE Deeplearning images. Therefore it '
+        'so cannot be managed by `datalab` tool.')
```
ojarjur: The "so " here is redundant.

yixinshi: Done. Can you grant me the write access to the repo? Thanks!

Issue #2126, opened 2019-03-25
--beta-no-external-ip does not prevent firewall rule creation


mdhedley:
Expected behavior for --beta-no-external-ip would be to prevent the use of an external IP address, and also to not make adjustments to the external firewall rules of the associated network.

However even though the flag does prevent the creation of an external IP address it still attempts to modify firewall rules to allow external access. This is not the desired behavior.

mdhedley:
Misunderstood purpose of fireall rule.

mdhedley:
The current system creates a firewall rule that allows port 22 from 0.0.0.0 on the network provided. In most cases where --no-public-ip is used these are secure environments where users want to limit the attack surface of the environment. So creating an open port 22 in this case does not make sense.

Additionally the existing firewall rule creation opens port 22 for all instances in the network which could create unexpected security issues for users.

Issue #2125, opened 2019-03-25
Shared VPC support is blocked by firewall rule


mdhedley:
The use of shared VPC appears to be supported but fails when it attempts to create the firewall based on the network name.

Example:
datalab create --network-name /projects/<host-project>/global/networks/<shared vpc> --subnet <subnet name>

Produces error:
Creating the firewall rule projects/<host-project>/global/networks/<shared vpc>-allow-ssh

Creating firewall...
failed.
ERROR: (gcloud.compute.firewall-rules.create) Could not fetch resource:
 - Invalid value for field 'resource.name': 'projects/<host-project>/global/networks/<shared vpc>-allow-ssh'. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)'

This prevents the use of shared vpcs with datalab.

PR #2127, opened 2019-03-25
Removed the creation of the firewall rule for --beta-no-external-ip.

mdhedley:
Public firewall creation for no-external-ip doesn't make sense as there will be no external access. Removing the firewall creation for this flag. This resolves both issue #2126 and #2125 .


mdhedley:
Closing as I messed up by git history. Will resubmit from new fork.

Issue #2128, opened 2019-03-25
Build does not complete due to issue with Py-YAML


mdhedley:
It appears that the build script is being impacted for the pip change documented in this issue:
https://github.com/pypa/pip/issues/5247

PyYAML no longer installs as a result of this change in pip 10, and as a result all builds are failing, and all pull requests fail to build.

mdhedley:
You can reproduce this error by running build.sh as is. It appears that the default pyyaml package installed is now 5.15 while the google-cloud-dataflow package requires 4.0 > pyyaml < 3.12. Working on a fix.

PR #2129, opened 2019-03-26
Fix Issue with Build

mdhedley:
Fix for build issue #2128. The pyyaml that gets deployed by conda by default is not supported with google-cloud-dataflow. Setting the version for pyyaml fixes this problem. 

This only requires a change for python 2, not python 3.


mdhedley:
Messed up by git branches. Will resubmit.

PR #2131, opened 2019-03-26
Resolves Issue with firewall rule creation when no public IP address is specified

mdhedley:
Fixes Issue #2126 & #2125 
--no-public-ip currently attempts the creation of the public firewall rule, even though there is no public ip address specified. This results in 2 issues, it blocks the use of shared vpcs and also doesn't behave as expected as a firewall rule is not required. This change disables firewall rule creation when the --no-public-ip address is specified. Build will fail as a result of #2128.


mdhedley:
Misunderstood use of firewall rule.

PR #2132, opened 2019-03-26
Resolves Firewall Rule Created when no public IP specified.

mdhedley:
Resolves Issue #2126 

Does not create firewall rule when --beta-no-external-ip is selected.

Issue #2128 prevents this from building.


yebrahim:
I think this is a reasonable change. Tagging @rileyjbauer and @ojarjur to take a look as well.

mdhedley:
2134, and 2135 resolve the need for this.

PR #2135, opened 2019-04-02
Added --no-firewall-rule

### LINKED ISSUE #2126: --beta-no-external-ip does not prevent firewall rule creation

mdhedley:
This creates a new flag --no-firewall-rule which disables firewall rule creation. fixes #2126


TV4Fun:
@yebrahim, @rileyjbauer, this change has been approved. How long to get it merged?

Issue #2133, opened 2019-03-27
Google Cloud Datalab Connect Failing with Beta Gpu Instance


nehal1005:
Stuck on the status below when attempting to reconnect using datalab connect gpu-instance.

![image](https://user-images.githubusercontent.com/29584853/55112626-e5c9b080-50d4-11e9-8809-576be708ec3c.png)

I have confirmed I can ssh using:
`gcloud compute ssh --zone=us-west1-b datalab@gpu-instance`

When I try the command below for more information I get....
`datalab connect --ssh-log-level=verbose gpu-instance`
`channel 1: open failed: connect failed: Connection refused`

PR #2147, opened 2019-07-10
Fixed issue in which datalab does not start for GPU instances

emailbob:
- Fixed issue for the GPU instance in which it does not start because it fails to pull down the docker container (https://github.com/googledatalab/datalab/issues/2133)

- Updated the NVIDA driver to 418.67

- Added ExecStop and ExecStopPost to the datalab service so systemctl can be used to stop the service.


sanjaykhobragade:
@emailbob When will this issue fix merge?

stasiek:
Can we get this done, PLEASE!?

PR #2186, opened 2022-09-02
Replace repo with deprecation announcement.

YuxuanChen:
Google Cloud Datalab has passed the deprecation notice date published on the [docs](https://cloud.google.com/datalab/docs).


crwilcox:
Is there a reason we wouldn't update the README, then archive this repository? It seems unnecessary to delete the source, and there are a few files that even if the choice was made should likely stay (LICENSE, CoC, etc)

YuxuanChen:
> Is there a reason we wouldn't update the README, then archive this repository? It seems unnecessary to delete the source, and there are a few files that even if the choice was made should likely stay (LICENSE, CoC, etc)

Either works for me, I think the archiving is the most important part.

crwilcox (approved):

crwilcox (approved):

